Issue #17, opened 2020-07-29
Standardize Web3API Manifest Schema


dOrgJelli:
Create tooling to enable standardization of the Web3API manifest schema. This tooling should include:
- sanitization (https://github.com/paazmaya/yaml-validator)
- versioning
- version migrations (https://levelup.gitconnected.com/an-approach-to-javascript-object-schema-migration-ae1bd9f0ce78)
- type safe usage (https://github.com/bcherny/json-schema-to-typescript)

PR #50, opened 2020-10-19
Feat/standardize web3api manifest

cbrzn:
closes: #17


packages/cli/src/lib/helpers/validators.ts
```diff
@@ -0,0 +1,58 @@
...
+    // First case
+    const objectNotDefined = /is not defined in the schema/.test(message);
+    if (objectNotDefined) {
+      const unacceptedVariable = message.match(
+        new RegExp("The property (.*) is not defined")
+      );
+      throw Error(
+        `Field ${
+          unacceptedVariable![1]
+        } is not accepted in the schema. Please check the accepted fields here: LINK_TO_SCHEMA`
```
cbrzn, line 28: I was thinking that we should show somewhere the structure accepted for the `web3api.yaml`

packages/cli/tsconfig.json
```diff
@@ -10,7 +10,8 @@
     "sourceMap": true,
     "esModuleInterop": true,
     "strict": true,
-    "lib": ["es2015", "es6", "esnext.asynciterable", "dom"]
+    "lib": ["es2015", "es6", "esnext.asynciterable", "dom"],
+    "resolveJsonModule": true
```
cbrzn, line 14: This is needed in order to read JSON files

demos/simple-storage/protocol/web3api.yaml
```diff
@@ -1,8 +1,9 @@
 description: SimpleStorage Web3API Example
+version: "1.0.0"
 repository: https://github.com/web3-api/prototype
 mutation:
   schema:
-    file: ./src/mutation/schema.graphql
+    file: ./src/mutation/index.ts
```
dOrgJelli, line 6: Why are you changing this here? This is incorrect. Were you doing this to test the validation logic? Please create tests & test projects for this purpose.

packages/cli/src/lib/Compiler.ts
```diff
@@ -247,7 +247,7 @@ export class Compiler {
     quiet?: boolean,
     verbose?: boolean
   ): Promise<string> {
-
+    console.log()
```
dOrgJelli, line 250: remove

packages/cli/src/lib/Manifest.ts
```diff
@@ -14,9 +14,12 @@ interface ClientModule {
   module: ModulePath;
 }
 
+type ManifestVersion = string;
```
dOrgJelli, line 17: Why is this useful?

packages/cli/src/lib/Manifest.ts
```diff
@@ -14,9 +14,12 @@ interface ClientModule {
...
+
 export interface Manifest {
```
dOrgJelli, line 19: This should be the generated type, no?

packages/cli/src/lib/Web3API.ts
```diff
@@ -1,8 +1,9 @@
-import { Manifest } from "./Manifest";
-
 import YAML from "js-yaml";
 import fs from "fs";
 
+import { Manifest } from "./Manifest";
+import { manifestValidation } from "./helpers/validators";
```
dOrgJelli, line 5: I think manifest validation should be inside of the client package, and the CLI can import it from there. Clients will need to validate manifests at run-time when they download new Web3APIs.

packages/cli/src/lib/helpers/schema.json
```diff
@@ -0,0 +1,94 @@
+{
```
dOrgJelli, line 1: looks great! Let's move this into its own package so that all other packages can use it (i.e. all clients & the CLI). Maybe call the package `manifest-schema`

packages/cli/src/lib/helpers/validators.ts
```diff
@@ -0,0 +1,98 @@
...
+          throw Error(
+            `Property ${pathMapping} has the value ${instance}, which is a file that does not exists`
+          );
+        }
+    }
+  }
+
+  compile(ManifestSchema as JSONSchema4, "Web3API").then((file) => {
+    // @TODO: Make sure where do we want to generate this file
+    writeFile("./Manifest.ts", file, (error: Error | null) => {
```
dOrgJelli, line 93: I don't think this should be here

packages/cli/tsconfig.json
dOrgJelli: makes sense

dOrgJelli (changes requested):

demos/simple-storage/protocol/Manifest.ts
```diff
@@ -0,0 +1,42 @@
+/* tslint:disable */
```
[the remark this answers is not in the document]
cbrzn: This is the autogenerated file based on JSON schema, I was wondering where should this autogenerated file should live? Just for test purposes

demos/simple-storage/protocol/web3api.yaml
cbrzn: That's correct, was just for testing validation logic - Since this is a WIP PR I was testing with the already existing project. If you think we should create a test project for this no worries I will create it

packages/cli/src/lib/Manifest.ts
cbrzn: Testing purposes at first, it is not useful at all. Will remove

packages/cli/src/lib/Manifest.ts
cbrzn: Yup yup

packages/cli/src/lib/helpers/validators.ts
cbrzn: Where should this go sir? I was wondering that after the validation, it should generate the type file. Would love to hear your thoughts tho

packages/cli/src/lib/Web3API.ts
cbrzn: Ok ok makes sense

cbrzn:
Did some changes. Feel free to check it - Anyways, let's talk about this on tuesday's call :-D @dOrgJelli

cbrzn:
Implemented a PoC of the migrator. Right now we are using any for the Schema object and we need to use the `Manifest` type which is autogenerated based on the JSON schema. Let's discuss when possible when should we generate this file and where.

cbrzn:
After making the `Manifest` file autogenerated, I've come to some thought about what is the CLI expecting from this file... I would like to discuss this further, to make sure we are on the same page on how the structure of this manifest is going to be

packages/cli/src/lib/Web3API.ts
```diff
@@ -12,10 +12,8 @@ export class Web3API {
     if (!manifest) {
       throw Error(`Unable to parse manifest: ${manifestPath}`);
     }
-
-    // TODO: add validation
-    // - validate manifest structure
-    // - ensure everything exists
+  
+    manifestValidation(manifest)
```
cbrzn, line 16: I think that here, we should redefine the `manifest` variable. The reason is that on this function, first, we check if the manifest passed as argument is outdated, if that's true, then it's going to be updated, hence, editing this variable that is being passed. If we don't update the `manifest` variable here, it will compile the older version of the manifest.

packages/client-js/src/manifest/generate.js
```diff
@@ -0,0 +1,17 @@
+const schema = require("@web3api/manifest-schema").default;
+const compile = require("json-schema-to-typescript").compile;
+const writeFileSync = require("fs").writeFileSync;
+
+const generateManifest = () => {
```
cbrzn, line 5: Should this function be moved to the `manifest-schema` package? 🤔  Just thinking out loud here

packages/cli/src/lib/Web3API.ts
dOrgJelli: Signature should be something like this: `manifestValidation(manifest: AnyManifest): LatestManifest`, so we can set `manifest` equal to return value.

Also change the function name to `sanitizeAndUpgrade`

dOrgJelli: Also change the import path from root level to some nested folder like `@web3api/client-js/manifest`, this way top level imports is very minimal and easy to understand.

packages/client-js/src/Web3API.ts
```diff
@@ -102,14 +104,20 @@ export class Web3API {
...
-         (name === "web3api.yaml" || name === "web3apy.yml")) {
+         (name === "web3api.yaml" || name === "web3api.yml")) {
         const manifestStr = await portals.ipfs.catToString(path);
         this._manifest = YAML.safeLoad(manifestStr) as Manifest | undefined;
 
         if (this._manifest === undefined) {
           throw Error(`Unable to parse web3api.yaml\n${manifestStr}`);
         }
 
+        const latestVersion: string = require("../package.json").version
```
dOrgJelli, line 115: Instead of doing run-time requiring of the package.json, let's try to do this instead inside of the `manifest/versions` directory let's add a `latest.ts` file that contains:
```typescript
export const version: string = "0.0.1-alpha.1";
```

dOrgJelli (changes requested):

cbrzn:
> Also change the import path from root level to some nested folder like @web3api/client-js/manifest, this way top level imports is very minimal and easy to understand.

Regarding that point i've been doing some research and came up with this: https://github.com/microsoft/TypeScript/issues/8305

Looks like there is no an official way to do this. Reading that thread you will come up to this: https://github.com/microsoft/TypeScript/issues/33079 but not sure how should we tackle this? @dOrgJelli

dOrgJelli (approved):

PR #79, opened 2020-12-04
Add eslint and pretttier

mpetrunic:
- added root eslint and prettier config
- added eslint extension for assemblyscript package
- added lint ci

I haven't fixed errors as it would produce huge diff. My plan is to fix errors in separate PR.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -9,10 +9,15 @@ jobs:
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@v2
+
+      - name: Read .nvmrc
+        run: echo ::set-output name=NVMRC::$(cat .nvmrc)
+        id: nvm
+      
       - name: Set up Node.js
         uses: actions/setup-node@master
         with:
-          node-version: 14.4.0
+          node-version: '${{ steps.nvm.outputs.NVMRC }}'
 
       - uses: actions/cache@v2
         id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
@@ -27,4 +32,4 @@ jobs:
         run: yarn install --nonInteractive --frozen-lockfile
 
       - name: Lint code
-        run: yarn run lint:code:ci
+        run: yarn run lint:ci
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,30 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    name: "Lint"
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Node.js
+        uses: actions/setup-node@master
+        with:
+          node-version: 14.4.0
+
+      - uses: actions/cache@v2
+        id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+        with:
+          path: |
+            node_modules
+            */*/node_modules
+          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Install dependencies
+        if: steps.cache-deps.outputs.cache-hit != 'true'
+        run: yarn install --nonInteractive --frozen-lockfile
+
+      - name: Lint code
+        run: yarn run lint:code:ci
--- a/package.json
+++ b/package.json
@@ -19,8 +19,8 @@
     "build:cli": "yarn workspace @web3api/cli build",
     "build:client-js": "yarn workspace @web3api/client-js build",
     "build:wasm-ts": "yarn workspace @web3api/wasm-ts build",
-    "lint:code": "lerna run lint:code",
-    "lint:code:ci": "lerna run --parallel lint:code",
+    "lint": "lerna run lint",
+    "lint:ci": "lerna run --parallel lint",
     "test": "yarn test:wasm-ts && yarn test:client-js",
     "test:wasm-ts": "yarn workspace @web3api/wasm-ts test",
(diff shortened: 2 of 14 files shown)


mpetrunic:
Something to think about:
https://github.com/dotansimha/graphql-eslint

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,30 @@
...
+  lint:
+    name: "Lint"
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Node.js
+        uses: actions/setup-node@master
+        with:
+          node-version: 14.4.0
```
dOrgJelli, line 15: Let's use the node version specified in the .nvmrc file at root. Is it possible to instead just use `nvm` in this job script, and run `nvm install` & `nvm use`? This way we don't write the version number in multiple places.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,30 @@
...
+            node_modules
+            */*/node_modules
+          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Install dependencies
+        if: steps.cache-deps.outputs.cache-hit != 'true'
+        run: yarn install --nonInteractive --frozen-lockfile
+
+      - name: Lint code
+        run: yarn run lint:code:ci
```
dOrgJelli, line 30: Will there be a case where we don't lint code? If not, can we change to just `lint:ci`?

package.json
```diff
@@ -19,13 +19,22 @@
...
+    "lint:code:ci": "lerna run --parallel lint:code",
     "test": "yarn test:wasm-ts && yarn test:client-js",
     "test:wasm-ts": "yarn workspace @web3api/wasm-ts test",
     "test:client-js": "yarn workspace @web3api/client-js test",
     "test:ci": "yarn test:wasm-ts && yarn test:ci:client-js",
     "test:ci:client-js": "yarn workspace @web3api/client-js test:ci"
   },
   "devDependencies": {
-    "lerna": "3.22.1"
+    "@typescript-eslint/eslint-plugin": "^4.9.0",
```
dOrgJelli, line 31: Can we use fixed versions wherever possible (no `^`)? I've been bit a number of times by minor version number changes introducing breaking changes :P

packages/cli/src/cli.ts
```diff
@@ -1,7 +1,7 @@
 import { build } from "gluegun";
 
 interface args {
-  [key: string]: any
+  [key: string]: unknown
```
dOrgJelli, line 4: Going to be using `unknown` more ;) thanks for introducing me to this: https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type

packages/cli/src/lib/textile/user-auth.ts
```diff
@@ -3,6 +3,7 @@ import { print } from "gluegun";
 
 const requireKeytar = () => {
   try {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
     return require("keytar");
```
dOrgJelli, line 7: Yah we should probably just get rid of this textile integration since it isn't complete. We can come back to the history and finish the integration.

dOrgJelli: Feel free to just leave this alone :P I can fix this

packages/client-js/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "rimraf build && yarn build:manifest && tsc",
     "build:manifest": "node ./scripts/generateManifest.js",
+    "lint:code": "eslint --color --ext .ts src/",
```
dOrgJelli, line 9: This will by default use the .eslintrc.js file at the root folder? cool!

dOrgJelli (changes requested):

.github/workflows/lint.yaml
mpetrunic: While, we could, setup-node action is probably better as it caches node version so it's not re-downloaded on every action run. I think I can read node-version variable from .nvmrc file anyways.

.github/workflows/lint.yaml
mpetrunic: There could also be lint:types which run tsc as a type checker. If you would like we can have lint:ci to just be eslint and have
something like `check:types` for type checking?

I'm gonna tackle type checking in separate PR. Issue is that you have excluded test files from tsconfig. I assume because you don't wan't them in your dist/build files but that also means they never get's type checked. Some common solution is to have tsconfig.json which includes everything and tsconfig.build.json which extends from tsconfig.json but excludes tests.

package.json
mpetrunic: Yes, of course!

packages/cli/src/cli.ts
mpetrunic: There is a recommended rule which won't allow explicit any anyways :smile:

packages/cli/src/lib/textile/user-auth.ts
mpetrunic: Yeah, wasn't planning on touching that. I mean, if you wan't to load some expensive dependency on demand, it's perfectly fine to use require. That is also valid for optional/peer dependencies. But those are usually rare occurencies so we can just disable rule on those particular exceptions.

packages/client-js/package.json
mpetrunic: It will use nearest .eslintrc.js and combine with every .eslintrc.js until it finds one with `{root: true}`

.github/workflows/lint.yaml
dOrgJelli: Okay that makes sense. Yes, having a tsconfig.build.json or tsconfig.dist.json sounds like a good solution here.

dOrgJelli (approved):

PR #83, opened 2020-12-07
Fix lint and type errors

mpetrunic:
Opened draft so we don't merge it before it's checked if it's gonna cause a lot of conflicts in issue28 branch.


lerna-debug.log
```diff
@@ -0,0 +1,14 @@
...
+0 silly argv   D: true,
+0 silly argv   globs: [],
+0 silly argv   lernaVersion: '3.22.1',
+0 silly argv   '$0': '/home/mpetrunic/.yarn/bin/lerna',
+0 silly argv   pkg: '@types/ipfs-http-client'
+0 silly argv }
+1 notice cli v3.22.1
+2 verbose rootPath /home/mpetrunic/projects/web3api-prototype
+3 error Error: 404 Not Found - GET https://registry.npmjs.org/@types%2fipfs-http-client - Not found
+3 error     at /home/mpetrunic/projects/web3api-prototype/node_modules/@evocateur/npm-registry-fetch/check-response.js:104:15
```
mpetrunic, line 14: woops, *.log files needs to be added in gitignore :smile:

packages/cli/src/cli.ts
```diff
@@ -1,15 +1,15 @@
-import { build } from "gluegun";
+import {build, GluegunToolbox} from 'gluegun';
```
dOrgJelli, line 1: Can we change the config to use `"` instead of `'`? Additionally, can we ensure all brackets have a space before and after values listed inside of them, for example: `{ build }` instead of `{build}`?

Both of these IMO help with readability, open to other opinions though.

dOrgJelli (changes requested):

packages/cli/src/cli.ts
mpetrunic: Yeah sure. Tbh I changed it because there was more instances of single quotes 😅

dOrgJelli (approved):
I'm alright with merging this into master, and I can deal with the merge conflicts on my end.

PR #88, opened 2020-12-28
CI and linting

mpetrunic:
Mostly cherry picked from master.
I will open separate PR with lint fixes so it's easier to review.
I've switch build to use lerna (it won't build packages marked private - test env and demos), client-js won't build but I guess it's known since it wouldn't build with old setup and in old setup `@web3api/core-js` isn't built at all.


package.json
```diff
@@ -27,7 +30,13 @@
...
   },
   "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "4.9.0",
+    "@typescript-eslint/parser": "4.9.0",
+    "eslint": "7.14.0",
+    "eslint-config-prettier": "6.15.0",
+    "eslint-plugin-import": "2.22.1",
+    "eslint-plugin-prettier": "3.2.0",
     "lerna": "3.22.1",
-    "rimraf": "3.0.2"
```
dOrgJelli, line 31: We need this for the clean command above

packages/cli/package.json
```diff
@@ -20,6 +21,8 @@
     "chalk": "4.1.0",
     "fs-extra": "9.0.1",
     "gluegun": "4.3.1",
+    "graphql": "15.3.0",
+    "graphql-tag": "2.11.0",
```
dOrgJelli, line 25: I don't think these two are needed since they'll be brought in from the @web3api/core-js package.

.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
+  "printWidth": 120,
+  "tabWidth": 2,
+  "useTabs": false,
+  "semi": true,
+  "singleQuote": true,
```
dOrgJelli, line 6: Based on the current code style, want to use double quotes `"` by default instead of single `'`. I prefer double for normal string as it's harder to get them mixed up with template string backticks ``` ` ```

dOrgJelli (approved):

dOrgJelli:
Looks great @mpetrunic, just requested some minor changes

packages/cli/package.json
mpetrunic: I would advise to always put dependency in package.json. I think we already have rule to enforce it (import/no-extraneous-dependencies).

From my experience with lodestar, it's quite easy to loose track of dependencies and publish code with missing dependency because code won't complain locally if this dependency is dev dependency in some other package or installed globally.

.prettierrc
mpetrunic: yeah, sure, seems like I failed to cherry pick that. Probably was commit with huge diff that was fixing linting

packages/cli/package.json
dOrgJelli: Got it, definitely agree. Isn't this not required here though since the cli project isn't directly importing "graphql" or "graphql-tag"?

mpetrunic: https://github.com/Web3-API/prototype/blob/prealpha-dev/packages/cli/src/commands/query.ts#L14

Gonna remove graphql as it's not used, I think that one came with cherry picked commit

dOrgJelli (approved):

Issue #89, opened 2021-01-04
Support Object Types as Method Arguments & Return Types


dOrgJelli:
Support:
```graphql
type Query {
  method(arg: CustomType!): CustomType!
}

type CustomType {
  ...
}
```

In order to implement this, we must (roughly speaking):
- Add support for this in the TypeInfo object model (schema-parse package)
- Add support in the extract/query-types visitor (schema-parse package)
- Decide what the generated code will look like (schema-bind package)
- Create a test case to make sure it works (schema-bind package)
- When importing a query type, also import all required object types that are used as arguments or return types (schema-compose package)

mpetrun5:
Hi,

new to this project and started working on parse end of this issue.

To add suport for Object Types can I add object field to `AnyDefinition` and thus to `PropertyDefinition` and then I handle the 3 cases (array, scalar, object) in mustache serialize/deserialize?

dOrgJelli:
Yes I think that sounds like the right approach @mpetrun5

dOrgJelli:
This is currently supported in both the Parse & Compose projects. Just need to implement Bind and we should be good to go :clap:

dOrgJelli:
Started implementing this here @mpetrun5:
https://github.com/Web3-API/prototype/tree/prealpha-dev_bind-nested-obj-types

PR #93, opened 2021-01-12
Parse object type in method argument and return 

mpetrun5:
Linked: #89


packages/schema/parse/src/extract/index.ts
```diff
@@ -9,7 +9,7 @@ export type SchemaExtractor = (astNode: DocumentNode, typeInfo: TypeInfo) => voi
 
 export const extractors: SchemaExtractor[] = [
   extractUserTypes,
-  extractQueryTypes,
   extractImportedObjectTypes,
+  extractQueryTypes,
```
mpetrun5, line 13: Is the order of extractors important for something else?
I need extractor to parse importedObjectTypes so I can recognize it is a object in `NamedType` parsing.

packages/schema/parse/src/extract/imported-query-types.ts
```diff
@@ -139,9 +141,18 @@ const visitorEnter = (typeInfo: TypeInfo, state: State) => ({
 
     if (method && argument) {
       // Argument value
-      argument.scalar = createScalarDefinition(
-        argument.name, modifier + node.name.value, state.nonNullType
-      );
+      if (
```
mpetrun5, line 144: Currently, imported query logic and query logic for extracing some types is the same so we should
probably extract that somewhere

packages/schema/parse/src/__tests__/cases/sanity/output.ts
```diff
@@ -86,6 +94,27 @@ export const output: TypeInfo = {
             createScalarPropertyDefinition("arg", "String", true)
           ],
           return: createScalarPropertyDefinition("queryMethod", "Int", true)
+        },
+        {
+          ...createMethodDefinition("query", "importedObjectMethod"),
+          arguments: [
+            createObjectPropertyDefinition(
```
mpetrun5, line 101: Do I need uri and namespace here for imported objects or I can just treat as regular user object?

mpetrun5 (commented):
@dOrgJelli 
General idea for parsing
Left a couple of questions for which I am not sure of

packages/schema/parse/src/__tests__/cases/sanity/output.ts
dOrgJelli: The UI & namespace are available on the TestImport_Object definition, so no need to carry those properties through to this instance.

dOrgJelli: So in short, all import metadata is only needed on the ImortedObjectDefinition, and not on the arguments + return type definitions.

packages/schema/parse/src/extract/index.ts
dOrgJelli: Ordering should be independent, as the extractors **shouldn't** require TypeInfo data from other extractors. This is of course an ideal scenario, and may break down in practice. Let me know if you there are dependencies between the extractors and we can figure out how to best handle this.

mpetrun5: Have dependecy here https://github.com/Web3-API/prototype/blob/fa6e10dad19e8b53b7c90117b3e87c88551aa4a7/packages/schema/parse/src/extract/query-utils.ts#L36

To recognize if it is object or scalar in named type extraction I need to check if the objects is defined somewhere
as I don't have `kind` from node.
Only way I found out was to check if that object exists in already extracted objects in `typeInfo`.

dOrgJelli: Chatted about this, will be handled in another PR.

Issue #65, opened 2020-11-13
Create a basic HTTP Request Client Plugin (JS)


Remscar:
A client plugin that will perform HTTP requests will be necessary for a variety of use cases. Since performing HTTP requests must be done in a language-specific context, it must be a client plugin.

To eliminate as much code smell as possible, the Graph QL schema for the HTTP Request Client Plugin should be as generalized as possible. If this client plugin gets utilized by Web3API's it will be assumed that the same GraphQL schema will be used on any Web3API platform.

At a minimum the following functionality should be supported:

- [ ] GET Request
- [ ] POST Request
- [ ] All requests will return appropriate data

Future iterations will add additional http verbs

PR #97, opened 2021-01-13
HTTP Request Client Plugin

MakMuftic:
### Changes
- Define graphql schema file
### Issues
Linked: #65


packages/js/plugins/http/src/schema.graphql
```diff
@@ -0,0 +1,30 @@
...
+type UrlParam {
+    key: String!
+    value: String!
+}
+
+type Response {
+    status: Integer!
+    statusText: String!
+    headers: [Header!]
+    body: String!
```
mpetrunic, line 15: We could have request error or timeout in which case all those stuff would be null and we would just return some error message

mpetrunic: Also body should probably be nullable?!

MakMuftic:
@dOrgJelli I opened the draft PR so you can comment on the initial schema.

I defined `body` of _request_ and _response_ as `String` with the idea to base64 encode bytes as a string.

packages/js/plugins/http/src/schema.graphql
mpetrunic: Another idea, make "get" and "post" ResponseType nullable since wasm module cannot do much with error message and maybe allow users to pass logger in plugin to log out error.

packages/js/plugins/http/src/schema.graphql
```diff
@@ -8,24 +8,31 @@ type UrlParam {
     value: String!
 }
 
+enum ResponseType {
+    TEXT
+    JSON
+    ARRAYBUFFER
+    BLOB
```
mpetrunic, line 15: should probably be just binary or text no?

PR #104, opened 2021-01-13
Schema sanitization and error handling

### LINKED ISSUE #61: Schema Sanitization
Ensure the following scenarios can be detected and properly reported:
- [x] Schema import namespace collisions
- [x] Type name collisions (post import + namespacing)
- [x] Unrecognized directives
- [x] Incorrect directive usage
- [x] Import directive must be on Schema or Mutation type
- [x] No underscores in user defined type names

namesty:


namesty (commented):
Schema import namespace collisions, Type name collisions (post import + namespacing) and No underscores in user defined type names are straightforward changes; however custom directives validation pose certain questions.

The current pattern is that validation is done first, and extraction is done second. When extraction is performed, we assume there are no errors because validation has already ran; but Typescript does not know this. Therefore, when extracting directive arguments we cannot get typesafety, we would have to do lots of type casting because we are humanly sure that validation has already ran and there are no risks by type casting. Should we be doing custom extractors for typesafety with validations? Or should we be keeping validation and extraction strictly separate and have the extractors be written with type casting?

My second issue is abstracting directive validation into a general validator capable of parsing directive definitions and validating argument values and types dynamically. GraphQL already does this to a certain extent. If a parameter is missing or the type is wrong, a `printSchemaWithDirectives` is enough for the vanilla GraphQL to validate the all schema directive usages and arguments and to show a semantic error. But there are some edge cases, that may not behave like we want (e.g. `[ ]` is valid for `[String!]!`, `["foo", false, 1]` is valid for `[String!]!`.

Also, some directives require custom validation that vanilla GraphQL will not validate (e.g. `@imports` can only be used on `Mutation` and `Query` type definitions). These custom validations have to be written manually for each case; so maybe it makes sense not to have a general dynamic all-directives validator, but rather specific validators for each case since custom directives are also written down manually, "deterministically", prior to their usage?

If we did develop something like a general all-directives validator, we would need to create a validator for each supported argument type, type and validator mappers, and a recursive algorithm to be able to traverse lists inside lists, custom object types, non-nullables and combinations of these; which would be sort of reimplementing vanilla GraphQL's validator, no?

In any case, I went with the following for now, until further discussion or comments:

- A directives map, which holds a key for each custom directive and has an arguments extractor with validations and a validate function to contemplate custom/special validation described above
- Running all validations prior to extraction. But also, when argument extractors are called separately/autonomously, validation is performed inside the extractor functions.

dOrgJelli:
@namesty 

> Or should we be keeping validation and extraction strictly separate and have the extractors be written with type casting?

I think the in-depth validation should happen as a first pass, with its own visitors. Then once the extractors run, they can have very simple validation (that will resolve typescript errors). It's nice to have 2 sets of validation, one thorough and one simple, because we won't want to run the thorough validation over APIs that have already been published. Thorough validation will be used during build-time scenarios, simple validation will be used during run-time scenarios. This will help minimize undefined behavior due to malformed schemas.

> My second issue is abstracting directive validation into a general validator capable of parsing directive definitions and validating argument values and types dynamically.

I think we need to make validators their own set of GraphQL visitors that run before the extraction visitors. No need to worry about type-correctness inside of the directive argument values, since GraphQL's parser will take care of that for us. We just want to make sure (1) only our supported directives are used (imports, imported), (2) and their arguments are correct.

> Also, some directives require custom validation that vanilla GraphQL will not validate (e.g. @imports can only be used on Mutation and Query type definitions). These custom validations have to be written manually for each case;

Yep you're correct, the validator visitor can take care of this and make sure the directive is in the right place.

> and a recursive algorithm to be able to traverse lists inside lists

visitor pattern fixes this.

PR #105, opened 2021-01-14
E2E Tests Functional

dOrgJelli:
- CLI Refactored & Functional
- Client Refactored & Functional
- Client Automated Testing
- Various bug fixes
- SimpleStorage CLI demo functional
- SimpleStorage dApp almost functional (web-worker source file isn't being bundled correctly)

To test locally:
- Clone branch
- `nvm use` && `yarn` at root
- `yarn build` to watch all packages be built
- `yarn test` to watch all tests be built
- Head over to `./demos/simple-storage/README.md` and follow the instructions for the CLI demo


packages/js/client/src/wasm/WasmWeb3Api.ts
```diff
@@ -0,0 +1,377 @@
...
+    if (!data) {
+      throw Error(
+        `Module was not found.\nURI: ${this._uri}\nSubpath: ${moduleManifest.module.file}`
+      );
+    }
+
+    this._wasm[module] = data;
+    return data;
+  }
+}
```
mpetrunic, line 377: not sure what's going on here or why there is manual data transfer but I found working with treads is much simpler with https://threads.js.org/

packages/schema/bind/package.json
```diff
@@ -4,10 +4,11 @@
   "license": "MIT",
   "main": "build/index.js",
   "scripts": {
-    "build": "rimraf ./build && tsc --project tsconfig.build.json",
+    "build": "rimraf ./build && tsc --project tsconfig.build.json && yarn copy:templates",
     "lint": "eslint --color --ext .ts src/",
     "test": "jest --passWithNoTests --runInBand --verbose",
-    "test:watch": "jest --watch --passWithNoTests --verbose"
+    "test:watch": "jest --watch --passWithNoTests --verbose",
+    "copy:templates": "copyfiles -u 1 src/**/*.mustache build/"
```
mpetrunic, line 11: so annoying that tsc doesn't have copy files option :(

mpetrunic: I wonder if it would be better to keep them out of src

mpetrunic (commented):
Nice work

morrigan (commented):
Awesome! However, would love to see CI passing as I'm having issues running all tests successfully.

Also, `.env` file should be git ignored in `test-env` :)

packages/js/client/src/wasm/WasmWeb3Api.ts
dOrgJelli: The first implementation was actually using Threads.js, but I had to switch since it doesn't support passing SharedArrayBuffer as a worker data argument. Here's the commit where I made the transition: https://github.com/Web3-API/prototype/pull/105/commits/855ea50f83428a910411685401d2c45637a63b20

dOrgJelli: What is going on here is:
- Create a thread for the WASM module we're about to call into
- The thread can dispatch actions for the main thread to do
- When the thread dispatches a "sub-invoke" action, it immediately pauses itself to stop the WASM VM from continuing execution. The main thread then performs the sub-invocation, passes the data back into the thread using a SharedArrayBuffer, and the thread continues its execution given its sub-invoke result.

package.json
```diff
@@ -8,25 +8,18 @@
...
       "**/@testing-library/*",
       "**/@types/testing-library*"
     ]
   },
   "scripts": {
-    "build": "lerna run build --no-private",
-    "build:cli": "yarn workspace @web3api/cli build",
-    "build:client-js": "yarn workspace @web3api/client-js build",
-    "build:wasm-as": "yarn workspace @web3api/wasm-as build",
+    "build": "lerna run build --no-private --ignore @web3api/cli && yarn build:cli",
```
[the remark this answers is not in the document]
dOrgJelli: The reason for this is the fact that there's a circular dependency between @web3api/cli and @web3api/client-js. So to fix this, we build the client first, since it only requires @web3api/cli as a devDependency

package.json
```diff
@@ -8,25 +8,18 @@
...
     "lint": "lerna run lint",
     "lint:fix": "lerna run lint -- --fix",
     "lint:ci": "lerna run --parallel lint",
-    "test": "yarn test:wasm-as && yarn test:client-js",
-    "test:wasm-as": "yarn workspace @web3api/wasm-as test",
-    "test:client-js": "yarn workspace @web3api/client-js test",
-    "test:ci": "yarn test:wasm-as && yarn test:ci:client-js",
-    "test:ci:client-js": "yarn workspace @web3api/client-js test:ci",
+    "test": "lerna run test --no-private",
+    "test:ci": "lerna run test:ci --no-private",
```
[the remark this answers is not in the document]
dOrgJelli: Awesome! Everything runs top to bottom in ~30 seconds now :D

packages/cli/package.json
```diff
@@ -7,17 +7,17 @@
...
-    "@textile/threads-core": "^0.3.1",
     "@web3api/client-js": "0.0.1-prealpha.1",
     "@web3api/client-test-env": "0.0.1-prealpha.1",
-    "assemblyscript": "0.14.7",
+    "@web3api/schema-compose": "0.0.1-prealpha.1",
+    "@web3api/schema-bind": "0.0.1-prealpha.1",
+    "@web3api/ens-plugin-js": "0.0.1-prealpha.1",
+    "@web3api/ipfs-plugin-js": "0.0.1-prealpha.1",
+    "@web3api/ethereum-plugin-js": "0.0.1-prealpha.1",
+    "assemblyscript": "0.17.13",
```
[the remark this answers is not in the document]
dOrgJelli: Yep agreed, this is because the CLI compiles your Assemblyscript modules for you. See here: https://github.com/Web3-API/prototype/blob/c090e72b20466f3caf41a86678215a8b551bf3cd/packages/cli/src/lib/Compiler.ts#L11

In order to make the build pipeline implementation language (and compiler) agnostic, we'd need to setup a standard for projects to follow for defining their build steps. This work is starting to be tracked in these two issues:
https://github.com/Web3-API/prototype/issues/95
https://github.com/Web3-API/prototype/issues/25

packages/schema/bind/package.json
dOrgJelli: Agreed, would be super helpful to just have this as a tsconfig option.

I'd prefer to keep the templates along-side the source, to help with readability of the code-base. I'm also considering making all `.mustache` files JS modules that export a string. This way file system access is required at run-time.

dOrgJelli:
Yep thanks for pointing that out @morrigan, going to leave it in the tree for now so that everyone doesn't have to configure this themselves upon cloning the code-base. These types of env-config options will handled when this issue is spec'd out: https://github.com/Web3-API/prototype/issues/95

package.json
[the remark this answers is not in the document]
mpetrunic: Might be wrong but I'm pretty sure lerna prioritizes dep over devDep and warny about circular

packages/js/client/src/wasm/WasmWeb3Api.ts
mpetrunic: Tnx for explanation. But isn't `Transfer(ArrayBuffer)` basically SharedArrayBuffer: https://threads.js.org/usage-advanced#transferable-objects ?

dOrgJelli: No this will not work since you need to have a valid reference to the ArrayBuffer on both ends (main-thread + wasm-thread). This is required so that you can use `Atomics.wait(...)` within the wasm-thread, and `Atomics.notify(...)` in the main-thread. Both of these require a reference to a SharedArrayBuffer.

package.json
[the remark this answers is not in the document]
dOrgJelli: Ya I thought that would be the case too. It looks like it is, but the fact that we're deleting the build directory with `rimraf` before compiling with `tsc` causes the build to fail like so:
```
lerna info run Ran npm script 'build' in '@web3api/client-js' in 5.3s:
$ rimraf ./build && tsc --project tsconfig.build.json
lerna ERR! yarn run build exited 2 in '@web3api/cli'
lerna ERR! yarn run build stdout:
$ rimraf ./build && tsc
src/commands/query.ts(10,49): error TS2307: Cannot find module '@web3api/client-js' or its corresponding type declarations.
src/lib/Compiler.ts(12,59): error TS2307: Cannot find module '@web3api/client-js' or its corresponding type declarations.
src/lib/Web3APIManifest.ts(3,26): error TS2307: Cannot find module '@web3api/client-js' or its corresponding type declarations.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

lerna ERR! yarn run build stderr:
error Command failed with exit code 2.
```

The use of `rimraf` before `tsc` is required so that we can remove old build artifacts that may not exist anymore. Here's an issue where this is talked about: https://github.com/microsoft/TypeScript/issues/13722

I'll look into how this can fixed. Maybe by running to yarn commands instead of using the inline `&&` to concatenate the CLI commands. I think what's happening is that lerna when it first runs takes a snapshot of all the files it needs and stores them in memory. These snapshotted file descriptors may become invalid after a deletion occurs.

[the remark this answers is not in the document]
dOrgJelli: No luck :P will revisit this in the future, but for now we'll just try our best to avoid circular build dependencies.

PR #114, opened 2021-01-21
Add http client plugin

### LINKED ISSUE #65: Create a basic HTTP Request Client Plugin (JS)

MakMuftic:
**Add HTTP client plugin**

### Changes

- Add plugin schema
- Implement `HttpPlugin` and resolver
- Add unit tests 

### TODO before closing PR
- ~~Add e2e tests~~

### Issues

Closes #65


packages/js/plugins/http/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "license": "MIT",
+    "main": "build/index.js",
+    "scripts": {
+      "build": "rimraf ./build && tsc",
+      "lint": "eslint --color --ext .ts src/",
+      "test": "jest --passWithNoTests --runInBand --verbose",
+      "test:watch": "jest --watch --passWithNoTests --verbose"
+    },
+    "dependencies": {
+      "@web3api/client-js": "0.0.1-prealpha.1",
```
mpetrunic: dev dependency?

packages/js/plugins/http/package.json
```diff
@@ -0,0 +1,25 @@
...
+      "@web3api/core-js": "^0.0.1-prealpha.1",
+      "axios": "0.21.1"
+    },
+    "devDependencies": {
+      "@types/jest": "26.0.8",
+      "jest": "26.2.2",
+      "rimraf": "3.0.2",
+      "ts-jest": "26.1.4",
+      "ts-node": "8.10.2",
+      "typescript": "3.9.7"
```
mpetrunic: @dOrgJelli Maybe we can move all those common dev dependencies to root package.json?

packages/js/plugins/http/src/__tests__/unit/index.test.ts
```diff
@@ -0,0 +1,178 @@
+import axios, { AxiosResponse, AxiosRequestConfig } from "axios"
+import {HttpPlugin} from "../../index"
+
+jest.mock('axios')
```
mpetrunic: don't we need to restore actual axios implementation on after

packages/js/plugins/http/src/__tests__/unit/index.test.ts
```diff
@@ -0,0 +1,178 @@
...
+
+describe('test http plugin get method', () => {
+    
+    let plugin: HttpPlugin = new HttpPlugin();
+
+    test('valid request: text response type', async() => {
+        // @ts-ignore
```
mpetrunic: We should be more typesafe:
```typescript
        const mockedAxios = jest.requireMock<jest.Mocked<typeof axios>>('axios');
        mockedAxios..get.mockResolvedValueOnce...
```

packages/js/plugins/http/tsconfig.json
```diff
@@ -0,0 +1,33 @@
+{
```
mpetrunic, line 1: extend from base config instead of duplicating

mpetrunic (changes requested):

packages/js/plugins/http/tsconfig.json
MakMuftic: @mpetrunic I am not sure on what base config do you think. I could extend from `js/core` ? , with something like this inside _tsconfig.json_:
```json
{
  "extends": "./../core/tsconfig",
}
```

mpetrunic: From root of repository

MakMuftic: But there is no root `tsconfig` file. Probably we should add one.

mpetrunic: oh heck, yeah. It should probably be separate issue

MakMuftic: I opened a separate issue for this #116

packages/js/plugins/http/package.json
dOrgJelli: I'd rather keep them here so that we can move these packages out of the monorepo in the future w/o worrying about finding all deps.

packages/js/plugins/http/README.md
```diff
@@ -0,0 +1 @@
+TODO
```
mpetrunic: @MakMuftic khm

packages/js/plugins/http/package.json
```diff
@@ -0,0 +1,27 @@
+{
+    "name": "@web3api/http-plugin-js",
```
mpetrunic: add description field

dOrgJelli (approved):
Amazing work all :clap:

mpetrunic:
@dOrgJelli Should we open issue to add linting/formatting to graphql and json files? :)

dOrgJelli:
Yes definitely, good call @mpetrunic, will create now

Issue #106, opened 2021-01-20
Remove Schema Parse Extractor Dependencies


dOrgJelli:
Currently schema parse extractors have dependencies between eachother. In order to eliminate this, we've think we can:
- Better define what the supported scalar types are. This will allow a single predicate to be called to see what supported scalar types there are (collection of supported types in parse package). Then we can have mappings to MsgPack types in Bind (throw if no map).

This will hopefully eliminate the deps between extractors.

dOrgJelli:
This has been completed

PR #115, opened 2021-01-21
Parse nested objects and arrays

mpetrun5:
Changes: 
 - add support for parsing nested objects
 - add support for objects in arrays 
 - remove typeInfo dependecy from query types parsing
 - defined currently supported scalars
 - matched imported object and user object Schema and extracted code to object-types-utils

TODO:
 - add imported object type property parsing to query and object extraction (@dOrgJelli)
 - Bind and Compose I will do in separate PRs after we finish parse

Issues: 
 - Closes: #106


dOrgJelli:
Looks great @mpetrun5 :muscle: :sponge: 

I'll create another PR that allows us to check if a property is "imported". A naive way to do this is to check the type name for an '_' character, but I dislike this as it relies on the sanitization code not allowing '_' in user object type names. Instead I think we should leverage the the `@imports` directive a bit more. I'll make another PR with this today.

PR #118, opened 2021-01-26
Refactor TypeInfo + Use TypeInfo Visitor for Ext Type Resolution

dOrgJelli:
Things that were refactored:
- TypeInfo's `name` and `type` property weren't being used correctly. Now `type` will always be set, and `name` may not be set if for example you're defining a new ObjectDefinition (ex: `type CustomType { ... }`).
- Type names no longer contain the `required` modifier within them. Previously, an optional type was prefixed with a "?" within the type name. This ended up creating more problems than it solved. Instead we will only use the `required` property on TypeDefinitions.
- The Schema Composer's resolve logic has been refactored to use the TypeInfo visitor / transformer.

Additional changes:
- Added support for "Boolean" GraphQL type.
- Removed unnecessary GraphQL Schema parsing within WasmWeb3Api invoke function.


packages/schema/bind/src/__tests__/cases/sanity/input.graphql
```diff
@@ -33,6 +33,8 @@ type CustomType {
   i16: Int16!
   i32: Int32!
   i64: Int64!
+  boolean: Boolean!
+  optBoolean: Boolean
```
mpetrun5, line 37: Minor change
But isn't this the other way around

packages/schema/parse/src/typeInfo/scalar.ts
```diff
@@ -0,0 +1,35 @@
+interface SupportedScalars {
```
mpetrunic: Instead of duplicating. We can probably write something like
```
export type SupportedScalars = typeof testInstance;
```

mpetrun5 (commented):
@dOrgJelli 
Looks fine to me.
What's the status of imported objects currently?
I know we talked about using @import (?) directive to handle recognizing them, but don't know
what is decided. Will probably need that for extracting dependent objects of type for imports
in schema bind.

packages/schema/parse/src/typeInfo/scalar.ts
dOrgJelli: Ah very good point

packages/schema/bind/src/__tests__/cases/sanity/input.graphql
dOrgJelli: No, in GraphQL you use “!” to show that something is required. If you don’t have a “!” that means it’s optional.

mpetrun5: :joy: I totally blanked there

dOrgJelli:
Hey @mpetrun5, it should be unblocked. I'll do a little investigation to make sure this is possible in another branch. Will DM you.

dOrgJelli:
I ended up not having to expand upon the `@imports` & `@imported` directives, which I think is for the best.

PR #120, opened 2021-01-26
Support Object Types as Schema Arguments - Bind Implementation

### LINKED ISSUE #70: Improve MsgPack Deserialization Sanitization
Currently, when msgpack (in WASM or in JS) deserializes a buffer, it does not have sanitization logic for handling unexpected types or malformed data. We need this to happen at run-time because we don't want to run the risk of malformed input data getting sent to a smart contract, making funds at risk.

### LINKED ISSUE #89: Support Object Types as Method Arguments & Return Types

### LINKED ISSUE #134: WASM Bind: Deserialization Throws When Required Props Are Missing
Make sure that all deserialization methods throw errors when there are missing properties. Currently we only do this for method arguments using the `_propSet: boolean = false` local variable.

dOrgJelli:
closes https://github.com/Web3-API/prototype/issues/89
closes https://github.com/Web3-API/monorepo/issues/134
closes https://github.com/Web3-API/monorepo/issues/70

- [x] Parse
- [x] Compose
- [x] Bind
- [x] Sanitize (Detect infinitely recursive relationships)
- [x] Tests

Detecting the infinitely recursive object relationships is currently implemented at the WASM deserialization level, but we still need to implement GraphQL schema sanitization logic to stop developers from implementing this in the first place. I've created this issue which describes this in more detail: https://github.com/Web3-API/monorepo/issues/149


packages/js/client/src/wasm/WasmWeb3Api.ts
```diff
@@ -56,7 +56,9 @@ export class WasmWeb3Api extends Api {
     // TODO: come up with a better future-proof solution
     while (threadsActive >= maxThreads) {
       // Wait for another thread to become available
-      await new Promise((resolve) => setTimeout(() => resolve(), 500));
+      await new Promise((resolve: (value?: unknown) => void) =>
+        setTimeout(() => resolve(), 500)
+      );
```
mpetrunic: you are using this in couple of places. Could be extracted to "sleep(ms: number)" utility method

packages/js/client/src/wasm/WasmWeb3Api.ts
```diff
@@ -91,112 +93,124 @@ export class WasmWeb3Api extends Api {
...
+            // Notify the thread that we've sent data, giving it a specific
+            // status code
+            Atomics.store(threadMutexes, threadId, status);
+            Atomics.notify(threadMutexes, threadId, Infinity);
+
+            // Wait until the transferPending flag has been reset
+            while (transferPending) {
+              await new Promise((resolve: (value?: unknown) => void) =>
+                setTimeout(() => resolve(), 100)
+              );
```
mpetrunic: Can be replaced with sleep from previous comment

mpetrun5 (commented):
:rocket:

demos/simple-storage/protocol/recipes/constants.json
```diff
@@ -1,3 +1,3 @@
 {
-  "SimpleStorageAddr": "0x2c5f3c004878923f55A2a255F89Fe29393177509"
+  "SimpleStorageAddr": "0xe78A0F7E598Cc8b0Bb87894B0F60dD2a88d6a8Ab"
```
benefacto, line 2: Should this change be included? It may not matter too much since the file is modified with every run.

packages/js/client/src/wasm/WasmWeb3Api.ts
dOrgJelli: Will do :D

demos/simple-storage/protocol/recipes/constants.json
dOrgJelli: yah doesn't really affect anything :P so it's fine

PR #126, opened 2021-01-28
Schema bind for objects, nested objects and objects inside arrays

mpetrun5:
Changes: 
 - make nullable objects be Object | null instead of Nullable<Object> mostly because typescript really doesn't like Nullable and makes it hard to check code
 - add missing return types on deserialize and serialize query methods
 - add support for objects, nested objects and objects in array for imported and user objects and querie

TODO: 
 - circular objects
 - ~~broke something in e2e tests which I need to check~~

### DIFF
--- a/packages/schema/bind/src/__tests__/cases/sanity/output/wasm-as/Query/wrapped.ts
+++ b/packages/schema/bind/src/__tests__/cases/sanity/output/wasm-as/Query/wrapped.ts
@@ -1,9 +1,12 @@
 import {
-  queryMethod
+  queryMethod,
+  objectMethod
 } from "../../index";
 import {
   deserializequeryMethodArgs,
-  serializequeryMethodResult
+  serializequeryMethodResult,
+  deserializeobjectMethodArgs,
+  serializeobjectMethodResult
 } from "./serialization";
 
 export function queryMethodWrapped(argsBuf: ArrayBuffer): ArrayBuffer {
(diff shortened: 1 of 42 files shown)


packages/schema/bind/src/__tests__/cases/sanity/output/wasm-as/Query/wrapped.ts
```diff
@@ -1,9 +1,12 @@
 import {
-  queryMethod
+  queryMethod,
+  objectMethod
```
mpetrun5, line 3: This doesn't make sense currently and are unresolved imports
Will look through diff to find what was changed

mpetrun5:
Fixed e2e test
It seems that assemblyscript compiler goes into an infinite loop when 
`import * as Objects from "../"` is used instead of 
`import * as Objects from ".."`

dOrgJelli:
Great PR @mpetrun5! I think we should solve circular dependencies by creating a sanitizer that ensures schemas do not implement an infinitely recursive circular dependency, like this:
```
type A {
  prop: B!
}

type B{
  prop: A!
}
```

The above example is infinitely recursive because they're both non-nullable. This is acceptable however:
```
type A {
  prop: B!
}

type B {
  prop: A
}
```

dOrgJelli:
I'm going to merge this PR, and implement the sanitization logic in the prealpha-dev_bind-nested-bjects branch + implement an example to show that this is working e2e

PR #132, opened 2021-01-29
Codegen Command

vhurryharry:
https://github.com/Web3-API/prototype/issues/101
https://github.com/Web3-API/prototype/commit/4717b0fe01deee348f3279050f9735a036e2113e

```
w3 codegen [<template-file>] [options]

Template file:
  Path to input template file (default: web3api.gen.js)

Options:
  -h, --help                              Show usage information
  -m, --manifest-path <path>              Path to the Web3API manifest file (default: web3api.yaml)
  -i, --ipfs [<node>]                     IPFS node to load external schemas (default: dev-server's node)
  -o, --output-dir <path>                 Output directory for generated types (default: types/)
  -e, --ens [<address>]                   ENS address
```

Example `web3api.gen.js` file.
```javascript
exports.run = (output, config) => {
  output.entries.push({
    type: "File",
    name: "./schema1.ts",
    data: config.generate("./templates/schema-ts.mustache", config.typeInfo),
  });

  output.entries.push({
    type: "Directory",
    name: "./folder",
    data: [
      {
        type: "File",
        name: "./schema2.ts",
        data: config.generate(
          "./templates/schema-ts.mustache",
          config.typeInfo
        ),
      },
    ],
  });

  output.entries.push({
    type: "Template",
    name: "./schema3.ts",
    data: "./templates/schema-ts.mustache",
  });
};
```


packages/cli/src/commands/codegen.ts
```diff
@@ -0,0 +1,114 @@
+/* eslint-disable prefer-const */
+import { fixParameters } from "../lib/helpers/parameters";
+import { CodeGenerator } from "../lib/generators/CodeGenerator";
+
+import chalk from "chalk";
+import axios from "axios";
+import { GluegunToolbox } from "gluegun";
+
+const defaultTemplate = "web3api.gen";
```
dOrgJelli: Shouldn't this be `web3api.gen.js`?

packages/cli/src/commands/codegen.ts
```diff
@@ -0,0 +1,114 @@
...
+const defaultManifest = "web3api.yaml";
```
dOrgJelli: We want to also support (by default) `web3api.yml`

packages/cli/src/commands/codegen.ts
```diff
@@ -0,0 +1,114 @@
...
+
+Template file:
+  Path to input template file (default: ${defaultTemplate})
+
+Options:
+  -h, --help                              Show usage information
+  -m, --manifest-path <path>              Path to the Web3API manifest file (default: ${defaultManifest})
+  -i, --ipfs [<node>]                     IPFS node to load external schemas (default: dev-server's node)
+  -o, --output-dir <path>                 Output directory for generated types (default: types/)
+  -e, --ens [<address>]                   ENS address
```
dOrgJelli: Let's have a better description here, maybe: "ENS address to lookup external schemas (default: 0x0000...2e1e)

packages/cli/src/commands/codegen.ts
```diff
@@ -0,0 +1,114 @@
...
+import { GluegunToolbox } from "gluegun";
+
+const defaultTemplate = "web3api.gen";
+const defaultManifest = "web3api.yaml";
+
+const HELP = `
+${chalk.bold("w3 codegen")} ${chalk.bold("[<template-file>]")} [options]
+
+Template file:
+  Path to input template file (default: ${defaultTemplate})
```
dOrgJelli: I think naming this "template file" is a bit confusing. Maybe calling web3api.gen.js the "Generation File" instead?

packages/cli/src/commands/codegen.ts
```diff
@@ -0,0 +1,114 @@
...
+    let ipfsProvider: string | undefined;
+    let ethProvider: string | undefined;
+    let ensAddress: string | undefined = ens;
+
+    if (typeof ipfs === "string") {
+      // Custom IPFS provider
+      ipfsProvider = ipfs;
+    } else if (ipfs) {
+      // Dev-server IPFS provider
+      // TODO: handle the case where the dev server isn't found
```
dOrgJelli: This can be handled by just try, catching the axios.get below this line.

packages/cli/src/lib/Compiler.ts
```diff
@@ -5,23 +5,20 @@
...
-import { EthereumPlugin } from "@web3api/ethereum-plugin-js";
-import { IpfsPlugin } from "@web3api/ipfs-plugin-js";
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports
 const fsExtra = require("fs-extra");
 // eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports
 const toolbox = require("gluegun/toolbox");
 
-export interface CompilerConfig {
+export interface BuildConfig {
```
dOrgJelli: Nit: I like using the class name + "Config", so viewers can immediately see what the interface is for (configuring the class via the constructor). This naming convention is used throughout the rest of the code as well. So long story short, let's rename this back to "CompilerConfig"

packages/cli/src/lib/SchemaComposer.ts
```diff
@@ -0,0 +1,152 @@
...
+import { Manifest, Uri, Web3ApiClient, UriRedirect } from "@web3api/client-js";
+import { composeSchema, ComposerOutput } from "@web3api/schema-compose";
+import { EnsPlugin } from "@web3api/ens-plugin-js";
+import { EthereumPlugin } from "@web3api/ethereum-plugin-js";
+import { IpfsPlugin } from "@web3api/ipfs-plugin-js";
+
+export class SchemaComposer {
+  constructor(private _config: BuildConfig) {}
+
+  public async loadManifest(quiet = false): Promise<Manifest> {
```
dOrgJelli: This can be made into its own helper function that takes the `manifestPath: string` as an argument. This will be helpful in making these classes smaller. I also don't think it makes sense to have "loadManifest" as a method on the SchemaComposer, seems like this mixes responsibilities a bit.

packages/cli/src/lib/Compiler.ts
```diff
@@ -31,15 +28,17 @@ export interface CompilerConfig {
...
+  constructor(private _config: BuildConfig) {
     this._manifestDir = path.dirname(_config.manifestPath);
+    this._schemaComposer = new SchemaComposer(_config);
   }
 
   public async compile(quiet?: boolean, verbose?: boolean): Promise<boolean> {
     try {
       // Load the manifest
-      const manifest = await this._loadManifest();
+      const manifest = await this._schemaComposer.loadManifest();
```
dOrgJelli: In this file, _loadManifest(...) still exists. It is also duplicated in the SchemaComposer. I left a comment in SchemaComposer saying that I think it should be made into its own function to help make these classes smaller.

packages/cli/src/lib/SchemaComposer.ts
```diff
@@ -0,0 +1,152 @@
...
+import fs from "fs";
+import path from "path";
+import { Manifest, Uri, Web3ApiClient, UriRedirect } from "@web3api/client-js";
+import { composeSchema, ComposerOutput } from "@web3api/schema-compose";
+import { EnsPlugin } from "@web3api/ens-plugin-js";
+import { EthereumPlugin } from "@web3api/ethereum-plugin-js";
+import { IpfsPlugin } from "@web3api/ipfs-plugin-js";
+
+export class SchemaComposer {
+  constructor(private _config: BuildConfig) {}
```
dOrgJelli: Instead of using the `BuildConfig` here, let's create a new interface called `SchemaConfig` and only include the properties you need for this class' methods.

packages/cli/src/lib/generators/CodeGenerator.ts
```diff
@@ -0,0 +1,104 @@
...
+import {
+  OutputDirectory,
+  OutputEntry,
+  writeDirectory,
+} from "@web3api/schema-bind";
+import { TypeInfo, parseSchema } from "@web3api/schema-parse";
+
+export class CodeGenerator {
+  private _schema: string | undefined = "";
+  constructor(private _templateFile: string, private _config: BuildConfig) {}
```
dOrgJelli: Let's use a `CodeGeneratorConfig` here that only has the specific properties required for this class' methods.

dOrgJelli: With this we can make "templateFile" apart of the config.

dOrgJelli: Also don't forget to rename "templateFile" to "GenerationFile" or maybe "GenerationRunner", depending on what you choose.

dOrgJelli: GenerationScript?

packages/cli/src/lib/generators/CodeGenerator.ts
```diff
@@ -0,0 +1,104 @@
...
+    // Make sure that the output dir exists, if not create a new one
+    if (!fs.existsSync(this._config.outputDir)) {
+      fs.mkdirSync(this._config.outputDir);
+    }
+
+    // Compose schema from manifest
+    const schemaComposer = new SchemaComposer(this._config);
+    const manifest = await schemaComposer.loadManifest();
+    const composedSchema = await schemaComposer.composeSchemas(manifest);
+    const typeInfo = parseSchema(composedSchema.combined!);
```
dOrgJelli: Why the `!`?

packages/cli/src/lib/generators/CodeGenerator.ts
```diff
@@ -0,0 +1,104 @@
...
+    this._schema = composedSchema.combined;
+
+    // Check the template file if it has the proper run() method
+    // eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports, @typescript-eslint/naming-convention
+    const { run } = await require(this._templateFile);
+
+    if (!run) {
+      console.log(
+        chalk.red(
+          "The template file provided is wrong or doesn't have the 'run' method."
```
dOrgJelli: It would be great to separate these two errors so the user knows exactly what's wrong. 1 error when we can't find the file, 1 error when we can't find the run function.

packages/cli/src/lib/generators/CodeGenerator.ts
```diff
@@ -0,0 +1,104 @@
...
+      return false;
+    }
+
+    const output: OutputDirectory = {
+      entries: [],
+    };
+
+    await run(output, {
+      typeInfo,
+      generate: (templatePath: string, typeInfo: TypeInfo) =>
```
dOrgJelli: Instead of the second property being `typeInfo`, we want this to be a generic object so they can pass in whatever they want (might only want a single ObjectTypeDefinition for example). So let's set it to `config: unknown`.

packages/cli/src/lib/generators/CodeGenerator.ts
```diff
@@ -0,0 +1,104 @@
...
+        this.generate(templatePath, typeInfo),
+    });
+
+    output.entries = await Promise.all(
+      output.entries.map((entry) => this.generateFile(entry, typeInfo))
+    );
+
+    writeDirectory(this._config.outputDir, output);
+
+    console.log(`🔥 Types were generated successfully 🔥`);
```
dOrgJelli: Instead of doing console.log, let's keep consistent with the rest of the logging using the gluegun toolbox logging functions. Maybe using the `withSpinner`

packages/cli/src/lib/generators/CodeGenerator.ts
```diff
@@ -0,0 +1,104 @@
...
+    return true;
+  }
+
+  public generate(templatePath: string, typeInfo: TypeInfo): string {
+    templatePath = path.join(
+      path.dirname(this._config.manifestPath),
```
dOrgJelli: We'll want to use the dirname of the web3api.gen.js file here instead

dOrgJelli (changes requested):

dOrgJelli:
Would also love to see some tests

packages/cli/src/commands/codegen.ts
vhurryharry: So if `web3api.yaml` doesn't exist, then use `web3api.yml` file?

packages/cli/src/lib/generators/CodeGenerator.ts
vhurryharry: GenerationFile looks good. Sth like `GeneratorScript` would be more suitable for `.gen.js` files, but it wouldn't make sense for `.gen.yaml` files.

packages/cli/src/lib/generators/CodeGenerator.ts
vhurryharry: Because `parseSchema` doesn't allow null values. 😣

dOrgJelli: I think we should add an if check here, and throw an error if combined is falsey. Also it might be worth revisiting the schema/compose/src/index.ts file to see why combined can be undefined, is that an option of that function? Might make sense to just make it always return a string and not string | undefined.

packages/cli/src/commands/codegen.ts
dOrgJelli: Yes, we need to check for both since both extensions are used heavily: https://stackoverflow.com/questions/21059124/is-it-yaml-or-yml

https://github.com/search?q=.yml&type=code

packages/cli/src/lib/generators/CodeGenerator.ts
vhurryharry: Then inside the template, do you want to use sth like `{{config.objectTypes}}`?

dOrgJelli: Yes you could do that. The reason for having `config: unknown` is to support use-cases like this:
`web3api.gen.js`
```
const singleType = typeInfo.objectTypes.find(
  (type) => type.name === "MySpecialType"
);

generate("./SpecialTemplate.mustache", singleType);
```
`SpecialTemplate.mustache`
```
{{name}}: {{type}}
```

dOrgJelli (approved):

PR #146, opened 2021-02-09
React Wrapper

### LINKED ISSUE #127: Web3API Client React Wrapper
Create a react wrapper around the Web3API JS Client, which enables to easily initialize the client & send queries to Web3APIs throughout their application.

We'd want to support both a Component & Hook interface. The Web3API Client will be created using a React provider, and passed through to all queries as a context.

A possible developer experience might look like:
```jsx
// Component Interface
<Web3ApiClient redirects={...}>
  <MyApp />
</Web3ApiClient>

// elsewhere in MyApp
<Web3ApiQuery uri={...} query={...} variables={...}>
{(result: QueryResult) => {
  // Do something with the result
}}
</Web3ApiQuery>
```
```jsx
// Hook Interface
// Web3ApiClient as we did above...

const { loading, error, data } = useWeb3ApiQuery({ uri, query, variables });
```

cbrzn:
closes: #127


packages/js/client/src/declaration.d.ts
```diff
@@ -0,0 +1 @@
+declare module "spawn-command";
```
mpetrunic: It would probably make sense to also declare method interface

packages/js/react/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "@testing-library/react": "^11.2.5",
+    "@testing-library/react-hooks": "^5.1.0",
+    "@types/jest": "26.0.8",
+    "jest": "26.2.2",
+    "jest-environment-jsdom": "^26.0.1",
+    "rimraf": "3.0.2",
+    "typescript": "^4.1.3"
+  },
+  "peerDependencies": {
+    "@react": ">=16.13.1"
```
mpetrunic: maybe we can relax this to accept all 16.x versions?

packages/js/client/src/declaration.d.ts
cbrzn: Okok agreed

packages/js/react/package.json
cbrzn: Sure :-D

morrigan (dismissed):
Looks great!

demos/simple-storage/dapp/package.json
```diff
@@ -35,6 +33,10 @@
     "prestart": "yarn install:web3api",
     "install:web3api": "npx web3api-worker-install ./public"
   },
+  "resolutions": {
+    "@types/react": "^16.9.0",
+    "@types/react-dom": "^16.9.0"
+  },
```
dOrgJelli, line 39: Just curious, why is this needed?

packages/js/react/src/__tests__/env/setup.js
```diff
@@ -0,0 +1 @@
+global.XMLHttpRequest = undefined;
```
dOrgJelli, line 1: Why is this needed?

packages/js/react/src/__tests__/env/index.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * A custom environment to set the TextEncoder that is required by ../client/src/wasm/WasmWeb3Api.ts.
+ */
+module.exports = class CustomTestEnvironment extends Environment {
+  async setup() {
+    await super.setup();
+    if (typeof this.global.TextEncoder === "undefined") {
+      const { TextEncoder } = require("util");
+      this.global.TextEncoder = TextEncoder;
```
dOrgJelli, line 11: Why is this needed?

dOrgJelli (dismissed):

demos/simple-storage/dapp/package.json
cbrzn: IIRC: since we are doing this is a monorepo, when we added the react dependencies in the react package, it wasn't working because it was taking the versioning from this project (dapp) so we had to create this resolutions attribute which tells yarn to use this versioning in this project

packages/js/react/src/__tests__/env/setup.js
cbrzn: We were having a cors error in the tests, this fixed it

packages/js/react/src/__tests__/env/index.ts
cbrzn: It was saying that TextEncoder was undefined in the tests, this fixed it

demos/simple-storage/dapp/package.json
dOrgJelli: Ah okay, makes sense

packages/js/react/src/__tests__/env/index.ts
dOrgJelli: Ah okay sounds good. Probably just a jest browser testing issue. Nothing to worry about in production, as the demo dApp works just fine.

dOrgJelli (approved):

PR #147, opened 2021-02-09
Enum parse and compose support

mpetrun5:
Changes:
  - add enum and imported enum types to definitions
  - add type definition extractor before extractors to differentiate between enums and objects
  - add enum to validation
  - add enum extractor
  - add enum to compose object type extractor


packages/schema/parse/src/extract/imported-object-types.ts
```diff
@@ -45,7 +47,7 @@ const visitorEnter = (
     state.nonNullType = true;
   },
   NamedType: (node: NamedTypeNode) => {
-    extractNamedType(node, state);
+    extractNamedType(node, state, typeDefinitions);
```
dOrgJelli: Is there a possible way we can avoid the object extractors being dependent on the enum extractors? Can't we check the GraphQL node's type and make sure it isn't an "EnumTypeDefinition"?

packages/schema/parse/src/extract/utils.ts
```diff
@@ -0,0 +1,39 @@
...
+  type: { type: string; required: boolean | undefined },
+  typeDefinitions: TypeDefinitions
+): void {
+  if (isScalarType(type.type)) {
+    property.scalar = createScalarDefinition({
+      name: name,
+      type: type.type,
+      required: type.required,
+    });
+  } else if (isEnumType(type.type, typeDefinitions.enumTypes)) {
```
dOrgJelli: Ideally we should be able to do this by looking at the GraphQL type's "kind" property.

packages/schema/parse/src/typeInfo/enum.ts
```diff
@@ -0,0 +1,5 @@
+export function isEnumType(type: string, enumTypes: string[]): boolean {
```
dOrgJelli: Ideally we can get rid of this, not sure it's needed if we can solve the dependency talked about above.

packages/schema/parse/src/typeInfo/object.ts
```diff
@@ -0,0 +1,5 @@
+export function isObjectType(type: string, objectTypes: string[]): boolean {
```
dOrgJelli: Ideally we can get rid of this

dOrgJelli (changes requested):

packages/schema/parse/src/extract/imported-object-types.ts
mpetrun5: Unfortunately, when extracing object and query arguments GraphQL only thing that I found in node is value that is e.g. `CustomEnum` and the kind that is `NamedType` which doesn't really help me.

That is why I created the type definitions extractor https://github.com/Web3-API/monorepo/blob/568094472eb29ff3f8965155a8cf1b9158f88ded/packages/schema/parse/src/extract/type-definitions.ts#L13 

That should be fast to get all definitions before going into extractors.
https://github.com/Web3-API/monorepo/blob/568094472eb29ff3f8965155a8cf1b9158f88ded/packages/schema/parse/src/index.ts#L48

As now the extractors can still be run in parallel, but they need the first extraction of definitions before going into them.
Looking for `EnumTypeDefinition` when checking if it is object or enum without that first extraction doesn't work because
enum definition can be below the argument where it is used.

This blocks all the comments below.

dOrgJelli: Ah okay, this was the part I didn't understand: "when extracing object and query arguments only thing that I found was the NamedType"

It makes sense that you have to do the first pass to gather all object & enum names, then proceed.

Will re-review now and probably get this merged ASAP.

dOrgJelli: In order to solve this I created a `Blackboard` class that extracts and stores metadata that's required by the extractors. Currently the only function on it is `getCustomTypes` which will return an array of all of the custom types defined in the schema (objects & enums). This array is cached in the blackboard so that other extractors can use the same data without having to re-extract the metadata.

I like this approach better because it creates a clear separation between the TypeInfo and the additional metadata required, making it clear that the TypeInfo is being written to by the extractors, and never read from.

dOrgJelli (approved):

PR #151, opened 2021-02-10
W3 logger

### LINKED ISSUE #33: Web3API Logging Interface
Create a host interface for logging, allowing client modules to:
- log info, warnings, errors
- start / end of a long running processes
- total & current step for sequential step processes (1 ..., 2 ..., 3 ..., etc)

mpetrun5:
@dOrgJelli  General idea

1. Do we plan to support args with message (eg. (message: string, ...args: unknown[]) and let logger implementation handle args or just handle string message?
2. In issue it says  to log long running processes and sequential step processes. Do we plan to handle that on interface level
or let client modules handle message formatting and just log that string?

Changes:
 - add w3 logger function
 - change LogInfoAction to LogAction with log level
 - add optional logger param when instatiating web3 js client

Closes: #33


MakMuftic (dismissed):
LGTM

MakMuftic:
@mpetrun5 conflicts

dOrgJelli (changes requested):
This PR does in fact complete the issue (https://github.com/Web3-API/monorepo/issues/33), but after more careful consideration and discussion with @mpetrun5, we've refactored the issue into this new one: https://github.com/Web3-API/monorepo/issues/185

morrigan:
Closing this one and starting fresh as mentioned above

PR #152, opened 2021-02-15
Enum schema bind

### LINKED ISSUE #72: Schema Binding: Reduce Duplicate Template Code
If you look in the templates directory for the schema-binding project, there are a few shared templates that are used in multiple places. Factor these out into a "common-templates" folder, and have them automatically added to the mustache renderer. This way the template code isn't be reproduced everywhere.

### LINKED ISSUE #107: Support GraphQL Enums
Currently, Web3API schemas do not support Enums. An example use-case we'd like to support is:
```graphql
 enum ResponseType {
    TEXT
    BINARY
}

type Request {
    ...
    responseType: ResponseType!
    ...
}
```

mpetrun5:
Changes:
 - wasm as bind for enums
 - added isNextEnumValue() to msgpack to check if enum value or key is sent
 - removed duplicate templates in schema bind and moved to common root

Closes: #107 
Closes: #72


dOrgJelli:
Hey @mpetrun5, you bring up a good point. I think the solution we should use is as follows:
1. Use constant integer based enums as the default. This is what's most common in programming languages, and is readily available in all.
2. For ease of use, we can support input arguments being either string or integer values for enums. In our generated serialization code, we can check whether the msgpack value is of type string, or number. If it's of type string, we can compare the string with the enum's string values, and manually convert it to an integer.
3. The return value from the query will always come in the form of an integer. If someone (in javascript for example) wanted to convert the number to the enum's corresponding string, this is already supported: https://www.cloudhadoop.com/typescript-enum-string-numbers/

Hopefully this doesn't require too much re-writing, but I think this is the most practical implementation path.

mpetrun5:
> Hey @mpetrun5, you bring up a good point. I think the solution we should use is as follows:
> 
> 1. Use constant integer based enums as the default. This is what's most common in programming languages, and is readily available in all.
> 2. For ease of use, we can support input arguments being either string or integer values for enums. In our generated serialization code, we can check whether the msgpack value is of type string, or number. If it's of type string, we can compare the string with the enum's string values, and manually convert it to an integer.
> 3. The return value from the query will always come in the form of an integer. If someone (in javascript for example) wanted to convert the number to the enum's corresponding string, this is already supported: https://www.cloudhadoop.com/typescript-enum-string-numbers/
> 
> Hopefully this doesn't require too much re-writing, but I think this is the most practical implementation path.

Hey @dOrgJelli 
Thanks for the input.

I will need to definitely write `match{{Enum}}ByValue` and `match{{Enum}}ByKey` for both string and i32 values as enums are typed as i32 and are not indexable (e.g. Enum[i32Value] won't compile in assemblyscript), but my question is do we need both string and i32? 

From UX perspective if we return integers as enum query results than already the client implementation has corresponding enum and if they need to convert it for the return value it shouldn't be a problem to convert  it for query as well. Also, if we support sending strings in query we should support strings for return also, but that complicates implementation a bit because union types are not supported.
As far as strings it made sense to me to match graphQL implementation which sends strings from and to server instead of expecting client to convert it to and from, but I see it complicates the job of writing web3 apis and converting on client is much easier.

dOrgJelli:
@mpetrun5, great points. I think we should:
- Accept string & int as input (client -> WASM) for usability sake.
- Return int (WASM -> client), and have the user convert it to a string if they'd like.
- Add the helper functions you mentioned in the generated code, helping the Web3API developer user enums in either int or string, based on their preference.

packages/wasm/as/assembly/msgpack/ReadDecoder.ts
```diff
@@ -370,6 +370,14 @@ export class ReadDecoder extends Read {
     return false;
   }
 
+  isNextEnumValue(): bool {
+    // Enum integer value is in range 0 - 127
+    if (this.view.peekUint8() <= 127) {
```
mpetrun5: @dOrgJelli 
This is a bit sketchy and don't really know what to do with it.
Msgpack handles string and int that are under some limit without format prefix (e.g. 2 to msgpack is only 1 byte with 2) which then comes as variabe when peeking first Uint8 for format.
https://github.com/msgpack/msgpack-javascript/blob/e218974eac946276def928b8a1cb54f097db0a6e/src/Encoder.ts#L93 

This creates a problem for me as it is hard to know if it is a string or int from format.
Currently, I hardcoded that an enum has max 127 values and if format is less than 127 then it is an int value sent as enum argument, but would to hear your opinion on doing this some way smarter.

dOrgJelli: Why not just use unsigned integers to represent enums? The size of the uint can be dynamically stored for compactness.

packages/schema/bind/src/bindings/wasm-as/index.ts
```diff
@@ -6,23 +6,27 @@ import {
...
 export function generateBinding(schema: string): OutputDirectory {
   const entries: OutputEntry[] = [];
   const typeInfo = parseSchema(schema, {
-    transforms: [extendType(Functions), addFirstLast, toGraphQLType],
+    transforms: [extendType(Functions), addFirstLast, toPrefixedGraphQLType],
   });
 
+  const absolutePath = path.join(__dirname, "./templates");
+  const directory = readDirectory(absolutePath);
+  const subTemplates = loadSubTemplates(directory.entries);
```
dOrgJelli, line 22: Thank you so much for doing this @mpetrun5!!! This removed a lot of duplicate code :sweat_smile:

dOrgJelli (dismissed):
Amazing work @mpetrun5 :clap: bravo! Just made 2 minor refactors. Will merge now.

packages/wasm/as/assembly/msgpack/ReadDecoder.ts
dOrgJelli: See this commit to see how I solved this: https://github.com/Web3-API/monorepo/pull/152/commits/fe1ff1c525e55f2b38f16690cb5afa2e177d90e6

dOrgJelli (approved):

PR #154, opened 2021-02-15
Extend tsconfigs from root

### LINKED ISSUE #116: Add root tsconfig file
TODO
- Add `tsconfig.json` file to root of project
- Where possible extend from this root tsconfig file instead of duplicating the configuration

mpetrun5:
Closes: #116


packages/cli/tsconfig.json
```diff
@@ -1,17 +1,14 @@
...
-    "declarationDir": "build/types",
-    "target": "es5",
-    "module": "commonjs",
-    "moduleResolution": "node",
-    "declaration": true,
-    "sourceMap": true,
+    "lib": [
+      "es2015",
+      "es5",
+      "dom",
```
mpetrunic, line 8: A bit odd that we use dom globals in cli though.
Also we can probably add newer es libs

tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "noImplicitReturns": true,
+    "noUnusedLocals": true,
+    "module": "commonjs",
+    "sourceMap": true,
+    "target": "es5",
+    "resolveJsonModule": true,
+    "strictNullChecks": true
+  },
+  "include": [
+    "./src/**/*"
```
mpetrunic, line 26:
```suggestion
    "./src/**/*.ts"
```

mpetrunic (commented):
Cehck if everything is builded as expected

packages/cli/tsconfig.json
mpetrun5: Use it for WebAssembly and URL
Removed in other places where it is not used

MakMuftic (dismissed):
LGTM

dOrgJelli (approved):

PR #160, opened 2021-02-18
Move tests to reusable package

MakMuftic:
Move test to reusable private package.

### Issues
Closes https://github.com/Web3-API/monorepo/issues/111


packages/test-cases/package.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "@web3api/test-cases",
+    "version": "0.0.1-prealpha.1",
+    "license": "MIT",
+    "main": "index.ts",
```
mpetrunic: private: true

mpetrunic: this package should also depend on assemblyscript so we can test and execute generated asssemblyscript code

dOrgJelli:
Hey @MakMuftic, how is this going? Is it ready for review you think?

dOrgJelli (approved):

PR #163, opened 2021-02-19
Reproducible build pipeline

### LINKED ISSUE #121: Reproducible build pipeline
Split from #95 

**Goals:**
- Easily & consistently build web3api projects from their source files, enabling source verification

**Build Changes:**
- [x] Identify where & how code is currently built
The Web3API project is built and deployed via the Web3API CLI to hardcoded, locally running instances of [IPFS](https://github.com/Web3-API/prototype/blob/034b8c2decb646e677d252da5c273f18934e016d/packages/client-test-env/.env#L1) and [the Graph](https://github.com/Web3-API/prototype/blob/579d2dc860d1eb269da09487de68ef8267ef2144/packages/client-test-env/docker-compose.yml#L26) that were previously setup as the test environment: `npx w3 build \
--ipfs http://localhost:5001 \
--graph simplestorage,http://localhost:8020 \
--test-ens simplestorage.eth`
- [x] Build in consistent, isolated environment: [example of how to do this for a Python environment](https://pythonspeed.com/articles/reproducible-docker-builds-python/)
- [x] Add build configuration to file
- [x] Snapshot build environment (e,g, [`docker commit`](https://docs.docker.com/engine/reference/commandline/commit/)) before compiling to ensure reproducibility
- [x] Verify build is reproducible in a container snapshot

namesty:
Related to #121

For the reproducible build pipeline we are:

- Creating a temp folder under a gitignored .w3 folder
- Creating a `web3api.build.yaml` file where we declare configuration parameters for the build
- Copy all sources and a bash script with build commands into temp folder
- Copy all temp folder contents into Docker environment
- Build sources inside Docker environment by executing the bash script
- Copy build artifacts to host computer
- Delete temp folder

R&D notes:

- Sources are built in the Docker image's build process, there is no container daemonizing and we can avoid needing one entirely.
- Copying directories in Docker is a thing; conditional and iterative logic are not supported in Dockerfile natively, they are supported in bash commands. That's why a temp folder is needed with all sources, and then they can all be copied in a single `COPY` instruction in Dockerfile.
- Copying files from an image to the host computer without using mounts or volumes require creating and running a temp container that does the copying and then gets removed; this seems to be the cleanest/easiest approach I've seen so far, given our use case.
- At this stage we should be aiming to diff built sources only. Since there are factors which are very hard to mitigate (discussed in #121) that make the diffing of 2 image hashes/signatures difficult.
- @dOrgJelli found interesting resources for exploring a docker image, layer contents, and discovering ways to shrink its size (https://github.com/wagoodman/dive). This is related to the standard: https://opencontainers.org/

Misc notes:
- There is a need for a multi-manifest abstraction/tooling now that we have a build manifest.
- The actual WASM build was tested in my fork. I copied the script used but won't work here until packages are published to NPM.
- The custom Dockerfile declared in the build manifest can be commented/removed and the CLI will automatically use the default Node image. I had the project's custom image use a different node alpine version on purpose.

Steps to test:

- Run `yarn` and `yarn build:cli` at root level
- Navigate through command line to `/demos/simple-storage/repro-build-protocol`
- Run `yarn reproducible-build`


demos/simple-storage/repro-build-protocol/deploy-contracts.js
```diff
@@ -0,0 +1,33 @@
+const { EthereumPlugin } = require("@web3api/ethereum-plugin-js");
+const fs = require("fs");
+const YAML = require("js-yaml");
+
+async function main() {
+  const contractAbi = require(`${__dirname}/src/contracts/SimpleStorage.json`);
+
+  const eth = new EthereumPlugin({
+    provider: "http://localhost:8545"
```
benefacto: Should this endpoint/port be defined in a .env file?

demos/simple-storage/repro-build-protocol/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@web3api/examples-simple-storage-repro",
+  "version": "0.0.1-prealpha.1",
+  "license": "MIT",
+  "private": true,
+  "scripts": {
+    "build": "w3 build",
+    "test:env:up": "w3 test-env up && yarn contracts:deploy",
+    "test:env:down": "w3 test-env down",
+    "reproducible-build": "w3 reproducible-build build --quiet=false"
```
benefacto: I feel like the CLI command for this should be reworked to something more intuitive. Maybe `w3 build --reproducible`? Or maybe it should even be the default `w3 build` and a local build should be `w3 build --local`. But probably something should change here.

demos/simple-storage/repro-build-protocol/web3api.build.yaml
```diff
@@ -0,0 +1,9 @@
+image: 
+  name: build-env
+  dockerfile: ./myCustomDockerfile #Comment to use default image
```
benefacto: I think eventually we will want to support various images. I'm not sure if we want to detect which one should be used automatically or have the user specify an image that's not necessarily published by Web3API like `rust:1.40` which we then use a base layer and doing whatever we need on top of it. That all may be outside the scope of this initial pull request though.

demos/simple-storage/repro-build-protocol/web3api.build.yaml
```diff
@@ -0,0 +1,9 @@
...
+env:
```
benefacto: I feel like something other than `env` should be used here as I feel like that's more typically for environment variables. Maybe `build` since you're defining the various files/directories for the build process and that can be expanded on later? Or maybe `source` and then the child properties can be `input`, `output` and eventually all support files/folders/globs?

benefacto (dismissed):

namesty:
@dOrgJelli some still missing TODOs based on our last conversation:

- **Have the copying of source files into the host be a step in the docker file**: this is not something straightforward; it's mostly recommended to create a dummy container that runs the image and then copy from the container to the host (straightforward docker cp) and then remove the dummy container (and this is what I'm doing).

- **Verify that the WASM modules have the correct exports (_w3_invoke, _w3_init)**

- **Make two seperate env variables for build folder & output folder. Don't do anything if there's no output directory specified.**

- **Make sure we support multiple build profile, for example web3api.build.debug.yaml -> docker/debug-DockerFile**: I think this was related to copying a Dockerfile by specifying the actual Dockerfile and not it's containing folder. This should be easy with `fs-extra`, which we are already using. Anyways I just stumbled upon this: https://stackoverflow.com/questions/26077543/how-to-name-dockerfiles

dOrgJelli:
closes https://github.com/Web3-API/monorepo/issues/207
closes https://github.com/Web3-API/monorepo/issues/121

dOrgJelli (dismissed):

dOrgJelli (dismissed):

dOrgJelli (approved):

PR #177, opened 2021-03-01
Added infinite recursion tests

### LINKED ISSUE #149: Detect Infinitely Recursive Object Relationships
Create schema sanitization logic that detects infinitely recursive object relationships, for example:
```graphql
type InfiniteRoot {
  prop: InfiniteProp!
}

type InfiniteProp {
  root: InfiniteRoot!
}
```

Currently if you do this, and include `InifiniteRoot` as an argument on a method (for example), the generated deserialization code is able to detect this, since it's impossible to ever fully satisfy the construction. For example:
```graphql
type Query {
  method4(
    arg: InfiniteRoot!
  ): Result!
}
```
given the schema above, and the query...
```graphql
query {
  method4(
    arg: {
      prop: {
        root: {
          prop: {

          }
        }
      }
    }
  )
}
```
the generated WASM deserialization code will throw the following error:
```
WasmWeb3Api: Thread aborted execution.
Message: __w3_abort: Missing required property: 'root: InfiniteRoot'
```

So without this schema sanitization code, developers are able to create these bad relationships, which can lead to them exposing query functions that can never be invoked properly.

namesty:
Closes #149.

Detecting GraphQL recursive types is a challenge, as far as I know, with current tooling. So, by doing research, I found the following paper, on this exact problem, from a swedish university: http://liu.diva-portal.org/smash/get/diva2:1302887/FULLTEXT01.pdf.

I also found this repo: https://github.com/LiUGraphQL/graphql-schema-cycles-webapp. This repo's based on the previously mentioned paper and contains the javascript implementation we needed. However the code was ugly, it also contained an unnecessary (for our use case) graphicating function, and the dependencies were not handled properly.

So I took their code, organized it as a monorepo, migrated it to typescript (has still lots of `any`s as I did not want to lose too much typing figuring out all types), wrote tests for it, added a simple to use, and typed, interface and we are now using it as an external dependency: https://github.com/namesty/graphql-cycles-detector

I'd appreciate some guidance on how to properly license the package, add the README, and give proper credits to do a proper publish under the web3api domain.

I also think this could be a very useful and interesting contribution to the current GraphQL ecosystem and tooling.


benefacto (commented):
LGTM! One suggestion, though: maybe we should use "cyclic" rather than "infinite" in the code. That'd be more in line with the the error we're displaying to the user and what's actually being tested for. We don't want cycles because they'd be infinite but we're testing for the cyclic references themselves, not infinity, as that might be difficult to test for :) .

dOrgJelli:
Looks great @namesty! It looks like the graphify.ts file is throwing "Field - reference not defined" for the parse's sanity schema: https://github.com/Web3-API/monorepo/blob/prealpha/packages/test-cases/cases/parse/sanity/input.graphql

From what I can tell, it doesn't like the "importedMethod" function on the "TestImport_Mutation" type (I think). Maybe we need to give the package a "ignore" function which allows us to ignore types of certain names. For example, ignore everything by the name of "Query", "Mutation", "..._Query", and "..._Mutation".

namesty:
@dOrgJelli You were right. Also, it doesn't like custom directives with arguments either, since this edge case wasn't contemplated by the original authors.

So I updated the package to accept a parameter to ignore types, and to remove custom dependencies for the circular definition evaluation and bumped the version number. I then updated the usage in the web3api monorepo to search for all type definitions that contain "Mutation", "Query" or "Subscription" and pass them to the validator as types to be ignored.

This worked, and detected circular dependencies in the bind and parse sanity checks, so I updated those and added more test cases to the infinite recursions test section.

Finally, I updated the naming as per @benefacto's recommendation, `infinite recursion` -> `circular definition`

packages/cli/src/__tests__/project/test/test/package.json
```diff
@@ -0,0 +1,5 @@
+
```
dOrgJelli: Can this file be removed? Is it used anywhere?

packages/schema/parse/src/validate/types.ts
```diff
@@ -149,3 +150,33 @@ export function propertyTypes(astNode: DocumentNode): void {
...
+
+export function circularDefinitions(astNode: DocumentNode): void {
+  const operationTypes: string[] = [];
+  const operationTypeNames = ["Mutation", "Subscription", "Query"];
+
+  visit(astNode, {
+    enter: {
+      ObjectTypeDefinition: (node) => {
+        const isOperationType = operationTypeNames.some((name) =>
+          node.name.value.includes(name)
```
dOrgJelli: This doesn't work for types like `type FooQueryOptions { ... }`. We instead want to look for `name === "Query" | "Mutation" | "Subscription"`, and also `name.endsWith("_Query" | "_Mutation" | "_Subscription")`

dOrgJelli: Fixed in my updates

packages/cli/src/__tests__/project/test/test/package.json
dOrgJelli: Removed in my updates

packages/schema/parse/src/__tests__/validate-types.spec.ts
```diff
@@ -118,6 +118,95 @@ type Bar {
...
 
+const circularTypes1 = `
+type A {
+  prop: B!
+}
+
+type B {
+  prop: A!
+}
+`
```
dOrgJelli, line 129: Let's add a test for:
```
type A {
  b: B!
}

type B {
  a: A
}
```
to make sure this doesn't throw an error, as it's a valid relationship.

dOrgJelli (changes requested):

namesty:
I updated the package to contemplate an option to allow for circular references on nullable fields, as it was requested above. I made this on the package side, and a configurable option as that is our Web3API's use case, but probably another project that also uses the package would want otherwise.

Anyways, after the update all tests are passing and I added the requested test case :D

dOrgJelli (approved):

PR #186, opened 2021-03-04
HTTP plugin upgrade

MakMuftic:
### Changes
- Use enums (as originally planned) for response type inside HTTP plugin.


packages/js/plugins/http/src/types.ts
```diff
@@ -11,6 +11,8 @@ export class UrlParam {
   value: string;
 }
 
+// even though in schema enum is used, this will be converted to string
+// so this is a reason to define ResponseType as string here
```
dOrgJelli: Where is this conversion happening? Enums coming from Web3APIs will be integers.

dOrgJelli (changes requested):

packages/js/plugins/http/src/types.ts
MakMuftic: Changed it to `enum`, I didn't refactor tests as I should and that mislead me to the wrong conclusion. Now everything should be good.

dOrgJelli: For the tests when you're sending a query into WASM, you can use a string or a number. But that string gets turned into a number immediately in the deserialization code, and it remains a number from there on out.

dOrgJelli (approved):

PR #188, opened 2021-03-04
Ipfs package

### LINKED ISSUE #135: IPFS WASM Web3API
Web3API package for IPFS that should replace IPFS plugin.

### LINKED ISSUE #247: Enable Web3API CI Using `w3 query ./recipe.json`
Enable the use of the `w3 query ...` command in CI by:
- Exiting the process with exitCode 1 upon an error being hit
- anything else needed?

This will allow us to run CI on the Web3APIs we have build so far (Console, IPFS, ENS, etc) under the ./packages/apis/ directory.

MakMuftic:
##### HTTP plugin
I have refactored the HTTP plugin to support sending body as form data (this is necessary as IPFS expects files to come in this way). Currently, I implemented this by defining `Body` type to be:

```graphql
type Body {
  stringBody: String
  rawBody: Bytes
  formDataBody: FormData!
}
```
This allows users to define what kind of request body (payload) is going to be sent. I am not entirely pleased with this solution...it would be for sure better to define a union type for the body that would include all defined types (`body: String | Bytes | FormData`) but we are not going to support this so I had to find a workaround. Another idea is to have multiple **post** functions defined within HTTP plugin (`postWithStringPayload`, `postWithFormDataPayload`,...).

@dOrgJelli I would like to hear what you think about the current solution.

##### IPFS package

Implemented **catFile**, **catToString** and **addFile** ipfs functions + e2e tests.

##### Other

I expanded **query** command of CLI to support defining paths toward files in format `file:path/to/file/example.txt`. This will read file content and convert it to `Uint8Array`. This is used when testing ipfs function **addFile**, but I think functionality itself could be useful in multiple usecases. 

### TODO
I can't figure out why, but when inside schema definition of HTTP plugin I set property `formDataBody` of `body` to be nullable then generated code has a problem on the compilation. For some reason in generated function `writeHttp_Body` even though there is a check `if (type.formDataBody)` before calling write, the type is being considered to be `Http_FormData | null`, and because of that compilation fails.

Here is the error message:
```
Compile Web3APIERROR TS2322: Type 'src/mutation/w3/imported/Http_FormData/index/Http_FormData | null' is not assignable to type 'src/mutation/w3/imported/Http_FormData/index/Http_FormData'.

     Types.Http_FormData.write(writer, type.formDataBody);
                                       ~~~~~~~~~~~~~~~~~
 in src/mutation/w3/imported/Http_Body/serialization.ts(29,39)
```

And here is a snip from generated code where an error occurs:
```ts
export function writeHttp_Body(writer: Write, type: Http_Body): void {
  writer.writeMapLength(3);
  writer.writeString("stringBody");
  writer.writeNullableString(type.stringBody);
  writer.writeString("rawBody");
  writer.writeNullableBytes(type.rawBody);
  writer.writeString("formDataBody");
  if (type.formDataBody) {
    Types.Http_FormData.write(writer, type.formDataBody);
  } else {
    writer.writeNil();
  }
}
```

### Issues
Closes #188
Closes #247


MakMuftic:
I also changed the CI timeout from 15 to 20 minutes as CI was failing because of this.

dOrgJelli:
Diving into this now, my apologies for lagging on getting to this @MakMuftic, looks fantastic at first look through!

dOrgJelli:
In reply to the lack of union type support in GraphQL & WASM (AssemblyScript), I think we'll hit this problem over and over again, and lacking unions will surely be a problem when porting SDKs to Web3API (especially those written in dynamic languages like JavaScript).

In order to get around this, I think we can simply support unions within any WASM compatible language by create union specific variant objects. For example:
`union Body = String | Bytes | FormData`

Would turn into a simple variant class. In languages that support dynamic memory allocation and as a result have pointers (for ex C/C++), we can be space efficient (psuedo code):
```c++
enum BodyType {
  String,
  Bytes,
  FormData
}

class Body {
  public const BodyType m_type;
  private void* _instance;

  public constructor(BodyType type, MsgPackBuffer* input = 0)
  {
    this.m_type = type;

    switch (this.m_type) {
      case BodyType::String:
        this._instance = input ? deserializeString(input) : new String();
        break;
      case BodyType::Bytes:
        ...
    }
  }

  public getString() String {
    if (this.m_type != BodyType::String) {
      throw new Error(
        `Incorrect Body instance access. Tried to get "String", when instance is of type "${this.m_type}".`
      );
    }
    return String(*this._instance);
  }

  ...
}
```

In languages that don't support dynamic allocations (like AssemblyScript), and also don't have support for union types, we can use a less memory efficient pattern like so:
```typescript
enum BodyType {
  String,
  Bytes,
  FormData
}

class Body {
  public readonly type: BodyType;
  private _string: String | null;
  private _bytes: Bytes | null;
  private _formData: FormData | null;

  constructor(type: BodyType, input?: MsgPackBuffer) {
    this.type = type;

    switch (this.type) {
      case BodyType::String:
        this._string = input ? deseralizeString(input) : new String();
        break;
      case BodyType::Bytes:
        ...
    }
  }

  public getString(): String {
    if (this.type != BodyType::String) {
      throw new Error(
        `Incorrect Body instance access. Tried to get "String", when instance is of type "${this.type}".`
      );
    }
    return this._string;
  }

  ...
}
```

dOrgJelli:
Created this issue: https://github.com/Web3-API/monorepo/issues/246

packages/apis/ipfs/src/mutation/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  const url = input.ipfsUrl + "/api/v0/add";
+  const addResponse = Http_Mutation.post({
+    url: url,
+    request: {
+      headers: [],
+      urlParams: [],
+      responseType: Http_ResponseType.TEXT,
+      body: {
+        formDataBody: {
+          data: [{key: input.fileName, data: String.UTF8.decode(input.data)}]
```
dOrgJelli: It looks like the "fileName" gets lost, as this is a single file and the name gets converted to the data's hash. I don't know if the fileName is needed in this method's arguments, can we just remove?

If you inspect the file on IPFS it looks like this:
image

And when the "name" is parsed from the response, it is the hash, as seen in the recipe output:
image

I think we can remove name from the mutation's arguments & the response object.

packages/apis/ipfs/recipes/constants.json
```diff
@@ -0,0 +1,6 @@
+{
+  "SampleCid": "QmcUZ2oVczdFaRsY8Fcqi8f27GLF8fY6mVboGZGYS8JE72",
+  "IpfsUrl": "http://127.0.0.1:5001",
+  "FileName": "test.txt",
+  "File": "file:./test.txt"
```
dOrgJelli: This is a really nice helper, but I'm unsure if "file:" is the best prefix, as I can see this being used in a case where the user doesn't want file loading to happen. For example, if someone is specifying a URL like so: `file://some/path` and wants this as an argument w/o file loading.

Maybe making another identifier, like `$` for variables, for special operations would be best. Maybe something like:
`#loadFile:./path` where `#` indicates a transform function is to follow, and `loadFile` is a default transformation function that's supported. Users in the future could define their own transformation functions if they so choose.

packages/apis/ipfs/src/mutation/index.ts
```diff
@@ -0,0 +1,47 @@
...
+        },
+        rawBody: null,
+        stringBody: null
+      },
+    }
+  })
+
+  if(addResponse == null || addResponse.status != 200) {
+    throw new Error(`Failed to add file: ${input.fileName}`);
```
dOrgJelli: Would be great to include the status code here as well for the user.

packages/apis/ipfs/src/mutation/schema.graphql
```diff
@@ -0,0 +1,11 @@
+#import { Mutation, ResponseType } into Http from "w3://ens/http.web3api.eth"
+
+type Mutation {
+  addFile(ipfsUrl: String!, fileName: String!, data: Bytes!): AddResult!
```
dOrgJelli: Based on the IPFS HTTP API:
https://docs.ipfs.io/reference/http/api/#api-v0-add
https://github.com/ipfs-shipyard/js-ipfs-http-client-lite/blob/HEAD/API.md#add

I think we can add the following methods to have a more "complete" add implementation:
- Create an `AddFileOptions` type which includes some supported options (like `wrap-with-directory`). If `wrap-with-directory` is specified, the fileName argument is a valid argument and shouldn't be removed (as specified in my prior comment).
- Rename `AddResult` to `AddFileResult`.
- Create an `addFolder` method, which takes an array of folders & files as an argument instead of the fileName + data.

packages/apis/ipfs/src/query/schema.graphql
```diff
@@ -0,0 +1,6 @@
+#import { Query, ResponseType } into Http from "w3://ens/http.web3api.eth"
+
+type Query {
+  catToString(ipfsUrl: String!, cid: String!): String!
```
dOrgJelli: I think we can remove this, as it's a nice helper, but doesn't exist in IPFS's official HTTP API.

packages/apis/ipfs/src/query/schema.graphql
```diff
@@ -0,0 +1,6 @@
...
+  catFile(ipfsUrl: String!, cid: String!): Bytes!
```
dOrgJelli: We can add `CatFileOptions` as another argument here. Some of the arguments I think would be useful (based on this: https://github.com/ipfs-shipyard/js-ipfs-http-client-lite/blob/HEAD/API.md#parameters-7) would be:
- headers
- offset
- length
- qs (query string)

dOrgJelli: Just FYI, in the future when Web3API ENV Variables are supported, the `ipfsUrl` argument will become optional. If it isn't specified in the method's input arguments, the Web3API will try to get it from the ENV. If neither exist, an error occurs.

packages/apis/ipfs/src/query/index.ts
```diff
@@ -0,0 +1,46 @@
...
+    request: {
+      headers: [],
+      urlParams: [{key: "arg", value: input.cid}],
+      responseType: Http_ResponseType.BINARY,
+      body: EmptyBody,
+    }
+  });
+
+  if(catResponse == null || catResponse.status != 200) {
+    throw new Error(`Failed to cat file: ${catResponse?.status} ${catResponse?.statusText}`);
```
dOrgJelli: I think a good standard error message pattern to use might be:
`IPFS catFile Failed.\nStatus: ${status}\nMessage: ${statusText}`

packages/apis/ipfs/src/query/index.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export function catFile(input: Input_catFile): ArrayBuffer {
+  const url = input.ipfsUrl + "/api/v0/cat";
+  const catResponse = Http_Query.get({
+    url: url,
+    request: {
+      headers: [],
+      urlParams: [{key: "arg", value: input.cid}],
+      responseType: Http_ResponseType.BINARY,
```
dOrgJelli: Http_ResponseType can be made into an option within the `CatFileOptions`. This way users can specify if they'd like the response as BINARY or TEXT

packages/apis/ipfs/src/query/index.ts
```diff
@@ -0,0 +1,46 @@
+import { Http_Query, Http_ResponseType, Http_Body } from "./w3/imported";
+import { 
+  Input_catToString,
+  Input_catFile
+} from "./w3";
+import {decode} from "as-base64"
+
+const EmptyBody: Http_Body = {rawBody:null, stringBody:null, formDataBody: {data: []}}
+
+export function catToString(input: Input_catToString): String {
```
dOrgJelli: This can go away if we have `responseType` as an option in the `CatFileOptions`

packages/js/plugins/http/schema.graphql
```diff
@@ -9,30 +9,43 @@ type UrlParam {
 }
 
 type Response {
-  status: Int!
+  status: UInt64!
   statusText: String!
   headers: [Header!]
-  body: String
+  body: String!
```
dOrgJelli: Do you think we should make `JSON` a base type, just like we made `Bytes` a supported base type in GraphQL?

dOrgJelli: Also, is this always present? Can responses have a null body? If so we can make this optional.

packages/js/plugins/http/schema.graphql
```diff
@@ -9,30 +9,43 @@ type UrlParam {
...
+}
+
+type FormDataEntry {
+  key: String
+  data: String
+}
 
 type Query {
-  get(url: String!, request: Request): Response
+  get(url: String!, request: Request!): Response
```
dOrgJelli: Is the `Request` object really required? I looked at axios and it looks like it doesn't have to be present.

packages/js/plugins/http/schema.graphql
```diff
@@ -9,30 +9,43 @@ type UrlParam {
...
 }
 
 type Mutation {
-  post(url: String!, request: Request): Response
+  post(url: String!, request: Request!): Response
```
dOrgJelli: Same as above, is the `Request` object required by axios?

packages/js/plugins/http/src/__tests__/unit/util.test.ts
```diff
@@ -45,8 +46,8 @@ describe("creating axios config", () => {
         { key: "Accept", value: "application-json" },
         { key: "X-Header", value: "test-value" },
       ],
-      responseType: "TEXT",
-      body: "body-content",
+      responseType: ResponseType.TEXT,
+      body: {stringBody: "body-content", formDataBody: {data: []}},
```
dOrgJelli: Can formDataBody just be left undefined here? Why do we need an empty array?

packages/js/plugins/http/src/__tests__/unit/util.test.ts
```diff
@@ -60,8 +61,8 @@ describe("creating axios config", () => {
   test("with url params", () => {
     const config = toAxiosRequestConfig({
       urlParams: [{ key: "tag", value: "data" }],
-      responseType: "BINARY",
-      body: "body-content",
+      responseType: ResponseType.BINARY,
+      body: {stringBody: "body-content", formDataBody: {data: []}},
```
dOrgJelli: Same as above.

packages/js/plugins/http/src/__tests__/unit/index.test.ts
```diff
@@ -151,8 +152,8 @@ describe("test http plugin", () => {
           { key: "X-Test-Header", value: "test-header-value" },
         ],
         urlParams: [{ key: "q", value: "test-param" }],
-        body: "{request: 1001}",
-        responseType: "BINARY",
+        body: {stringBody: "{request: 1001}", formDataBody: {data: []}},
```
dOrgJelli, line 158: Can formDataBody just be left undefined here? Why do we need an empty array?

packages/js/plugins/http/src/__tests__/unit/index.test.ts
```diff
@@ -113,8 +114,8 @@ describe("test http plugin", () => {
           { key: "X-Test-Header", value: "test-header-value" },
         ],
         urlParams: [{ key: "q", value: "test-param" }],
-        body: "{request: 1001}",
-        responseType: "TEXT",
+        body: {stringBody: "{request: 1001}", formDataBody: {data: []}},
```
dOrgJelli, line 120: Same comment as below

packages/apis/ipfs/README.md
```diff
@@ -0,0 +1,45 @@
+# IPFS Web3API package
+
+## Functions
+
+### Query
+
+`catFile(ipfsUrl: String!, cid: String!): Bytes!`
+
+Returns byte content of file, defined by provided _cid_.
```
dOrgJelli: These sorts of comments can be inline in the schema, instead of in the README. Ideally the README just points to the schema so that way it's "self documenting" as changes are made.

packages/js/plugins/http/src/index.ts
```diff
@@ -13,6 +14,9 @@ import {
   PluginPackage,
 } from "@web3api/core-js";
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports
+const formData = require("form-data");
```
dOrgJelli: Can we use a types package for the form-data package? That way we don't have to do `require(...)`?

packages/js/plugins/http/src/index.ts
```diff
@@ -38,11 +42,30 @@ export class HttpPlugin extends Plugin {
   }
 
   public async post(url: string, request: Request): Promise<Response> {
-    const response = await axios.post(
-      url,
-      request.body,
-      toAxiosRequestConfig(request)
-    );
+    const axiosConfig = toAxiosRequestConfig(request);
+    let data: unknown = "";
```
dOrgJelli: Can we encapsulate this logic below into the `toAxiosRequestConfig` function?

dOrgJelli (changes requested):

packages/apis/ipfs/src/mutation/index.ts
dOrgJelli: Based on my other comments below, the fileName option may still be valid :P sorry for the confusion

dOrgJelli:
Also @MakMuftic I found a solution to the bug you found with the AssemblyScript compiler not respecting the null check in the generated code. Pushing a fix now.

dOrgJelli:
^^^ this will allow for formDataBody to be nullable as you've described. This will resolve some of my comments I added to the review.

packages/apis/ipfs/src/query/index.ts
MakMuftic: @dOrgJelli  I agree on the format, so I created a custom `IpfsError` for this, as I think that is the best way to handle this uniformly:

```ts
export class IpfsError extends Error { 
    constructor(method: string, status: number, statusText: string) {
      super(`IPFS ${method} Failed.\nStatus: ${status}\nMessage:${statusText}`);
      this.name = "IPFSError"
    }
}
```

packages/js/plugins/http/schema.graphql
MakMuftic: You are right, it doesn't have to be present.

packages/apis/ipfs/src/query/index.ts
MakMuftic: In the case of _BINARY_ responseType - the return type of `catFile` function should be `ArrayBuffer` and in the case of _TEXT_ responseType - the return type of `catFile` function should be `String`. 

@dOrgJelli  This means that the actual return type should be `ArrayBuffer | String` and that this also relies on union types #246

Maybe we can open a separate issue for this just not to get lost (I left an in-code comment as an additional reminder)?

packages/js/plugins/http/schema.graphql
MakMuftic: > Do you think we should make `JSON` a base type, just like we made `Bytes` a supported base type in GraphQL?

It would be convenient here for sure but I am not sure if there are additional use-cases for this - if we think there will be additional use-cases then we should implement `JSON` base type for sure

cbrzn:
hey @MakMuftic I am trying to build this locally but I am getting an error:

Screenshot 2021-09-08 at 04 30 12

Probably I am missing something, do you have any idea?

MakMuftic:
> hey @MakMuftic I am trying to build this locally but I am getting an error:
> 
> Screenshot 2021-09-08 at 04 30 12
> 
> Probably I am missing something, do you have any idea?

Sorry for the late reply, I will investigate what happened and try to resolve this problem asap

dOrgJelli:
Thank you so much for your work on this @MakMuftic, it's wonderful!

We're migrating this wrapper to the integrations repo in this PR: https://github.com/polywrap/integrations/pull/101

Going to close this.

PR #212, opened 2021-03-09
Logging Web3Api and plugin

### LINKED ISSUE #185: Core Web3API: Logging
Create a "Core Web3API" for Web3API developers to use for logging information, warnings, and errors. Clients must provide a default plugin that implements this core API. App developers can always override it using a uri redirect.

Here're some "half baked" implementation details...

URI: `w3://w3/logger`

Schema:
```graphql
# taken from https://www.tutorialspoint.com/log4j/log4j_logging_levels.htm
enum LogLevel {
  ALL,
  DEBUG,
  INFO,
  WARN,
  ERROR,
  FATAL
}

type Query {
  log(
    level: LogLevel!
    message: String!
  ): Boolean!
}
```

Auto-Generated Helper WASM Code (Assemblyscript):
```typescript
enum LogLevel {
  ALL,
  DEBUG,
  INFO,
  ...
}

class Console {
  static log(message: string): boolean {
    return logger_query_log(LogLevel.INFO, message);
  }

  static error(message: string): boolean {
    return logger_query_log(LogLevel.ERROR, message);
  }
  
  ...
}

function logger_query_log(level: LogLevel, message: string): boolean {
  const argsBuf = serializelogArgs(level, message);
  const result = w3_query(
    "w3://w3/logger",
    "query",
    "log",
    argsBuf
  );
  return deserializelogResult(result);
}
```

Web3API Developer Experience (Assemblyscript):
```typescript
import { Console } from "./w3";

function someFunc() {
  Console.log("some useful info about my API");
}
```

mpetrun5:
Changes:
 - add logger plugin in js that logs to console
 - add default redirect to w3://w3/logger to logger plugin
 - add core web3api that calls logger plugin

Closes: #185


packages/core-apis/console/yarn.lock
```diff
@@ -0,0 +1,2697 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
cbrzn: I guess this file should be removed, right?

dOrgJelli: yes only the root yarn.lock file should exist in the monorepo

mpetrun5: Oups :sweat_smile:

dOrgJelli (dismissed):

dOrgJelli (approved):

PR #214, opened 2021-03-10
Graphql aliases

### LINKED ISSUE #191: Use GraphQL Aliases
Currently, in order to support multiple parallel queries, we post-fix a number to the query result's key. For example, the query:
```graphql
{
  someQuery(arg: 5)
  someQuery(arg: 10)
}
```
Would have its results returned as:
```typescript
{
  someQuery_0: "result 5",
  someQuery_1: "result 10"
}
```

This is not consistent with the GraphQL standard, where the recommended approach is to use aliases for each query, as described here: https://graphql.org/learn/queries/#aliases

This would change the above to:
```graphql
{
  a: someQuery(arg: 5)
  b: someQuery(arg: 10)
}
```
```typescript
{
  a: "result 5",
  b: "result 10"
}
```

mpetrun5:
Changes:
  - removed logic for making query results unique
  - added alias field to InvokeApiOptions
  - extracted alias when parsing query and validated that aliases and methods are unique

Closes: #191


packages/js/core/src/types/Invoke.ts
```diff
@@ -13,6 +13,9 @@ export interface InvokeApiOptions {
   /** Method to be executed. */
   method: string;
 
+  /** Method alias in executed result */
+  alias?: string;
```
dOrgJelli: This shouldn't be necessary, since this is a query specific construct. The InvokeApiResult is the value itself, and is not nested under the "${alias}" key

packages/js/core/src/algorithms/parse-query.ts
```diff
@@ -46,6 +46,14 @@ export function parseQuery(
         );
       }
 
+      let alias: string | undefined = undefined;
+      if (selection.alias) {
+        alias = selection.alias.value;
+        if (invokeOptions.find((io) => io.alias === alias)) {
```
dOrgJelli: Instead of doing this, we can change the return type of parseQuery so that it's a mapping from alias name to InvokeOptions. Here's the type:
```
interface QueryInvocations {
  [alias: string]: InvokeApiOptions;
}
```

dOrgJelli:
Making the above changes now, about to push

dOrgJelli (dismissed):

dOrgJelli (approved):

PR #218, opened 2021-03-11
Pretty uri redirects

### LINKED ISSUE #189: Pretty URI Redirects & Plugins
In order to make this look less gross:
```typescript
const redirects: UriRedirect[] = [
  {
    from: new Uri("w3://ens/ethereum.web3api.eth"),
    to: {
      factory: () => new EthereumPlugin({ provider: ethereum }),
      manifest: EthereumPlugin.manifest(),
     },
  }
]

const client = new Web3ApiClient({ redirects });
```

We can...

1. Use strings (`"..."`) instead of the `new Uri("...")` class. We'll just construct the `Uri` class behind the scenes. This is easier for users as they have to think about less.
2. Give plugins a face lift so they look less verbose when adding them to the "to: ..." property. Here's some rough psuedo code from @mpetrunic on how we can do this without loosing type safety:
```typescript
type Web3ApiPluginModule<Opts> = (opts: Opts) => {
   manifest: Manifest,
   factory: (opts: Opts) => Web3ApiPlugin<Opts>;
};

interface UrlRedirect<Opts = Record<string, unknown>> {
   from: string,
   to: string | Web3ApiPluginModule<Opts>
}
```

Adopting the above improvements would simplify it to:
```typescript
import { plugin as ethereum } from "@web3api/ethereum-plugin-js";

const redirects: UriRedirect[] = [
  {
    from: "w3://ens/ethereum.web3api.eth",
    to: ethereum({ provider: window.web3 })
  }
]

const client = new Web3ApiClient({ redirects });
```

mpetrun5:
Changes:
 - changed redirect uri from Uri to string
 - exported defualt plugin factory from js plugins
 - changed uri redirect types to accept plugin factory
 - update plugin template

Closes: #189


MakMuftic (changes requested):
@mpetrun5 I think these changes should also be made for the plugin template `packages/templates/plugin`

mpetrun5:
> @mpetrun5 I think these changes should also be made for the plugin template `packages/templates/plugin`

Should already be here:
https://github.com/Web3-API/monorepo/blob/e27addd671fef58ff95156f7f7b9a9f3d465ae75/packages/templates/plugin/typescript/src/index.ts#L44

MakMuftic (dismissed):
LGTM - my eyes tricked me before

packages/templates/plugin/typescript/src/index.ts
```diff
@@ -35,3 +40,12 @@ export class SamplePlugin extends Plugin {
     return this._config.defaultValue;
   }
 }
+
+export const samplePlugin: PluginFactory<SamplePluginConfig> = (
```
dOrgJelli, line 44: Minot nit, but I think each plugin module should export its PluginFactory by the name `plugin`. In the future this will help us to dynamically import plugin module, and know that there should be an export named `plugin` that represents its factory.

We could do this as a default module export as well, but I prefer having the named export. Just personal opinion.

dOrgJelli: When you import multiple, you'll just have to rename the import like so `import { plugin as ethereumPlugin } from "..."`

dOrgJelli (changes requested):

packages/js/core/src/types/UriRedirect.ts
```diff
@@ -1,14 +1,14 @@
-import { Uri, PluginPackage } from ".";
+import { PluginPackage } from ".";
 
 /**
  * Redirect from one URI, or a set of URIs, to a new URI or a plugin.
  */
 export interface UriRedirect {
   /** Redirect from this URI */
-  from: Uri;
+  from: string;
```
dOrgJelli: I think that keeping the underlying `UriRedirect` type the same is useful. Here's why:
- Switching to strings is purely to help keep the user/developer facing interface simple and easy to use.
- Having the backing type remain a `Uri` is useful so we can be sure that we're operating on a "correct" URI string that's been sanitized.

In the current implementation you have, a user wouldn't know if their provided string is an incorrect URI until it gets converted into a `new Uri(userString)` somewhere in the algorithms at run-time. This is bad sanitization.

We should instead be converting the strings to `new Uri`s within the Web3ApiClient constructor, so that we can catch bad URI strings right off the bat.

dOrgJelli (changes requested):

packages/templates/plugin/typescript/src/index.ts
mpetrunic: Renaming every plugin import is quite bad dev experience. Maybe export both named as samplePlugin and default export for dynamic importing?

packages/js/core/src/types/UriRedirect.ts
mpetrun5: Created new type `UriRedirectDefinition` that has from and to as string which is what users will need to pass into client and then converted to UriRedirect in constructor.
Is this more in line with what you had in mind?

dOrgJelli: Yes that sounds great

dOrgJelli (approved):

PR #226, opened 2021-03-18
Query-Time Re-Configuration

### LINKED ISSUE #155: Query Time Redirects
Currently, we define URI Redirects upon instantiating the Web3API Client. After chatting with a launch partner it has become clear that it'd also be beneficial to have query-specific-redirects, like so:
```typescript
client.query({
  uri: "some-protocol",
  query: "mutation { someFunction(...) }",
  redirects: [{
    from: "ethereum.web3api.eth",
    to: "ethereum-proxy.protocol.eth",
    // maybe also configuration values here too for ethereum-proxy
  }]
})
```

This is very useful for 1-time redirects that aren't needed for the entire life-time of the application.

mpetrun5:
@dOrgJelli Have some issues with this feature.

It is quite hard to get query redirects to propagate to all invoke and query calls. Currently we have client.redirects() that is called in `resolveUri` and in further calls client instance is propagated which has the redirects. It is quite hard to do this with query time redirects as I can't seem to find all the places query redirects should be propagated and get `No web3api found` error and I can't bind redirects to instance as then it would not be safe with other query calls. 

Also one issue is api cache. As I understand when query time redirects are used api cache can't be used because the address on which that api is used can always be redirected to something other than required and api cache should not be used with query time redirects?
 
Linked issue: #155


dOrgJelli:
This is indeed a tricky situation to handle @mpetrun5, I'm not sure what the best approach would be here... I'd like to brainstorm with you a bit about it if you have the time! Please feel free to schedule something here: https://calendly.com/dorgjelli

packages/js/client/src/Web3ApiClient.ts
```diff
@@ -46,10 +53,18 @@ export class Web3ApiClient implements Client {
 
     // Add all default redirects (IPFS, ETH, ENS)
     this._config.redirects.push(...getDefaultRedirects());
+    this._invokeContextMap.set(DEFAULT_CONTEXT_ID, {
```
mpetrun5: This is a bit of a hack because we use loadWeb3Api directly inside SchemaCompose and not query id is specified to fetch context.

https://github.com/Web3-API/monorepo/blob/5c1fbddf253783709c923a6247e74b1769e694fb/packages/cli/src/lib/SchemaComposer.ts#L113

mpetrun5 (commented):
@dOrgJelli  Discussed draft changes

 - Moved wrap client to invoke() instead of PluginWeb3Api. IMO it makes more sense because it covers WasmWeb3Api as well now and classes that extend Api don't have to worry about dealing with query ids
 - enabled sub invokes to set custom invoke context and merged that with parent context to enable developers to set redirects in code. Don't know if that is what you had in mind?

I forgot on the call we had to ask about cache. If someone redirects from some uri that is already cached won't he get the cached result instead of the redirected one? We probably need some way to check if to cache result is the valid we are looking for.

packages/js/client/src/Web3ApiClient.ts
cbrzn: This has been removed since we are not using it anymore in the CLI

packages/js/core/src/algorithms/resolve-uri.ts
```diff
@@ -42,13 +52,16 @@ export const resolveUri = Tracer.traceFunc(
     const uriResolverImplementations = getImplementations(
       coreInterfaceUris.uriResolver,
       interfaces,
-      redirects
+      finalRedirects
     );
 
+    let clientInstance = client;
+    if (id) clientInstance = wrapClient(client, id);
```
Niraj-Kamdar:
```suggest
    let clientInstance =id ? wrapClient(client, id) : client;
```

Niraj-Kamdar (dismissed):
LGTM! Just a small change!

dOrgJelli (approved):

PR #228, opened 2021-03-19
Network Specific ENS Lookup

dOrgJelli:
# Notable Changes
### Constructor
The EthereumPlugin can now be constructed with multiple network connections like so:
```typescript
import { ethereumPlugin } from "@web3api/ethereum-plugin-js";
import { Wallet } from "ethers";

const client = new Web3ApiClient({
  redirects: [
    {
      from: "ens/ethereum.web3api.eth",
      to: ethereumPlugin({
        mainnet: {
          provider: window.ethereum
        },
        rinkeby: {
          signer: new Wallet("mnemonic")
        },
        testnet: {
          provider: "http://localhost:8080
        },
        // If not specified, mainnet is used
        defaultNetwork: "rinkeby"
      })
    }
  ]
})
```

If no provider is provided, the name of the network (ex: "rinkeby" above) is used to get a default provider from ethers.js.

### Connection Class
Multiple network connections is made possible by abstracting away the concept of a network connection into the `Connection` class.

### Schema
The EthereumPlugin's schema has been updated to allow for custom network configurations to be specified. These configuration options are used to select which network connection you'd like to transact with.

```graphql
type Query {
  callView(
    address: String!
    method: String!
    args: [String!]
    connection: Connection
  ): String!
}

type Mutation {
  sendTransaction(
    address: String!
    method: String!
    args: [String!]
    connection: Connection
  ): String!

  deployContract(
    abi: String!
    bytecode: String!
    args: [String!]
    connection: Connection
  ): String!
}

type Connection {
  node: String
  networkNameOrChainId: String
}
```

## ENS Plugin
The EnsPlugin has been updated to allow for network names in the URI's path like so:
```
w3://ens/testnet/simplestorage.eth
w3://ens/rinkeby/other.web3api.eth
```

If no network is specified, `mainnet` is used.

## SimpleStorage Web3API Demo
The SimpleStorage Web3API demo has been updated in the following ways:
1. The `e2e.json` query recipe now specifies `/ens/testnet/simplestorage.eth` for the API it is connecting to.
2. In the SimpleStorage.eth's schema, it now supports configuring what Ethereum connection to be used. For example:
```graphql
#import { Query, Connection } into Ethereum from "w3://ens/ethereum.web3api.eth"

type Query {
  getData(
    address: String!
    connection: Ethereum_Connection
  ): Int!
}
```
If nothing is specified, the EthereumPlugin's `defaultProvider` is used.

NOTE: in the future nested configuration values like this will be configured through Web3API env variables (see: https://github.com/Web3-API/monorepo/issues/140)
3. The `e2e.json` queries set the query method's `connection` to `testnet`. Otherwise, the Web3API would try and send transactions to `mainnet`.


packages/js/plugins/ethereum/src/Connection.ts
```diff
@@ -0,0 +1,148 @@
...
+    return connections;
+  }
+
+  static fromNetwork(networkish: Networkish): Connection {
+    return new Connection({
+      provider: ethers.providers.getDefaultProvider(
+        ethers.providers.getNetwork(networkish)
+      ) as JsonRpcProvider,
+    });
+  }
```
namesty, line 58: Isn't it possible that the networkish value is a websocket URI? Because if that's the case, `getDefaultProvider` will return a `WebsocketProvider` instead of a `JsonRpcProvider` and we may have unexpected bugs, no?

namesty (commented):
I noted we cannot have more than one connection to the same network. I don't know if this is something we should allow anyways

packages/js/plugins/ethereum/src/Connection.ts
dOrgJelli: `WebsocketProvider` derives from `JsonRpcProvider`, so this shouldn't be an issue: https://github.com/ethers-io/ethers.js/blob/3b1d3fcee6bfb5178861e26ff1a1e9daa0663ec9/packages/providers/src.ts/websocket-provider.ts#L46

dOrgJelli: Also note that `Networkish` isn't supposed to be a URI string, but rather a network name as a string ("mainnet", "rinkeby", etc).

dOrgJelli:
> I noted we cannot have more than one connection to the same network. I don't know if this is something we should allow anyways

Hmmm interesting... I don't think this is a use-case we need to worry about since current wallets don't support this anyway. The most I've seen done in relation to this is having multiple fallback endpoints for a single network (ethersjs does this). I've never seen a library support multiple providers for a single network. How would we differentiate between them?

Maybe an interface like this may make sense?
```
connections: {
  mainnetA: {
    network: "mainnet",
    provider: ...
  },
  mainnetB: {
    network: "mainnet",
    provider: ...
  }
}
```

PR #263, opened 2021-04-06
Spanish translation of CLI messages

rihp:
All lines translated to spanish. 
Could do a final re-read tomorrow to make sure there are no mistakes slipping through.


dOrgJelli:
Closing this for now as it has become pretty out of date, and we plan to add internationalized text in the future through a better process.

PR #279, opened 2021-04-13
Trade refactor and tests

mpetrun5:
@krisbitney @dOrgJelli Here is what I had in mind to refactor current implementation to match uniswap more closely.
Removed separate `ProcessedTrade` and I expect the trade to be created with all fields and created separate helper method to create `Trade` as you would in js sdk.

Still has some bugs to iron out and we don't support WETH yet, but we are getting close to finishing this portion of queries 
and then with this I can move on to mutations.

Also @krisbitney, is it okay if I finish the trade methods?


dOrgJelli (approved):
LGTM!

mpetrun5:
Think this is ready for review. It should work in most cases now except that WETH is not implemented, but I would leave that to a separate PR and computeMidPrice implementation.

krisbitney:
> @krisbitney @dOrgJelli Here is what I had in mind to refactor current implementation to match uniswap more closely.
> Removed separate `ProcessedTrade` and I expect the trade to be created with all fields and created separate helper method to create `Trade` as you would in js sdk.
> 
> Still has some bugs to iron out and we don't support WETH yet, but we are getting close to finishing this portion of queries
> and then with this I can move on to mutations.
> 
> Also @krisbitney, is it okay if I finish the trade methods?

This is really great--huge for the UX. You can finish the trade methods if you'd like, sure.

krisbitney (approved):

PR #294, opened 2021-04-14
Handle IPFS Stalling W/ Timeouts & Fallback Providers

### LINKED ISSUE #290: IPFS `cat` Timeout
Currently, if a file is not accessible via the IPFS `cat` query, the query stalls and hangs forever.

We should add the ability to specify a timeout for `cat` queries. Also along with the timeout, it may be good to specify a fallback IPFS endpoint to use.

This situation is common when talking to a local IPFS node, and requesting an outside document the local node does not know about yet. With the above features implemented, the timeout would be triggered, and the fallback endpoint (public IPFS gateway) would be use.

dOrgJelli:
When an IPFS node cannot access a requested document, it will hang. In order to work around this, I've added options for:
- Registering fallback providers to try if an operation fails to complete on the default provider.
- Specifying a timeout to be used for a given operation.
- Specifying a provider override to be used for a given operation.
- Added support for the `resolve` API, allowing you to quickly check if a file exists.

closes https://github.com/Web3-API/monorepo/issues/290

### DIFF
--- a/packages/js/plugins/ipfs/package.json
+++ b/packages/js/plugins/ipfs/package.json
@@ -22,11 +22,12 @@
   "dependencies": {
     "@web3api/core-js": "0.0.1-prealpha.18",
     "cids": "^1.1.4",
-    "ipfs-http-client-lite": "0.3.0",
+    "@dorgjelli-test/ipfs-http-client-lite": "0.3.1",
     "is-ipfs": "1.0.3"
   },
   "devDependencies": {
     "@types/jest": "26.0.8",
+    "abort-controller": "3.0.0",
     "jest": "26.6.3",
     "rimraf": "3.0.2",
     "ts-jest": "26.5.4",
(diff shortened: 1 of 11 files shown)


packages/js/plugins/ipfs/package.json
```diff
@@ -22,11 +22,12 @@
   "dependencies": {
     "@web3api/core-js": "0.0.1-prealpha.18",
     "cids": "^1.1.4",
-    "ipfs-http-client-lite": "0.3.0",
+    "@dorgjelli-test/ipfs-http-client-lite": "0.3.1",
```
dOrgJelli: This forked package has the following modification to it: https://github.com/ipfs-shipyard/js-ipfs-http-client-lite/pull/5

MakMuftic (dismissed):
LGTM :+1: 

I will include these changes into IPFS package #188

MakMuftic:
@dOrgJelli just some linter errors to fix

PR #310, opened 2021-04-23
`w3 query --redirects ./redirects.js ...` Support

### LINKED ISSUE #244: Register Custom Redirects When Using CLI
Support `--redirects` flag where you provide a `js` file that returns custom redirects to be added to the client.

This will enable custom plugins to be used through the CLI.

Use this for all commands that instantiate a Web3API Client (`w3 query`, other?)

dOrgJelli:
closes: https://github.com/Web3-API/monorepo/issues/244

TODO:
- [ ] See if other commands could use this same feature
- [ ] Create tests


dOrgJelli:
An example `redirects.js` file may look something like:
```javascript
var customPlugin = require(__dirname + "/node_modules/custom-plugin-js");

module.exports.redirects = [
  {
    from: "ens/custom.eth",
    to: customPlugin.plugin()
  },
  {
    from: "ens/api.eth",
    to: "ens/myapi.eth"
  }
];
```

PR #319, opened 2021-04-29
Uniswap API query updates

krisbitney:
Updates include:
* query unit tests -- all passing
* basic mainnet fork setup and working (ganache)
* addition of BigFloat (handles decimal numbers and formatted string output)
* newest version of BigInt (huge performance improvements and some bug fixes)
* updates to string output for decimal strings
* updates to fetch functions to use network-specific ethereum plugin calls
* includes e2e tests for fetch, pair, route, and trade core functions -- all passing
* various fixes to bugs found while testing


packages/apis/uniswapv2/src/__tests__/e2e/fetch_e2e.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+          )
+        }
+      `,
+      variables: {
+        chainId: tokens[0].chainId,
+        address: tokens[0].address,
+      },
+    });
+
+    expect(tokenData.errors).toBeFalsy();
```
mpetrun5: Is there any way to assert the value in this tests?
This is not really that usefull. We should probably call 
the ether separately for the result and compare.

packages/apis/uniswapv2/src/query/trade.ts
```diff
@@ -22,18 +22,26 @@ import {
...
+import { wrapIfEther } from "../utils/utils";
 
 import { BigInt } from "as-bigint";
+import { BigFloat } from "as-bigfloat";
+import { Nullable } from "@web3api/wasm-as";
 
 export function createTrade(input: Input_createTrade): Trade {
   const amounts: TokenAmount[] = new Array(input.route.path.length);
-  const nextPairs: Pair[] = new Array(input.route.pairs.length);
+  const nextPairs: Pair[] = new Array(input.route.pairs.length); // TODO this array is created and updated, but never read from. Do we need it?
```
mpetrun5: Can you delete if it is not used?

packages/apis/uniswapv2/src/query/trade.ts
```diff
@@ -392,17 +403,9 @@ function _bestTradeExactOut(
...
-  const midPrice = routeMidPrice({
-    route: trade.route,
-  });
-  const midPriceBI = BigInt.fromString(midPrice.amount);
-  const inputAmountBI = BigInt.fromString(trade.inputAmount.amount);
-  const outputAmountBI = BigInt.fromString(trade.outputAmount.amount);
-
-  const exactQuote = midPriceBI.mul(inputAmountBI);
-  return exactQuote.sub(outputAmountBI).div(exactQuote);
+function computePriceImpact(trade: Trade): BigFloat {
```
mpetrun5, line 388: Can you please implement this now as a exported query function?
I think I left this as TODO as we didn't have floats, but should be implemented and exported now I think

krisbitney: This is the same as the tradeSlippage function, which is exported

packages/apis/uniswapv2/src/query/trade.ts
krisbitney: I deleted it. I didn't delete it before because I wasn't sure if there was a reason it was added.

packages/apis/uniswapv2/src/__tests__/e2e/fetch_e2e.spec.ts
krisbitney: Still working on this. Will update you when it's done.

krisbitney: done

krisbitney:
I updated the fetch tests. The fetchPairData and fetchTotalSupply tests can sometimes fail because a block is mined at the wrong moment, leading the api query and uniswap sdk query to provide slightly different numbers. If that occurs, just re-run the tests.

mpetrun5 (approved):

PR #328, opened 2021-05-10
Infra Pipeline: Modular Environment Setups

namesty:
## Overview

This PR aims to enable the new `infra` CLI command. This command would be used to provide a better, modular and extensible environment orchestration leveraging `docker-compose`. It also introduces the new `infra manifest` extension.

The user would declare `modules`, which are packages that expose a `docker-compose` YAML file. The CLI would fetch these packages, cache them, resolve the `docker-compose` YAML of each, and bundle them into a single `docker-compose project`.

Additionally, the user can write a custom `docker-compose` file and declare additional services or custom functionality in case no modules suffice their needs. This custom `docker-compose` is declared and gets bundled and orchestrated along with the rest of the modules.

The CLI also provides some "preset" (or default) environments; which are already properly configured for typical/common use cases. If the user wishes to use one of these, his project's `infra manifest` gets ignored, and the preset's environment's gets used instead.

The `infra` command has the following subcommands:

- `up`: takes all modules and performs a `docker-compose up`
- `down`: same as `docker-compose down`
- `config`: takes all modules, performs the bundling and outputs the bundled `docker-compose` project. Useful for validation and debugging.
- `vars`: prints a list of all required environment variables needed across all bundled modules

And the following flags:

- `--modules`: acts as a filter of modules. Instead of bundling all modules declared in the infra manifest, it only bundles those specified in the flag.
- `--preset`: ignores project's infra manifest and uses the selected preset's configuration.

## Additional changes

- Given the modular, extensible capability of the `infra` pipeline, a preset was created to replace the current "test environment". It orchestrates ganache + ens + ipfs, and provides deterministically consistent and permanent ENS contract addresses. The ENS redeployment endpoint was removed from the DevServer. This package is maintained [here](https://github.com/namesty/infra-EthEnsIpfs).
- Leveraging the previous point, a new deployer package was created to assume and use the previous preset environment. No need to specify the ENS registry address in the deploy manifest when deploying on local testnet anymore.


namesty:
- Abstracted env logic in its own class, similar to what the reproducible build pipeline does with `Project` class
- Added new manifest format for environments
- Added default manifest for users who don't explicitly declare an env manifest but want a basic ganache + ipfs + local server with ENS (we use this to test other parts of the monorepo, like the Ethereum plugin)
- Improved messages and prompts
- Improved error handling and argument validation
- Added e2e tests

namesty:
## Important implementation details

- Modules can be fetched from a local path or from a remote registry (like npm). However, to make it platform agnostic, I created a `InfraDependencyFetcher` abstract class that aims to encapsulate all registry-specific logic. That way we could fetch from npm, pip or cargo behind a common, very simple, interface. The implementation for npm is already done and being used in this PR.

- In case a module's `docker-compose` file declares a build context, the CLI modifies this path for an absolute one. This is because, when using `docker-compose`'s bundling functionality with `-f`, all paths used are relative to the first one used in the command. More info on this behavior: https://docs.docker.com/compose/extends/#multiple-compose-files

- In case two different modules declare equally named services, we are letting `docker-compose` handle it. The default behavior is as follows: if the services are using the same image, but with a different configuration (e.g.: two equally named ganache services with different ports) it merges the configurations; but if the the images are different, each equally named service silently overwrites the previous one. More info on this behavior: https://github.com/docker/compose/issues/5267

cbrzn:
blocked by https://github.com/polywrap/monorepo/pull/866

cbrzn (dismissed):

cbrzn:
Note that once this is merged & released, the integrations repository needs to be updated - Probably creating an issue would make sense? :thinking:

dOrgJelli (dismissed):

dOrgJelli (approved):

Issue #301, opened 2021-04-17
Make BigInt A Supported Schema Base Type


dOrgJelli:
@krisbitney has created an [Assemblyscript "BigInt" library](https://github.com/Web3-API/as-bigint), which enables math with arbitrarily large integers.

Currently, these integers are passed to the API (using the schema's types) as integers. I think it'd make sense to add support for `BigInt` as a supported schema base type. The underlying serialization format could just be a string (or number) that gets used to construct a BigInt class object.

PR #331, opened 2021-05-12
changed return types of price amounts from TokenAmount to string

krisbitney:
As I'm implementing the web3api in the uniswap interface, I'm realizing it's a bad idea to mix the expected form amount string in the TokenAmount type, which generally expects an integer but in some cases was given a decimal string by query functions such as tradeExecutionPrice(...)

This PR also includes two other bug fixes:

I added a bug fix to this PR that changes the output of the tradeSlippage function (aka price impact). It previously returned values that were multiplied by 100 to be formatted as percents. I changed that because I realized the JS sdk doesn't do that. Now the results will match the SDK output.

I also fixed bestTrade functions so they will no longer wrap Ether.


dOrgJelli:
Should we add BigInt as a base type @krisbitney?

krisbitney:
> Should we add BigInt as a base type @krisbitney?

I think it would be useful for clarifying and enforcing type expectations, and to keep people from making the kinds of mistakes I made here (which this PR is intended to fix). It would be a good long-term addition to web3api. I can work on it once we get the uniswap javascript sdk replaced with the uniswap web3api in the uni interface, or anyone else can work on it. I know there is an issue for it: Issue #301.

krisbitney:
I added a bug fix to this PR that changes the output of the tradeSlippage function (aka price impact). It previously returned values that were multiplied by 100 to be formatted as percents. I changed that because I realized the JS sdk doesn't do that. Now the results will match the SDK output.

PR #332, opened 2021-05-13
Swap mutation functions

mpetrun5:
- added swap, exec and approve (swap only) mutation functions
 - addded deploy function to package.json
 - added e2e tests
 - changes ethereum plugin to accept value, gasLimit and gasPrice and to now wait for tx as it could take a long time
 - hardcoded uniswap token list in tests to avoid occasional ipfs errors when fetching token list
 
 I created args JSON array manually because I couldn't find logic in https://github.com/nearprotocol/assemblyscript-json to create JSON array without key.

@krisbitney Do you maybe know the logic to unwrap WETH when creating trade. Swap can't send or receive ETH now as bestTrade functions wrap ETH in WETH for calculations, but I can't seem to find logic to unwrap that in interface. In exec tests I manually changed it back to ETH for testing purposes, but swap needs to have that logic.
Also, currently I just use first trade that bestTrade returns, but that should be revised also which I need to look further in uniswap interface logic.


krisbitney:
The uniswap interface logic finds the best trade and then compares it to the best trade if uniswap v1 were to be used. The trade that is executed is the better of the two. Since we aren't worrying about v1, the trade returned by bestTrade is the right one.

I'm not yet sure where the actual transaction is called for unwrapping when Eth is the intended output, but this logic is used for unwrapping when a user explicitly requests to unwrap:
original: https://github.com/Web3-API/uniswap-interface/blob/main/src/hooks/useWrapCallback.ts
modified for web3ai: https://github.com/Web3-API/uniswap-interface/blob/replace-sdk-with-web3api/src/hooks/useWrapCallback.ts

krisbitney (approved):

krisbitney:
There is also this, which might be helpful: https://github.com/Web3-API/uniswap-interface/blob/replace-sdk-with-web3api/src/utils/w3WrappedCurrency.ts

krisbitney:
You can disregard my prior comments about wrapping ether. I fixed a bug in the bestTrade functions. They will no longer wrap Ether.

PR #354, opened 2021-05-29
MsgPack Context -> pseudo-tracing for MsgPack serialization

krisbitney:
The MsgPack Context class adds more information to errors thrown during serialization and deserialization of wasm objects. 

Example:
image


packages/test-cases/cases/bind/sanity/output/wasm-as/mutation/Mutation/serialization.ts
```diff
@@ -36,14 +38,22 @@ export function deserializemutationMethodArgs(argsBuf: ArrayBuffer): Input_mutat
     numFields--;
     const field = reader.readString();
 
+    reader.context().push(field, "unknown", "searching for property type");
     if (field == "str") {
+      reader.context().push(field, "string", "type found, reading property");
       _str = reader.readString();
       _strSet = true;
+      reader.context().pop();
+      reader.context().pop();
```
dOrgJelli: This second "pop" for the "searching for property type" message above can be moved below the if / else if chain. This will make sure that the "searching..." message doesn't get left on the stack if the name of the property is unknown.

Making a push now with this update.

packages/wasm/as/assembly/msgpack/Read.ts
```diff
@@ -48,4 +49,6 @@ export abstract class Read {
 
   abstract isNextNil(): bool;
   abstract isNextString(): bool;
+
```
dOrgJelli, line 52: I'd like to add context messages when reading items from an array. Will make a push now with this added.

dOrgJelli (approved):

PR #364, opened 2021-06-07
Support client.getManifest(...) and client.getFile(...)

### LINKED ISSUE #261: Support `client.getManifest(...)` and `client.getFile(...)`
Create a method on the Web3API Client for fetching known manifest file types.

Some implementation details:
```typescript
type ManifestFile = "web3api.yaml" | "web3api.meta.yaml" | "other..."
type ManifestType = Manifest | ... // will extend in the future when issue #207 is complete
interface GetManifestOptions {
  uri: string
  manifest: ManifestFile
}

class Web3ApiClient {
  ...
  public async getManifest(options: GetManifestOptions): Promise<Manifest> {
    const uri = new Uri(options.uri);
    const api = await this.getApi(uri);
    // TODO: need to parse return result using manifest parser, which has validation built in
    return await api.getFile(options.manifest);
  }
}
```

related: #207

krisbitney:
Closes #261.


cbrzn (dismissed):
Tested on the hub and worked perfectly! Just need to fix merge conflicts and detail in the build manifest schema which I already talked about with Kris 

Amazing job :)

dOrgJelli:
Awesome work @krisbitney @cbrzn! Will get this merged ASAP!

dOrgJelli:
Almost finished fixing this up. Will post a comment once complete explaining all the changes made @krisbitney @cbrzn.

dOrgJelli (dismissed):

dOrgJelli (approved):

dOrgJelli:
Changes Made:
* Removed the default meta manifest, as it didn't make much sense for something that should be optional.
* Copy all files referenced in the `web3api.meta.yaml` file into the build folder.
* Added `web3api.meta.yaml` test files to the CLI test project.
* Added CLI build command test case, ensuring the meta manifest & associated files are copied into the build directory.
* Fixed the meta manifest's schema, and added missing fields
* Removed the plugin manifest from "getManifest(...)". Instead "getSchema(...)" can be used.
* Improved core typings for Client, Plugin, Query, and Invoke.
* Created a new version of the web3api manifest format (0.0.1-prealpha.5), so that we don't modify existing formats.
* Found a bug, where schemas that include interfaces can have duplicate methods. Added a validator for this, so it causes an error.
* Removed the "client.loadWeb3Api(...)" method, as it returns an Api class instance. This is not desired when garbage collection is merged in, since it would mean users could "hang onto" Api instances, causing them to not get garbage collected. Instead all Api interactions happen within the client's methods, and not externally.

PR #366, opened 2021-06-08
Fixes for build and test process on linux vps

nerfZael:


dOrgJelli:
@nerfZael and I talked about this, and decided to not merge this right now, as it changes the default ganache port which is not ideal.

PR #408, opened 2021-07-12
Support GraphQL Subscriptions

### LINKED ISSUE #198: Support GraphQL Subscriptions
Support GraphQL subscription queries. The exact implementation & considerations are still to be determined.

Some initial thoughts:
- Subscriptions methods would be able to run in the background, dispatching results back to the caller.
- Specifying the subscription method's update frequency (allotted CPU time) would be nice.
- Research existing GraphQL subscription implementations.

Use Cases:
- Subscribe to a collection, receiving new items as they become available.
- Perform some arbitrary update logic that must run at some frequency.

krisbitney:
Implements a client.subscribe(...) function, which returns an AsyncIterator that returns values at a specified frequency (e.g. 1000ms) and can be used in a for await...of loop.

closes #198


kev-ngo (commented):
Hey @krisbitney , the following tests fail for me when I run `yarn test`.

image

kev-ngo (commented):
I also run into the following error when using `yarn prebuild` or `yarn build`:

image

krisbitney:
> I also run into the following error when using `yarn prebuild` or `yarn build`:
> 
> image

Thanks. I'll look into this and the other errors. :)

dOrgJelli:
Just double checked and it looks like RXJS now supports async iterators :D

https://github.com/ReactiveX/rxjs/pull/5297

dOrgJelli (approved):

PR #415, opened 2021-07-14
Support CLI Command Middlewares

### LINKED ISSUE #161: Support CLI Command Middlewares
For certain commands, there are things we'd like to do before they're executed.

For example, we have a number of commands that require docker to be available. We could have a middleware that we insert before certain commands that queries the system to see if it can gain access to docker. Ideally this would be done through a common stateful component / module that could retain the docker executable's location once it's queried within the middleware.

### LINKED ISSUE #165: CLI: Safely Handle Parallel Operations
Currently, if you run 2 parallel instances of `w3 test-env up`, there will be 2 simultaneous channels of communication with the docker daemon which will cause "duplicate container" errors.

In order to solve this, I propose we have a locking system in place for commands that cannot be run in parallel. This locking system would:
- Create a named lock file (ex: DOCKER_LOCK) when a sensitive operation is taking place.
- All other commands that require the lock will have to wait for it to be available.
- The lock file will be released when the command finishes its sensitive operations.

krisbitney:
This PR includes a MiddlewareHandler class that is added to the Gluegun Toolbox as a Gluegun extension, a middleware for verifying that docker is installed, and a middleware that implements a file lock to safely handle parallel operations.

Closes #161. Closes #165.


dOrgJelli (approved):
Truly phenomenal work @krisbitney :clap: 

FYI: currently we have `--concurrency 1` within the root `package.json` to avoid running more than one CLI command at once. After this PR gets merged, and a new CLI release is made, we should be able to remove this flag and run all tests in parallel!

krisbitney:
> Truly phenomenal work @krisbitney 👏
> 
> FYI: currently we have `--concurrency 1` within the root `package.json` to avoid running more than one CLI command at once. After this PR gets merged, and a new CLI release is made, we should be able to remove this flag and run all tests in parallel!

Thanks! 

I was just looking back at the PR files and I noticed a potential bug. In`request()`, I think there is a catch block that will never be thrown because the catch block in `release()` doesn't re-throw the exception.

In `request()`:
```typescript
      if (!isRunning) {
        try {
          await this.release();
          break;
        } catch (e) {
          // can ignore exception if cause is race condition
          if (!fs.existsSync(this.lockFilePath)) {
            throw e;
          }
        }
```
And `release()`:
```
  // release lock by deleting lock file
  async release(): Promise<void> {
    try {
      await fs.promises.unlink(this.lockFilePath);
    } catch {
      print.error("Tried to release a file lock that doesn't exist");
    }
  }
```

PR #422, opened 2021-07-23
Web3Api environment implementation

### LINKED ISSUE #140: Configure Web3APIs Using Env Variables
After talking with Luis from Pocket Network, it's become clear that we'll need a way to set configuration values for Web3APIs.

## Current Problem
There is currently one way to configure Web3APIs, and that's through method arguments. For example, think of an IPFS HTTP Client Web3API, you'd have to tell it what endpoint to talk to via a method argument:
```graphql
type Query {
  cat(
    hash: String!
    node: String!
  ): Bytes!
}
```

While this is simple and it works, this becomes problematic when you think about composability. If the IPFS Web3API is being used within another Web3API "Foo", you'd have to pass in the `node: String!` into each of Foo's methods, so it can pass it through to IPFS:
```graphql
#import { Query } into IPFS from "w3://ens/ipfs-http-client.web3api.eth"

type Query {
  doFoo(
    fooArg: [UInt32!]!
    ipfsNode: String!
  )
}
```

If you think about scaling this farther, it's clearly an anti-pattern, leading to N number of arguments needed to configure N number of dependencies.

NOTE: Web3APIs do not have persistent state (heap), meaning that the node URL cannot be stored in a global and used for subsequent method invocations.

## Proposed Solution
Enable the configuration of Web3APIs through env configuration variables. In the example above, this would allow the user to define an IPFS node's endpoint once.

## Developer Experience
*Still Very WIP...*

1. Web3API developer defines ENV variables using the schema.graphql
```graphql
type Env {
  requiredSetting: String!
  optionalSetting: UInt
}
```

2. Web3API developer uses ENV variables in WASM code
```typescript
import { Env } from "./w3";

Env.requiredSetting
```

3. Users configure the Web3API through the Web3API Client
```typescript
new Web3ApiClient({
  constants: [
    {
      uri: "w3://ens/some.api.eth",
      env: {
        requiredString: "Something Important",
        optionalString: nullableValue
      }
    }
  ]
});
```

Additional usability thoughts here: https://github.com/Web3-API/monorepo/issues/140#issuecomment-782519627

## Worries
Does this create "configuration hell"?

mpetrun5:
This pr adds: 
 - environement types to type info 
 - environment schema visitor
 - compose implementation of environment types
 - environment bind implementation
 - WasmWeb3Api loading of enviorment
 - enviroment config in WEb3ApiClient constructor

TODO in separate PRs:
 - environment sanitization
 - enviroment in plugins
 - query time enviroment defining

Refernces: #140


packages/js/client/src/wasm/WasmWeb3Api.ts
```diff
@@ -40,17 +41,21 @@ export class WasmWeb3Api extends Api {
     mutation?: ArrayBuffer;
   } = {};
 
+  private _sanitizedEnviroment?: ArrayBuffer;
```
Niraj-Kamdar: I see that environment sanitization will be done in separate PR so it's not implemented yet, right?

packages/schema/parse/src/typeInfo/index.ts
```diff
@@ -13,14 +13,31 @@ export * from "./scalar";
...
+}
+
 export interface TypeInfo {
   objectTypes: ObjectDefinition[];
   queryTypes: QueryDefinition[];
   enumTypes: EnumDefinition[];
   importedObjectTypes: ImportedObjectDefinition[];
   importedQueryTypes: ImportedQueryDefinition[];
   importedEnumTypes: ImportedEnumDefinition[];
+  environment: {
```
Niraj-Kamdar: We can use it this way once we merge it with master
```
environment: Record<InvokableModules, Environment>
```
Shouldn't we call it `environmentTypes` to make it consistent

packages/schema/parse/src/extract/object-types-utils.ts
```diff
@@ -1,13 +1,27 @@
...
+    type === EnvironmentType.QueryClientEnvType ||
+    type === EnvironmentType.QueryEnvType ||
+    type === EnvironmentType.MutationEnvType ||
+    type === EnvironmentType.MutationClientEnvType
+  ) {
+    return true;
+  }
+
+  return false;
+}
```
Niraj-Kamdar:
```suggest
export function isEnviromentType(type: string): boolean {
  return (
    type === EnvironmentType.QueryClientEnvType ||
    type === EnvironmentType.QueryEnvType ||
    type === EnvironmentType.MutationEnvType ||
    type === EnvironmentType.MutationClientEnvType
  );
}
```

Niraj-Kamdar: This can be simplified like this

packages/test-cases/cases/compose/sanity/input/mutation.graphql
```diff
@@ -35,6 +35,10 @@ implObject comment
   ): [Int32!]!
 }
 
+type MutationEnv {
```
Niraj-Kamdar, line 38: Should we use inheritance here? `type MutationEnv implements Env`

packages/schema/compose/src/resolve.ts
```diff
@@ -767,3 +773,45 @@ async function resolveLocalImports(
...
+      genericEnvProperties.find(
+        (genericProperty) => genericProperty.name === specificProperty.name
+      )
+    ) {
+      throw new Error(
+        `Type '${envType.type}' contains duplicate property '${specificProperty.name}' of type 'Env'`
+      );
+    }
+  }
+}
```
Niraj-Kamdar:
```suggest
export function checkDuplicateEnvProperties(
  envType: ObjectDefinition,
  genericEnvProperties: AnyDefinition[]
): void {
  const genericEnvPropertiesSet = new Set(genericEnvProperties.map((genericProperty) => genericProperty.name))
  for (const specificProperty of envType.properties) {
    if (genericEnvPropertiesSet.has(specificProperty.name)) {
      throw new Error(
        `Type '${envType.type}' contains duplicate property '${specificProperty.name}' of type 'Env'`
      );
    }
  }
}
```

Niraj-Kamdar: Finding in array takes linear time while in set it take constant time. So, overall complexity of the function reduces from `O(m * n)` -> `O(m + n)` where m and n are length of properties  for specific and generic env types.

Niraj-Kamdar (changes requested):
Left some questions and suggestions. Apart from that once we resolve the conflicts, it looks ready to go.

mpetrun5:
So sorry that I forgot about this. Been really busy with stuff for ChainSafe and haven't had the time to look at this. I'll try to help as much as I can, but I haven't been in the loop for half a year.

packages/js/client/src/wasm/WasmWeb3Api.ts
mpetrun5: TBH it has been so long since I have done this that I don't remember the idea at all, but I guess sanitization is done from this branch: https://github.com/polywrap/monorepo/pull/576

dOrgJelli:
No problem @mpetrun5, thank you for taking a look again. If you're too busy others will surely pick this up and get it merged :D

mpetrun5:
> No problem @mpetrun5, thank you for taking a look again. If you're too busy others will surely pick this up and get it merged :D

TBH I don't really have the time to finish this right now, but feel free to DM me in Discord if you have any question and I hope that I will be able to answer them.

lerna.json
```diff
@@ -4,7 +4,7 @@
   "version": "independent",
   "command": {
     "publish": {
-      "ignoreChanges": [
+      "ignore": [
```
Niraj-Kamdar: I found that there was a lerna warning of this being deprecated.

packages/js/client/src/__tests__/Web3ApiClient.spec.ts
```diff
@@ -1818,7 +1818,7 @@ describe("Web3ApiClient", () => {
     );
   });
 
-  it("environment types", async () => {
+  it.only("environment types", async () => {
```
Niraj-Kamdar: What's the reason for using `only` here?

dOrgJelli: Was for testing

lerna.json
dOrgJelli: Ah interesting, what command did you run to see this warning?

packages/test-cases/cases/compose/sanity/input/mutation.graphql
dOrgJelli: I thought about it, but decided against it. It actually ends up increasing the "hackery" we have to write to get it to work as intended.

It's better instead to treat `Env` as a "special" type name (like `Query` or `Mutation`). In the future we won't have to worry about this problem if we remove the coupling of Mutation & Query modules, and instead let developers define any independent module they want.

packages/schema/parse/src/typeInfo/index.ts
dOrgJelli: Changed to `envTypes`

dOrgJelli (approved):
Amazing work @mpetrun5 @Niraj-Kamdar :clap: :heavy_check_mark:

PR #427, opened 2021-08-06
Ethereum Smart Contract Struct Method Argument Fix

dOrgJelli:
Added a fix + a test, making sure Ethereum smart contracts that use structs within method arguments is functional.

## Fix
Within the Ethereum plugin, we needed to `JSON.parse(...)` structured objects, here's the code:
```typescript
public parseArgs(args?: string[]): unknown[] {
    if (!args) {
      return [];
    }

    return args.map((arg: string) =>
      (
        (arg.startsWith("[") && arg.endsWith("]")) ||
        (arg.startsWith("{") && arg.endsWith("}"))
      ) ? JSON.parse(arg) : arg
    );
  }
```

## Test
In the Ethereum plugin's tests, I've added a test smart contract that looks like this:
```solidity
contract StructArg {

  struct Struct { 
      string str;
      uint256 unsigned256;
      uint256[] unsigned256Array;
   }

  function method(Struct memory _arg) returns (string, uint256) {
    uint256 total = _arg.unsigned256;

    for (uint256 i = 0; i < _arg.unsigned256Array.length; ++i) {
      total += _arg.unsigned256Array[i];
    }

    return (_arg.str, total);
  }
}
```

and am invoking it within the `e2e.spec.ts` ethereum plugin tests like so:
```typescript
const response2 = await client.query<{ callContractMethodAndWait: Schema.TxReceipt }>({
  uri,
  query: `
    mutation {
      callContractMethodAndWait(
        address: "${address}",
        method: "function method(tuple(string str, uint256 unsigned256, uint256[] unsigned256Array) _arg) returns (string, uint256)",
        args: [$structArg]
      )
    }
  `,
  variables: {
    structArg
  }
});

expect(response2.errors).toBeUndefined()
expect(response2.data?.callContractMethodAndWait).toBeDefined()
expect(response2.data?.callContractMethodAndWait.transactionHash).toBeDefined()
```

## NOTE!
When defining the `method` ABI, we use the keyword `tuple` to define the structured argument. For example:
```solidity
struct Struct { 
  string str;
  uint256 unsigned256;
  uint256[] unsigned256Array;
}
```
becomes...
`tuple(string str, uint256 unsigned256, uint256[] unsigned256Array)`


krisbitney (approved):
Looks good! I linted with --fix and pushed a commit to pass the CI checks. (The force-push is because I messed up the commit message and then fixed it)

PR #434, opened 2021-08-17
Add JSON type support

### LINKED ISSUE #302: Support JSON as a Base Schema Type
Within API schemas, we've found multiple use cases for using JSON as a base type. Currently `String` is used in its place, but this makes it a bit ambiguous to the user.

Ideally `JSON` could be a supported base type, and in the future this could be implemented using our extendable base-type standard hypothesized here:
https://github.com/Web3-API/monorepo/issues/103

namesty:
closes #302


namesty:
I needed to make changes in the @web3api/wasm-as package, so temporarily, in order to test the changes, I needed to use my own @namestys/wasm-as package. When this PR gets reviewed and is ready to merge, we'll need to change the package in use back to @web3api/wasm-as

dOrgJelli (approved):

PR #447, opened 2021-08-22
Schema Binding: Handle Reserved Words For The Target Language

### LINKED ISSUE #420: Schema Binding: Handle Reserved Words For The Target Language
Currently, there is a bug in our binding code, where users can use reserved words in their GraphQL schemas that can't be used in the generated bindings.

For example:
```graphql
type Object {
  const: String!
}
```

The above schema uses a reserved word in Assemblyscript. The generated code would look like this:
```typescript
class Object {
  const: string;
}
```

:x: which does not compile

In order to fix this, we should create a list of "reserved words" in our binding code, and when one of them is used as a property (or argument), we should prefix it in some way that makes sense. For example, given the above example, the proper output would be:
```typescript
class Object {
  m_const: string;
}
```
where the `m_` is taken from hungarian notation.

The mustache template may look something like this:
```mustache
{{#properties}}
{{#toAscProp}}{{name}}{{/toAscProp}}: {{#toWasm}}{{toGraphQLType}}{{/toWasm}};
{{/properties}}
```
where `toAscProp` is a newly added function that checks a list of known keywords, and prefixes keywords with `m_`.

This same problem exists in Rust, and every other target language, so handling this consistently across all is ideal.

krisbitney:
Reserved words, such as "const", can now be used as properties and arguments in graphql schemas. The resulting AssemblyScript will compile. From the API user perspective, the query and return values can use the reserved words like so:
```
    const query = await client.query<{
      method1: {
        const: string;
      };
    }>({
      uri: ensUri,
      query: `
        query {
          method1(
            const: {
              const: "successfully used reserved keyword"
            }
          )
        }
      `,
    });
```

However, API developers must prefix reserved words with "m_" in their AssemblyScript code.

The implementation pattern can be used for additional wasm languages, such as Rust, by passing an appropriate array of reserved keywords to the "fromReservedWord" function and modifying the binding templates to use the result:
```
const data = Mustache.render(
            dirent.data,
            {
              ...(config as Record<string, unknown>),
              toAscProp: fromReservedWord(reservedWordsAS),
              toRustProp: fromReservedWord(reservedWordsRust),
            },
            subTemplates
          );
```

Closes #420.


dOrgJelli:
:clap: :clap: :clap: this looks awesome @krisbitney!!!

packages/schema/bind/src/utils/templateFunctions.ts
```diff
@@ -0,0 +1,11 @@
+export const fromReservedWord = (reservedWords: string[]) => {
+  return () => {
+    return (text: string, render: (text: string) => string): string => {
+      const rendered: string = render(text);
+      if (reservedWords.includes(rendered)) {
+        return "m_" + rendered;
```
Niraj-Kamdar, line 6: Just curious, what's the reason for prefixing it with `m_`

Niraj-Kamdar (dismissed):
LGTM!

packages/schema/bind/src/utils/templateFunctions.ts
krisbitney: It comes from Hungarian notation: https://en.wikipedia.org/wiki/Hungarian_notation

dOrgJelli (dismissed):

dOrgJelli (approved):

PR #453, opened 2021-08-30
Saving docker image as tar to build folder

### LINKED ISSUE #424: Put Docker Image Inside Build Folder
Keep the built docker image inside of a "build" folder (or some other folder). This way we're "hosting" docker images ourselves instead of relying on the docker hub.

namesty:
Closes #424


dOrgJelli:
Closing this for now, as I left some additional considerations here: https://github.com/polywrap/monorepo/issues/424

In its current state, I think this would hurt the tool-chains future to put the image in the build folder. Apologies for the naive guidance, but I think there's a pretty cool future ahead if we explore this further.

PR #454, opened 2021-08-30
feat: add version validators

Niraj-Kamdar:


packages/schema/parse/src/version_validate/comparaters.ts
```diff
@@ -0,0 +1,47 @@
+import { SetComparisionType } from "./setUtils";
+export function compare(obj1: unknown, obj2: unknown): SetComparisionType {
+  let result: SetComparisionType;
```
dOrgJelli: Be sure to use `const` for all variables that do not need to be modified after declaration.

packages/schema/parse/tsconfig.build.json
```diff
@@ -1,9 +1,5 @@
 {
   "extends": "./tsconfig.json",
-  "include": [
-    "./src/**/*.ts"
-  ],
-  "exclude": [
-    "./src/**/__tests__"
-  ]
+  "include": ["./src/**/*.ts"],
+  "exclude": ["./src/**/__tests__"]
```
dOrgJelli: Please turn off any auto-formatters you have running. Needless styling changes bloat PRs. Additionally the other tsconfig.build.json files are not formatted like this, making this one out of line w/ existing styling.

Same for the tsconfig.json file below in this changeset

packages/schema/parse/src/version_validate/schemas/outv1.json
```diff
@@ -0,0 +1 @@
...,[[[[["null","null",["4","uArrayArray","null","UInt"],"18","uArrayArray","null","null",["4","uArrayArray","null","UInt"],"[UInt]","null"],"null",["null","null",["4","uArrayArray","null","UInt"],"18","uArrayArray","null","null",["4","uArrayArray","null","UInt"],"[UInt]","null"],"18","uArrayArray","null","true","null","[[UInt]]","null"],"null","34","uArrayArray","null","true","null","[[UInt]]","null"],["null","null","34","optStr","null","null",["4","optStr","null","String"],"String","null"],["null","null","34","str","null","true",["4","str","true","String"],"String","null"],["null","null","34","u","null","true",["4","u","true","UInt"],"UInt","null"]],"64","method1","true",["null","null","34","method1","null","true",["4","method1","true","String"],"String","null"],"query"],[[[["null","null",["4","arg","true","String"],"18","arg","null","true",["4","arg","true","String"],"[String]","null"],"null","34","arg","null","true","null","[String]","null"]],"64","method2","true",[["null","null",["4","method2","true","Int32"],"18","method2","null","true",["4","method2","true","Int32"],"[Int32]","null"],"null","34","method2","null","true","null","[Int32]","null"],"query"]],"null","null","Query"]]]
```
dOrgJelli: In these arrays, you only have the values of each property in the object. This does not catch the case where property names are mismatched. I'd recommend making these arrays of key-value pairs `[key=>value, ...]` instead of arrays of values `[value, ...]`

packages/schema/parse/src/version_validate/setUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function difference<T>(set1: Set<T>, set2: Set<T>): Set<T> {
+  return new Set([...set1].filter((x) => !set2.has(x)));
+}
+
+export function symmentric_difference<T>(set1: Set<T>, set2: Set<T>): Set<T> {
+  return difference(union(set1, set2), intersection(set1, set2));
+}
+
+export function compareSets<T>(set1: Set<T>, set2: Set<T>): SetComparisionType {
```
dOrgJelli: Really interesting approach, can you add some comments to describe the theory behind this and why it makes sense?

packages/schema/parse/src/version_validate/schemas/outv1.json
Niraj-Kamdar: I changed the approach since the generic one wasn't covering all the edge cases.

packages/schema/parse/src/version_validate/setUtils.ts
Niraj-Kamdar: In set theory, we can prove that 2 sets are equal by taking the symmetric difference of 2 sets and checking that it's an empty set. Symmetric difference is difference between union and intersection of two sets. In case of equal sets both union and intersection will be always same hence giving empty set after symmetric difference

To check if a set is a subset of another set, we can take take difference of another set with respect to the first set and if it's an empty set it proves every element in the first set are present in the second set. If none of the above cases matches it can only mean one thing that, both the set are Different.

packages/schema/parse/src/version_compare/comparators.ts
```diff
@@ -0,0 +1,489 @@
...
+    result = Math.max(
+      result,
+      compareGenericTypes(obj1.interfaces, obj2.interfaces)
+    );
+    return result;
+  }
+  return VersionRelease.MAJOR;
+}
+
+export function compareObjectType(
```
nerfZael: I like the compare function per type approach
A potential problem I see is introducing a new property on the type, but then forgetting to add it to one of these functions

packages/schema/parse/src/version_compare/comparators.ts
```diff
@@ -0,0 +1,489 @@
+import {
```
nerfZael: I think "comparers" would be a better name for the file

also "schema-comparison" for the directory (note the use of dash instead of underscore. We use the dash convention in TS)

packages/schema/parse/src/version_compare/comparators.ts
```diff
@@ -0,0 +1,489 @@
...
+  ImportedEnumDefinition,
+  ImportedObjectDefinition,
+  ImportedDefinition,
+} from "../typeInfo/definitions";
+import { TypeInfo } from "../typeInfo";
+import { compareSets, SetComparisionType } from "./setUtils";
+
+export type ImportDefinition = { type: string };
+
+export enum VersionRelease {
```
nerfZael: The problem with this approach is that for example: 
when comparing schemas to see if they're valid as a patch update, as soon as we see in the algorithm that they are not, we don't need to keep checking if they're valid as minor or major. 

That said, I'm fine with this approach for now, I'll make a GH issue to improve this process later on

packages/schema/parse/src/version_compare/comparators.ts
Niraj-Kamdar: Agree but we wouldn't know how to categories and compare this newly added property and generic approach wouldn't work.

packages/schema/parse/src/version_compare/setUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function intersection<T>(set1: Set<T>, set2: Set<T>): Set<T> {
+  return new Set([...set1].filter((x) => set2.has(x)));
+}
+
+export function difference<T>(set1: Set<T>, set2: Set<T>): Set<T> {
+  return new Set([...set1].filter((x) => !set2.has(x)));
+}
+
+export function symmentric_difference<T>(set1: Set<T>, set2: Set<T>): Set<T> {
```
nerfZael: You're python background is showing :D
in TS it would be camel case: symmetricDifference()
also it's "symmetric" not "symmentric"

packages/schema/parse/src/version_compare/setUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+  return difference(union(set1, set2), intersection(set1, set2));
+}
+
+export function compareSets<T>(set1: Set<T>, set2: Set<T>): SetComparisionType {
+  const sd: Set<T> = symmentric_difference(set1, set2);
```
nerfZael: A faster way would be:
```typescript
if(set1.length > set2.length) {
  //No need to check anything else in this case, so we wan't to return early
  return SetComparisionType.DIFFERENT;
}

for(const item of set1) {
  if(!set2.has(item)) {
    return SetComparisionType.DIFFERENT;
  }
}

if(set1.length === set2.length) {
  return SetComparisionType.EQUAL;
} else {
  return SetComparisionType.SUBSET 
}
```

nerfZael: No need for other functions

packages/schema/parse/src/version_compare/comparators.ts
```diff
@@ -0,0 +1,489 @@
...
+  PATCH,
+  MINOR,
+  MAJOR,
+}
+
+export function compare(t1: TypeInfo, t2: TypeInfo): VersionRelease {
```
nerfZael: While I find it interesting to be able to know what the "VersionRelease" of two schemas is, there is no use for it currently(as we want to either just check for patch or just check for minor). If in the future there does arise some use for it then we can implement it. Right now, it's just an unnecessary performance drain.

Also, I'd say we are more interested in schema comparison than version comparison, since the former is a subset of the latter. So we need two functions: 
```typescript
export const areTypeInfosFunctionallyIdentical = (
  typeInfoA: TypeInfo,
  typeInfoB: TypeInfo
): boolean
```
and
```typescript
export const areTypeInfosBackwardCompatible = (
  baseTypeInfo: TypeInfo,
  derivedTypeInfo: TypeInfo
): boolean
```

Note the absence of minor and patch being mentioned

nerfZael (changes requested):

packages/schema/parse/src/version_compare/comparators.ts
nerfZael: Yea, a generic approach wouldn't.
A standard pattern is to have the compare function be in each of the types, so you would call it like this:
QueryDefinition.compare()
But, I'm not a 100% sure that it should be the responsibility of that type.
It would help in solving the problem, since it'd be a lot easier for the dev to know to update it.

Thoughts, @dOrgJelli ?

packages/schema/parse/src/version_compare/setUtils.ts
Niraj-Kamdar: Makes sense, I was following mathematical theorems but this looks better computationally :)

packages/schema/parse/src/version_compare/setUtils.ts
Niraj-Kamdar: Oops, typos and wrong conventions. Will fix it

packages/schema/parse/src/version_compare/comparators.ts
Niraj-Kamdar: I agree that creating compare method for every type would solve this issue. I believe from the OOP perspective it would make sense to have this as an object method. We can also leverage that inheritance property to reduce repetition in our compare functions.

packages/schema/parse/src/comparison/compareSets.ts
```diff
@@ -0,0 +1,24 @@
+export enum SetComparisonType {
```
nerfZael: I think SetComparisonResult is a better name
Type implies a type of comparison...

packages/schema/parse/src/comparison/comparers.ts
```diff
@@ -0,0 +1,903 @@
...
+  ImportedEnumDefinition,
+  ImportedObjectDefinition,
+  ImportedDefinition,
+} from "../typeInfo/definitions";
+import { TypeInfo } from "../typeInfo";
+import { compareSets, SetComparisonType } from "./compareSets";
+
+export type ImportDefinition = { type: string };
+
+export enum VersionRelease {
```
nerfZael: This should be in it's own file

packages/schema/parse/src/comparison/comparers.ts
```diff
@@ -0,0 +1,903 @@
...
+export interface CompareOptions {
+  shortCircuit?: VersionRelease;
+}
+
+export interface CompareResult {
+  versionRelease: VersionRelease;
+  hasShortCircuit: boolean;
+}
+
+export function compare(
```
nerfZael: Move this to compare.ts

packages/schema/parse/src/comparison/comparers.ts
```diff
@@ -0,0 +1,903 @@
...
+  if (importedQueryResult.hasShortCircuit) return importedQueryResult;
+  result.versionRelease = Math.max(
+    result.versionRelease,
+    importedQueryResult.versionRelease
+  );
+
+  return result;
+}
+
+export function compareGenericType(
```
nerfZael: It'd be a lot more readable if each of these was it's own file. It's generally a really good practice to keep files short. It's easier to find the logic you need from the file explorer and one can use that as a high level overview of the logic.
Something like:

src/comparison/VersionRelease.ts
src/comparison/compare.ts
src/comparison/comparers/compareGenericType.ts
src/comparison/comparers/compareQueryType.ts
...

packages/schema/parse/src/comparison/comparers.ts
```diff
@@ -0,0 +1,903 @@
...
+  MAJOR,
+}
+
+export interface CompareOptions {
+  shortCircuit?: VersionRelease;
+}
+
+export interface CompareResult {
+  versionRelease: VersionRelease;
+  hasShortCircuit: boolean;
```
nerfZael: I think hasShortCircuited is a bit clearer

packages/schema/parse/src/comparison/compapers/ArrayType.ts
```diff
@@ -0,0 +1,89 @@
+import { ArrayDefinition } from "../../typeInfo";
```
nerfZael: The dir should be named comparers not compapers :D
Also, in TS, it's better to either name the file by the component in it (if the components is a func: getTest then getTest.ts, if the component is a class SomeClass then SomeClass.ts (keeping the casing)

Alternatively if there are more components in a single file, which IMO should be avoided if possible. Then use dash casing: some-name.ts

For the current file that means either: array-type.ts or compareArrayType.ts
The former (array-type.ts) might be best since other files in this dir have multiple components, and there's also "compare" in the name of the dir "compare-rs" so it's kind of redundant.

packages/schema/parse/src/comparison/comparers/enum-type.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  for (let i = 0; i < arr1.length; i++) {
+    const obj1: EnumDefinition = arr1[i];
+    if (indexMap2.get(obj1.type) == undefined) {
+      result.versionRelease = VersionRelease.MAJOR;
+      result.hasShortCircuit = true;
+      return result;
+    }
+    const obj2: EnumDefinition = arr2[indexMap2.get(obj1.type) as number];
```
nerfZael, line 73: Is this necessary: indexMap2.get(obj1.type) as number 
indexMap2.get(obj1.type) doesn't return a number?

packages/schema/parse/src/comparison/comparers/enum-type.ts
```diff
@@ -0,0 +1,83 @@
...
+    hasShortCircuit: false,
+  };
+  const indexMap2: Map<string, number> = new Map<string, number>();
+  for (let i = 0; i < arr2.length; i++) {
+    indexMap2.set(arr2[i].type, i);
+  }
+
+  for (let i = 0; i < arr1.length; i++) {
+    const obj1: EnumDefinition = arr1[i];
+    if (indexMap2.get(obj1.type) == undefined) {
```
nerfZael: use === to compare

packages/schema/parse/src/comparison/comparers/generic-type.ts
```diff
@@ -0,0 +1,62 @@
+import { GenericDefinition } from "../../typeInfo";
+import { CompareOptions, CompareResult, VersionRelease } from "./utils";
+
+export function compareGenericType(
+  obj1: GenericDefinition,
+  obj2: GenericDefinition,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
nerfZael, line 7: since it's not used you can just remove it

Niraj-Kamdar: All the other compare functions have the same signature so I was just reluctant to have a different one for just this one.

packages/schema/parse/src/comparison/comparers/enum-type.ts
Niraj-Kamdar: It can be number or undefined so if I don't typecast I will get an error.

dOrgJelli:
Closing this PR for now as we've decided to hold off on implementing this.

PR #460, opened 2021-09-01
update uniswapv2 tests for polywrap prealpha.36

krisbitney:
Modifies e2e tests to work with prealpha.36 Polywrap client


dOrgJelli (approved):

dOrgJelli:
Hey @krisbitney, for some reason the swap e2e tests are throwing an Eth VM exception for me in the "EXACT_IN" swap call. Not sure why, maybe it has to do with different mainnet conditions...? For example, maybe the slippage is too high, or gas prices are too high, so the tx is reverting. Not sure.

Going to merge this anyway though since all other tests are working, and just that one swap (out of half a dozen or so) is failing.

PR #464, opened 2021-09-04
Allow for local npm package linking in docker build context

### LINKED ISSUE #425: Docker: Allow For Locally "linked" Packages
Currently, if you're making changes to the `@web3api/wasm-as` package locally, and then run a `w3 build ...` on a Web3API, the docker build image pulls `@web3api/wasm-as` from npmjs.com. This means that your local changes are not used.

In order to get around this, we should implement a way to "locally link" packages into the docker build image. Here's an initial spec of how we might implement this:
- new build-env "assemblyscript-dev"
- copy linked packages to ".w3/build/linked-packages". This is needed so the files are within the docker "build context".
- in docker file, copy them over, build them, yarn link them
- after install of API, yarn link all package names
- in the web3api.build.yaml, have a "linked-packages" config prop, where you only have to list the package name (ex: "@web3api/wasm-as")

We might not need to build them within the docker image, instead we could just assume the linked package is already built and simply run `yarn link`.

namesty:
This PR aims to close #425.

Enables users to declare, in their build manifest, npm packages that they want to link so that the docker build context doesn't try to pull them from the npm registry.

Implementation is as follows:

- Declaring package names and paths in build manifest
- Copying these packages to cache (.w3) folder
- Passing the copied package names and their routes in the cache folder to Dockerfile's mustache
- Each build-env Dockerfile contains language specific logic to handle linking. Currently only assemblyscript is supported.

This is easily extensible to other WASM implementation languages like Rust, without modifying compiler code


dOrgJelli:
Hey @namesty, I made the following changes:
- Upgraded all test-cases to use v40
- All test-cases use the local @web3api/wasm-as package now. This is necessary for CI/CD to pass, when version numbers get bumped for "relaese PRs".
- Renamed "linked-packages" to "linked_packages" per our existing manifest naming conventions (see "uri_redirects").
- Added the "filter" property to a "linked_packages" items. This allows users to specify a regex string to be used to filter out files. We're using this in the test project to filter out all test files.
- Moved all logic for caching linked packages into the Web3ApiProject class. See the `cacheBuildManifestLinkedPackages` and `getBuildManifest` functions.
- Stopped using "yarn link" and instead just set the package.json dependency to the local file path. See the `Dockerfile.mustache` file.

dOrgJelli (approved):

PR #467, opened 2021-09-11
Fetch dependencies and interfaces

### LINKED ISSUE #432: Client: Fetch Dependencies & Interfaces
Enable the fetching of dependencies & interfaces using the client. This feature will require the typeinfo JSON file being present in the package, so that parsing the GraphQL isn't required.

```typescript
const deps = client.getDependencies({
   uri: "/ens/domain.eth",
   module: "query" | "mutation" | undefined | "all",
   // include: DependencyType.Object|Query|Enum|Interface
   // ignore: DependencyType.Object|Query|Enum|Interface
});

deps[0].uri   // "w3://ens/another-domain.eth
deps[0].types /* [ {
   name: "Query",
   type: DependencyType.Object|Query|Enum|Interface,
   interface: true|false
} ] */
deps[0].namespace // "AnotherDomain"
```

### LINKED ISSUE #433: Client: Fetch Implemented Interfaces
Enable the fetching of implemented interfaces using the client. This will require having the TypeInfo JSON file present in the package.

```typescript
const imps = client.getImplementedInterfaces({
   uri: "/ens/domain.eth",
   // ignore: ImplementedType.Object|Query
});

imps[0].type // "Query"
imps[0].interfaces /* [ {
   uri: "w3://ens/another-domain.eth",
   namespace: "AnotherDomain"
   type: DependencyType.Object|Query
} ] */
```

krisbitney:
Implements two functions on the client.

`client.getDependencies(uri, options)` returns a list of dependencies given a uri. Options can be used to filter results by module, type, and whether a dependency is an interface.

`client.getImplementedInterfaces(uri, options)` returns a list of objects and queries that implement interfaces, along with information about the interfaces they implement. Options can be used to filter results to include only objects or only queries.

Closes #432. Closes #433.


packages/js/client/src/__tests__/Web3ApiClient.spec.ts
```diff
@@ -2118,5 +2118,254 @@ enum Logger_LogLevel @imported(
...
+    expect(includesConnection).toBeTruthy();
+
+    // check if Mutation methods are present
+    let includesMutation = false;
+    for (const type of dependencies[0].types) {
+      if (type.name === "Mutation" && type.type === DependencyType.Query && !type.interface) {
+        includesMutation = true;
+        break;
+      }
+    }
```
Niraj-Kamdar, line 2153:
```suggest
    const mutationIdx: number = dependencies[0].types.findIndex(
      (type) => ((type.name === "Mutation" && type.type === DependencyType.Query && !type.interface))
    );
    expect(mutationIdx).not.toBe(-1);
```

Niraj-Kamdar: We can use built-in function `findIndex ` here.

Niraj-Kamdar (dismissed):
LGTM!

dOrgJelli:
The client exposes a `getManifest(uri)` method, which can then be used to extract information from the `abi` field. I think we should recommend users use this to extract type information about the wrapper, instead of providing these additional helper methods. I'll close the associated issues for now, since they've become very stale.

PR #500, opened 2021-10-08
Feat: Multisend wrapper

cbrzn:


packages/apis/multisend/scripts/createPayload.js
```diff
@@ -0,0 +1,20 @@
+const { ethers } = require("hardhat");
+const fs = require("fs");
```
Niraj-Kamdar, line 2: What's the use of these scripts? If it's for just testing then we should create jest tests instead

packages/apis/multisend/deploy/00_delegator.js
```diff
@@ -0,0 +1,17 @@
+module.exports = async ({ deployments, getNamedAccounts }) => {
+  const { deployer } = await getNamedAccounts();
+  const { deploy } = deployments;
```
Niraj-Kamdar, line 3: Let's use typescript here.

Niraj-Kamdar (changes requested):
We should move this to integrations repo.

cbrzn:
> We should move this to integrations repo.

Hey @Niraj-Kamdar I totally agree with you - Currently, this wrapper is not working as expected because it only works with Contract Accounts and not with EOAs (because of the `msg.sender`). I will update this and use instead the multicall library (In the integrations repository)

dOrgJelli:
Closing this for now @cbrzn

PR #510, opened 2021-10-18
Improving regex capture in schema compose

nerfZael:


packages/schema/compose/src/resolve.ts
```diff
@@ -43,15 +43,17 @@ type ImplementationWithInterfaces = {
...
 
+const TYPE_NAME_REGEX = `[a-zA-Z0-9_]+`;
+
 export async function resolveImportsAndParseSchemas(
   schema: string,
   schemaPath: string,
   mutation: boolean,
   resolvers: SchemaResolvers
 ): Promise<TypeInfo> {
-  const importKeywordCapture = /^[#]*["{3}]*import[ \n\t]/gm;
```
nerfZael, line 52: using #* instead of #+ allowed imports that don't start with # to work, which I assume we don't want

packages/schema/compose/src/resolve.ts
```diff
@@ -43,15 +43,17 @@ type ImplementationWithInterfaces = {
...
-  const externalImportCapture = /[#]*["{3}]*import[ \n\t]*{([a-zA-Z0-9_, \n\t]+)}[ \n\t]*into[ \n\t]*(\w+?)[ \n\t]*from[ \n\t]*[\"'`]([\-a-zA-Z0-9_~.:\/]+?)[\"'`]/g;
-  const localImportCapture = /[#]*["{3}]*import[ \n\t]*{([a-zA-Z0-9_, \n\t]+)}[ \n\t]*from[ \n\t]*[\"'`]([a-zA-Z0-9_~\-:.\/]+?)[\"'`]/g;
```
nerfZael, line 54: we're now capturing anything between {} and between quotes where appropriate
also +? instead of + here didn't make much sense, someone correct me if I'm mistaken

packages/schema/compose/src/resolve.ts
```diff
@@ -43,15 +43,17 @@ type ImplementationWithInterfaces = {
...
+  const importKeywordCapture = /^#+["{3}]*import\s/gm;
+  const externalImportCapture = /#+["{3}]*import\s*{([^}]+)}\s*into\s*(\w+?)\s*from\s*[\"'`]([^\"'`\s]+)[\"'`]/g;
+  const localImportCapture = /#+["{3}]*import\s*{([^}]+)}\s*from\s*[\"'`]([^\"'`\s]+)[\"'`]/g;
```
nerfZael, line 56: using \s instead of [ \n\t]

packages/schema/compose/src/resolve.ts
```diff
@@ -107,6 +112,9 @@ export async function resolveImportsAndParseSchemas(
     .replace(externalImportCapture, "")
     .replace(localImportCapture, "");
 
+  // Remove all non documentation comments
+  newSchema = newSchema.replace(/#[^\n]*\n/g, "");
```
nerfZael, line 116: Normal comments before this change would break the parser if they contained keywords

dOrgJelli (approved):

PR #514, opened 2021-10-22
Union Types Support

### LINKED ISSUE #246: Support Union Types (GraphQL <> WASM)
Lacking unions will surely be a problem when porting SDKs to Web3API (especially those written in dynamic languages like JavaScript). We do not currently support unions as not all WASM compatible languages (AssemblyScript, Rust, etc) have native support for them, and where there is support it's rather unsafe (see Rust).

In order to get around this, I think we can simply support unions within any WASM compatible language by create **union specific variant objects**. For example, in GraphQL we may define a union like so:
`union Body = String | Bytes | FormData`

Body would be turn into a simple variant class which wraps the 3 possible types. In languages that support dynamic memory allocation and as a result have pointers (for ex C/C++), we can be space efficient (psuedo code):
```c++
enum BodyType {
  String,
  Bytes,
  FormData
}

class Body {
  public const BodyType m_type;
  private void* _instance;

  public constructor(BodyType type, MsgPackBuffer* input = 0)
  {
    this.m_type = type;

    switch (this.m_type) {
      case BodyType::String:
        this._instance = input ? deserializeString(input) : new String();
        break;
      case BodyType::Bytes:
        ...
    }
  }

  public getString() String {
    if (this.m_type != BodyType::String) {
      throw new Error(
        `Incorrect Body instance access. Tried to get "String", when instance is of type "${this.m_type}".`
      );
    }
    return String(*this._instance);
  }

  ...
}
```

In languages that don't support dynamic allocations (like AssemblyScript), and also don't have support for union types, we can use a less memory efficient pattern like so:
```typescript
enum BodyType {
  String,
  Bytes,
  FormData
}

class Body {
  public readonly type: BodyType;
  private _string: String | null;
  private _bytes: Bytes | null;
  private _formData: FormData | null;

  constructor(type: BodyType, input?: MsgPackBuffer) {
    this.type = type;

    switch (this.type) {
      case BodyType::String:
        this._string = input ? deseralizeString(input) : new String();
        break;
      case BodyType::Bytes:
        ...
    }
  }

  public getString(): String {
    if (this.type != BodyType::String) {
      throw new Error(
        `Incorrect Body instance access. Tried to get "String", when instance is of type "${this.type}".`
      );
    }
    return this._string;
  }

  ...
}
```

namesty:
Closes #246. This PR adds support for Union Types.

In order to pass a union argument, it has to comply with the following interface:

```ts
{
  type: string,
  value: $unionValue
}
```

Where `type` is the name of the union member type and `value` is the actual value.

**Example**:

Schema:
```graphql

type ObjA {
  prop: String!
}

type ObjB {
  propB: [String!]!
}

type Union = ObjA | ObjB

Query {
 foo(arg: Union): Boolean!
}
```

Client:
```ts
await client.query({
      uri: ensUri,
      query: `
        query {
          foo(
            arg: $arg
          )
        }
      `,
      variables: {
        arg: {
          type: "ObjA",
          value: {  prop: "bar"  },
        }
      }
    });
```


packages/schema/compose/src/resolve.ts
```diff
@@ -961,13 +1080,8 @@ async function resolveLocalImports(
...
         def: GenericDefinition,
-        rootTypes: EnumOrObject[]
-      ) => {
-        // Skip objects that we've already processed
-        if (typesToImport[def.type]) {
-          return def;
-        }
-
+        rootTypes: EnumOrUnionOrObject[]
+      ): EnumOrUnionOrObject => {
```
namesty, line 1079: Changed this function to be more similar to its external imports's equivalent. Felt this function should indeed find the imported definition and return it. Instead it was finding the import, mutating the `typesToImport` object and then returning the `def` parameter.

This made sense given that the visitor should return the `def`, but it ties the intention of this function to its visitor consumer and also, I needed an actual findImport function to retrieve the union's memberTypes object defs and import them

dOrgJelli:
Going to close this PR as we're working on supporting this within future WRAP ABI versions (>= 0.2)

PR #523, opened 2021-11-03
Rust/WASM Support

### LINKED ISSUE #217: Support Rust Based Web3APIs
It's becoming increasingly clear that we'll want to support another Web3API implementation language, other than Assemblyscript, in the near future. This is because there is a lot of code that we'd like to leverage that does not have an implementation in Assemblyscript. For example, the SHA3 library.

The beauty of adding support for Rust, is that you can also link and build C object files as well (though support is rather fringe as far as I can tell): https://github.com/rustwasm/team/issues/291#issuecomment-644946504

kobby-pentangeli:
**CHANGES**
This PR introduces Rust support for the toolchain. In particular, the following subdirectories contain the fundamental packages and/or modules allowing polywrappers to be built in Rust:

1. [The WASM Runtime](https://github.com/polywrap/monorepo/tree/pentangeli-wasm-rust/packages/wasm/rs)

2. [The Output Sanity Test Cases](https://github.com/polywrap/monorepo/tree/pentangeli-wasm-rust/packages/test-cases/cases/bind/sanity/output/wasm-rs)

3. [The Schema Bindings](https://github.com/polywrap/monorepo/tree/pentangeli-wasm-rust/packages/schema/bind/src/bindings/wasm-rs)

4. [The Example Polywrappers](https://github.com/polywrap/monorepo/tree/pentangeli-wasm-rust/packages/test-cases/cases/apis/wasm-rs)

**TEST PLAN**
Covered in the existing testing framework found at:
[Web3APIClient_spec](https://github.com/polywrap/monorepo/blob/pentangeli-wasm-rust/packages/js/client/src/__tests__/Web3ApiClient.spec.ts#L2206)


dOrgJelli:
TODO:
- [ ] Get rid of individual docker files for each API, instead move this into the CLI's default build-env folder.
- [ ] Modify the cargo.toml file of the query modules before building the mutation modules.

packages/cli/src/lib/build-envs/wasm/rust/Dockerfile.mustache
```diff
@@ -0,0 +1,79 @@
+FROM rust:{{rust_version}} as base
+
+# Install nightly rust
+RUN rustup toolchain install nightly
```
nerfZael, line 4: Not using a specific version would break source code verification once we implement it, I think
But, we can always cross that bridge when we get to it :D

packages/js/client/src/__tests__/wasm-rs.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+      ipfsProvider,
+      ensAddress
+    );
+
+    await TestCases.runEnumTypesTest(
+      await getClient(), `ens/testnet/${api.ensDomain}`
+    );
+  });
+
+  /*it("implementations - e2e", async () => {
```
nerfZael: Why is this commented? I'm guessing it doesn't work? :D

packages/schema/bind/src/bindings/wasm-rs/transforms/detectCircular.ts
```diff
@@ -0,0 +1,5 @@
+// TODO:
```
nerfZael: Just leaving a comment here to make sure this doesn't end up being forgotten
Or, if it's intended as a separate PR, then make sure to make an issue to track it

packages/schema/bind/src/bindings/wasm-rs/types.ts
```diff
@@ -0,0 +1,102 @@
...
+  typeof: "typeof",
+  unsized: "unsized",
+  virtual: "virtual",
+  yield: "yield",
+  try: "try",
+  macro_rules: "macro_rules",
+  union: "union",
+};
+
+export type KeyWords = typeof keywords;
```
nerfZael: Small nitpick: Keywords is one word (so not KeyWords)

packages/schema/bind/src/bindings/wasm-rs/types.ts
```diff
@@ -0,0 +1,102 @@
...
+  String: "String",
+  Boolean: "Boolean",
+  Bytes: "Bytes",
+};
+
+export type BaseTypes = typeof baseTypes;
+
+export type BaseType = keyof BaseTypes;
+
+export function isBaseType(type: string): type is BaseType {
```
nerfZael, line 22: I'd expect an isSomething function to return a boolean

packages/test-cases/cases/apis/wasm-rs/asyncify/mutation/src/lib.rs
```diff
@@ -0,0 +1,127 @@
+pub mod w3;
+use polywrap_wasm_rs::JSON;
+pub use w3::mutation;
+pub use w3::*;
+pub use w3::imported::ethereum_mutation;
+pub use query::imported::ethereum_query;
+
+pub fn set_data_with_large_args(input: mutation::InputSetDataWithLargeArgs) -> String {
```
nerfZael, line 8: If we wanted to keep this similar to the AssemblyScript version it'd be something like "Input_set_data_with_large_args" or maybe better "Input_SetDataWithLargeArgs"

kobby-pentangeli: In idiomatic Rust, we'd have to suppress some compiler warnings (specifically, `#[warn(non_camel_case_types)]`) in order to have a type like `Input_SetDataWithLargeArgs`. That would also mean all Rust Polywrap devs must write unidiomatic Rust code while building a wrapper. Thus, I think we should keep to the convention.

packages/cli/src/lib/build-envs/wasm/rust/Dockerfile.mustache
kobby-pentangeli: Are you referring to `{{rust_version}}` in the Mustache file? That's a placeholder for the specific version being used.

packages/schema/bind/src/bindings/wasm-rs/types.ts
kobby-pentangeli: Unless otherwise indicated, keeping it consistent with the AssemblyScript implementation: https://github.com/polywrap/monorepo/blob/prealpha/packages/schema/bind/src/bindings/wasm-as/types.ts#L16

packages/cli/src/lib/build-envs/wasm/rust/Dockerfile.mustache
nerfZael: I was referring to "RUN rustup toolchain install nightly" , though the {{rust_version}} has the same side effect since supposedly the build of the version can change. There was talk about it here https://youtu.be/uWBnO11Y7bc?t=1123 
Suggested fix was using hash of the version instead

packages/test-cases/cases/apis/wasm-rs/asyncify/mutation/src/lib.rs
nerfZael: Oh damn, that's a shame. I do agree, InputSetDataWithLargeArgs is fine then

packages/cli/src/lib/build-envs/wasm/rust/Dockerfile.mustache
kobby-pentangeli: Ah okay. `rustup toolchain install nightly` defaults to the latest nightly release, though.

nerfZael (changes requested):
I think you missed this: https://github.com/polywrap/monorepo/pull/523/files/4b20f4d952532ae73ff76f9134e42855281ce335#r763508655

PR #534, opened 2021-11-16
feat: add factory class

### LINKED ISSUE #532: Implement Interface_Query_Factory to create Interface_Query from any URI
Add a factory class that allows users to invoke any implementation of the given factory interface. It can be used in combination with getImplementations.

```ts
export class Interface_Query_Factory {
  public uri: string;

  constructor(uri: string) {
    this.uri = uri;
  }
  //...methods

}
```

It can be use like this in the polywrapper:
```ts
const implQuery = new Interface_Query_Factory(implUri)  // where implUri maybe fetched from getImplementations.
```

Niraj-Kamdar:
- closes: #532


dOrgJelli:
TODO:
- [ ] Add interface test to [bind sanity](https://github.com/polywrap/monorepo/tree/prealpha/packages/test-cases/cases/bind/sanity/input) test case. For example:
```graphql
#import { Query } into Interface from "interface.eth"

type Query implements Interface_Query {
  ...
}
```
- [ ] Merge the `Interface_Query` and `Interface_Query_Factory` classes. A rough example of what this might look like:
```typescript
class Interface_Query {
  static interfaceUri: string = "/ens/interface.eth";
  
  constructor(public uri: string) { }

  method(args: Input_method): string {
    w3_subinvoke(this.uri, etc);
  }
}
```

Niraj-Kamdar:
From what I understand, we are going to use `interfaces`(InterfaceImplementedDefinition[]) property  of `QueryDefinition` to see whether the imported namespace is interface or module, if it's imported and not in the `interfaces` it's a module and we won't be providing constructor for this and generating code for it like following so that user can directly use `Ethereum_Query.method` just like they've been doing.

```ts
class Ethereum_Query {
  static uri: string = "/ens/ethereum.eth";
  
  static method(args: Input_method): string {
    w3_subinvoke("/ens/ethereum.eth", etc);
  }
}
```
But in the case where it is in interfaces array, it will be interface and we will be generating code for this like:

```ts
class Interface_Query {
  static interfaceUri: string = "/ens/interface.eth";
  
  constructor(public uri: string) { }

  method(args: Input_method): string {
    w3_subinvoke(this.uri, etc);
  }
}
```
But doing it like this would require user to implement all the interfaces that he just want to use.
Ex: I have core polywrapper which uses token interface and for token interface there are USDC, DAI, and USDT 
implementation polywrappers. From OOP perspective, there is only inheritance from token -> USDC, token -> DAI and token -> USDT. while we don't need to have token -> core inheritance but just need composition instead. So, I don't think it's a good idea to mark the core polywrapper as implementation of token interface polywrapper.

Let me know what do you think and whether I get it wrong.

dOrgJelli:
Made an issue here to describe some future changes we might want to make: https://github.com/polywrap/monorepo/issues/547

dOrgJelli (approved):

Issue #561, opened 2021-12-01
LinkError: WebAssembly.instantiate(): memory import # is smaller than initial #, got #


dOrgJelli:
# Problem
As reported by @krisbitney, we've encountered this error while instantiating one of the UniV3 modules.
![image](https://user-images.githubusercontent.com/5522128/144185592-fd186456-aa08-4314-91d7-9d00ee16fa59.png)

The problem stems from this validation step within Chromium's WebAssembly run-time:
https://chromium.googlesource.com/v8/v8/+/644556e6ed0e6e4fac2dfabb441439820ec59813/src/wasm/module-instantiate.cc#924

It is being thrown when we run:
```typescript
const memory = WebAssembly.Memory({ initial: 1 });
const { instance } = await WebAssembly.instantiate(wasmBytecode, {
  env: {
    memory,
  },
  w3: w3Imports,
});
```

# Root Cause
After doing some digging into this, I've found that the problem stems from how Wasm modules expect their imported memory to be configured.

WebAssembly memory buffers must be configured with an `initial` number of memory pages (64 kb each). The module's bytecode is actually hard-coded with an expected number of initial pages:
`UniswapV2 query.wasm Module`
```wast
(module
  ...
  (import "env" "memory" (memory (;0;) 1))
  ...
)
```
and `UniswapV3 query.wasm Module`
```wast
(module
  ...
  (import "env" "memory" (memory (;0;) 2))
  ...
)
```
Notice the difference? The [`import`](https://github.com/sunfishcode/wasm-reference-manual/blob/master/WebAssembly.md#import-section) command for the `env.memory` object is defined differently, using the number `2` instead of `1`.

If we look at the [Wasm bytecode standard](https://github.com/sunfishcode/wasm-reference-manual/blob/master/WebAssembly.md), we can see that the memory import configuration is represented like so:
https://github.com/sunfishcode/wasm-reference-manual/blob/master/WebAssembly.md#linear-memory-description

If we change our above JavaScript code to:
```typescript
// Change initial to anything >= 2
const memory = WebAssembly.Memory({ initial: 2 });
const { instance } = await WebAssembly.instantiate(wasmBytecode, {
  env: {
    memory,
  },
  w3: w3Imports,
});
```
everything builds and runs just fine. Hooray! Now how do we make this future proof?

# Solution
[...]

PR #562, opened 2021-12-01
Fix: Issue #561 "LinkError: WebAssembly.Instantiate..."

dOrgJelli:
closes: https://github.com/polywrap/monorepo/issues/561


krisbitney (dismissed):
Looks good!

dOrgJelli:
CI is still not passing. It seems to be a regression / behavior change with docker, so unsure what we can do in our code to fix this. Have checked prealpha-dev and it is failing as well. Will merge regardless.

PR #576, opened 2021-12-13
Environment sanitization

mpetrun5:


packages/cli/src/lib/build-envs/wasm/assemblyscript/Dockerfile.mustache
```diff
@@ -13,6 +13,7 @@ WORKDIR /project
 # when just the Web3API schema & implementation files change
 COPY package.json .
 RUN yarn
+RUN yarn add @mpetrun5/wasm-as
```
mpetrun5: This as I remember was a problem at the time you couldn't use unreleased version of wasm-as

dOrgJelli: This is how you do so now @mpetrun5:
**web3api.build.yaml** - https://github.com/polywrap/monorepo/blob/949f31f0b0a1f8081d03f79ff3232566863511ae/packages/test-cases/cases/apis/bytes-type/web3api.build.yaml#L4-L6

**web3api.yaml** - https://github.com/polywrap/monorepo/blob/949f31f0b0a1f8081d03f79ff3232566863511ae/packages/test-cases/cases/apis/bytes-type/web3api.yaml#L2

dOrgJelli (approved):
:partying_face:

PR #579, opened 2021-12-16
client.invoke Sane Defaults (Decode & Input)

dOrgJelli:
Context: https://discord.com/channels/796821176743362611/814588073581084723/920812601036910704

### DIFF
--- a/packages/js/client/src/__tests__/Web3ApiClient.spec.ts
+++ b/packages/js/client/src/__tests__/Web3ApiClient.spec.ts
@@ -162,7 +162,7 @@ describe("Web3ApiClient", () => {
             networkNameOrChainId: "testnet"
           }
         },
-        decode: false
+        noDecode: true
       });
 
       expect(result.error).toBeFalsy();
--- a/packages/js/client/src/__tests__/Web3ApiClient.spec.ts
+++ b/packages/js/client/src/__tests__/Web3ApiClient.spec.ts
@@ -22,6 +22,7 @@ import {
   Client,
   PluginModules,
 } from "@web3api/core-js";
+import * as MsgPack from "@msgpack/msgpack";
 import { readFileSync } from "fs";
 
 jest.setTimeout(200000);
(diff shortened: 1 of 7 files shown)


krisbitney (dismissed):
Everything looks good. I wonder if it would be better to change `decode` to something like `noDecode` and reverse the true/false usages, so that undefined is the same as false?

packages/js/core/src/__tests__/resolve-uri.spec.ts
```diff
@@ -29,13 +29,9 @@ import {
 
 describe("resolveUri", () => {
   const client = (apis: Record<string, PluginModules>): Client => ({
-    getInvokeContext: (id: string) => {
```
nerfZael, line 32: How did this build before? Aren't the "getRedirects" and other methods required on the Client?

packages/js/client/src/__tests__/Web3ApiClient.spec.ts
```diff
@@ -121,6 +122,56 @@ describe("Web3ApiClient", () => {
...
+      const result = await client.invoke({
+        uri: ensUri,
+        module: "mutation",
+        method: "deployContract",
+        input: {
+          connection: {
+            networkNameOrChainId: "testnet"
+          }
+        },
+        decode: false
```
nerfZael: It's usually not a great practice to have to set params to false to override behavior
Defaults should be false by default

nerfZael (changes requested):

nerfZael:
> Everything looks good. I wonder if it would be better to change `decode` to something like `noDecode` and reverse the true/false usages, so that undefined is the same as false?

I'm 100% on board with this, not sure about the naming though (but am not against it)

(Why can't I reply to this comment but have to quote it, GitHub is weird)

packages/js/core/src/__tests__/resolve-uri.spec.ts
dOrgJelli: It didn't build before :P Jest was throwing warnings, but still running the tests just fine.

packages/js/client/src/__tests__/Web3ApiClient.spec.ts
dOrgJelli: Great point, as discussed I'll change this to `noDecode` instead.

PR #595, opened 2021-12-29
feat: add env for plugins

Niraj-Kamdar:


Niraj-Kamdar:
Discussed about this with @nerfZael and decided it would be nice if the dev experience of using env would be similar to wasm polytwrappers.

Niraj-Kamdar:
> Discussed about this with @nerfZael and decided it would be nice if the dev experience of using env would be similar to wasm polytwrappers.

Opened the issue #603 for supporting separate resolvers and w3 per modules for plugins to support this.

dOrgJelli (approved):

PR #600, opened 2022-01-04
feat: add support for custom configs for query command

Niraj-Kamdar:
closes: https://github.com/polywrap/monorepo/issues/244


Niraj-Kamdar:
This will replace #310

Niraj-Kamdar:
@dOrgJelli I think we should use `it` wherever we are using sentence for the test title otherwise we should use test

dOrgJelli (approved):

PR #604, opened 2022-01-08
Fix & Refactor: Plugin Codegen, ManifestLanguage, BindLanguage, SchemaKind

dOrgJelli:
Fix:
- Plugin codegen now properly picks up all dependencies from both Query & Mutation modules.
- `type Query` and `type Mutation` in plugin schemas should have duplicate `@imports`, since there is no way of isolating access to these dependencies. This will become an irrelevant problem if we move away from the `Query` & `Mutation` model, and instead use a singular `Module`. Users can still define multiple.

Refactor:
- Added the concept of `ManifestLanguage`, `BindLanguage`, and `SchemaKind` to better separate concerns and make the code more readable inside `@web3api/schema-composer`.


packages/cli/src/lib/project/Project.ts
```diff
@@ -79,4 +85,32 @@ export abstract class Project {
...
+    validPatterns: string[]
+  ): void {
+    if (!language) {
+      throw Error(intlMsg.lib_project_language_not_found());
+    }
+
+    const languagePatternValid = (test: string) => {
+      let valid = false;
+      validPatterns.map((x) => (valid = valid || test.indexOf(x) > -1));
+      return valid;
```
Niraj-Kamdar:
```suggest
      return validPatterns.every((x) => (test.indexOf(x) > -1));
```

packages/schema/compose/src/index.ts
```diff
@@ -34,24 +32,26 @@ export async function composeSchema(
...
   }
 
-  for (const name of Object.keys(schemas)) {
-    const schema = schemas[name];
+  for (const kind of schemaKinds) {
+    const schema = schemas[kind];
+
+    if (schema === undefined) {
+      continue;
+    }
```
Niraj-Kamdar, line 46: Why don't use `Object.entries(schemas)` here?

packages/schema/compose/src/resolve.ts
```diff
@@ -1028,41 +1042,41 @@ async function resolveLocalImports(
...
+  checkDuplicateEnvProperties(moduleEnvDef.sanitized, sharedEnvDef.properties);
+  moduleEnvDef.sanitized.properties.push(...sharedEnvDef.properties);
 }
 
 export function checkDuplicateEnvProperties(
   envType: ObjectDefinition,
-  genericEnvProperties: AnyDefinition[]
+  envProperties: AnyDefinition[]
 ): void {
   const genericEnvPropertiesSet = new Set(
```
Niraj-Kamdar: Did you missed this refactor? `genericEnvPropertiesSet` -> `envPropertiesSet `

Niraj-Kamdar (changes requested):
There are a few changes I requested otherwise this looks really good.

packages/cli/src/lib/project/Project.ts
krisbitney: In this case, I think the one liner should use `validPatterns.some(...)` instead of `.every(...)`

krisbitney (dismissed):

packages/cli/src/lib/project/Project.ts
Niraj-Kamdar: Ah my bad, It's `OR` haha

dOrgJelli: Amazing, thanks so much, did not think of using some here :D

packages/schema/compose/src/index.ts
dOrgJelli: It's because we require `kind` to be fully typed in this codeblock as `SchemaKind`. Doing Object.keys(...) or Object.values(...) removes this typing as it's too generic for our purpose here.

packages/schema/compose/src/resolve.ts
dOrgJelli: Yes, I felt it was a bit more confusing than it had to be using the word `generic`. This function is simply checking to see if an array of properties exist on a type. The type in this case being `envType`, and the properties being `envProperties`. It helps it be easier to read and understand IMO.

PR #610, opened 2022-01-14
Web3ApiClient - resolveUri

### LINKED ISSUE #326: Web3ApiClient URI Routing Optimization
In cases where there are multiple jumps when routing URIs (ex: A => B => C => D), we should cache each intermediate URI in the linked list to the final destination to avoid looking up the path over and over again.

In the example above, the cache would store a mapping of:
* `A => D`
* `B => D`
* `C => D`

Currently, the only route that gets cached in the client is `A => D`. This means that if someone provides the URI `B`, the URI resolution logic would have to run again, leading to the same path `B => C => D` being resolved. This is needless and should be cached to avoid duplicate lookups.

### LINKED ISSUE #431: Client: Get URI Resolution Path
Resolution Path:
```typescript
const path = client.getUriPath({
   uri: "/ens/domain.eth",
   // ignore: UriModifier.Redirects|Plugins
});

console.log(path[0].uri) // "w3://ens/domain.eth"
console.log(path[0].fromRedirect) // false
console.log(path[0].isPlugin) // false
console.log(path[1].uri) // "w3://ipfs/QmHASH"
```

Complex Resolution Path:
```typescript
const client = new Web3ApiClient({
   plugins: [{
      uri: "/ens/my-plugin.eth",
      plugin: myPlugin()
   }],
   redirects: [{
      from: "/ens/domain.eth",
      to: "/ens/my-plugin.eth"
   }]
});

const path1 = client.getUriPath({
   uri: "ens/domain.eth"
})

console.log(path1);
/*
[
   0: { uri: "w3://ens/domain.eth", fromRedirect: false, isPlugin: false },
   1: { uri: "w3://ens/my-plugin", fromRedirect: true, isPlugin: true }
]
*/

const path2 = client.getUriPath({
   uri: "ens/domain.eth",
   ignorePlugins: true
});

console.log(path2);
/*
[
   0: { uri: "w3://ens/domain.eth", fromRedirect: false, isPlugin: false },
   1: { uri: "w3://ens/my-plugin", fromRedirect: true, isPlugin: false },
   2: { uri: "w3://ipfs/QmHASH", fromRedirect: false, isPlugin: false }
]
*/
```

### LINKED ISSUE #581: Resolve-URI: Not Respecting Cache Or Redirects
When a URI is resolved to another URI, the redirects are cache are ignored.

### LINKED ISSUE #583: Add Client "resolveUri" Method
Add a utility method to the client for resolving a URI. The intention is for this method to be used to introspect the URI resolution process.

@nerfZael is working on a new URI resolver, and would like to write tests verifying it's working correctly "under the hood". This method would enable that.

The developer experience should be something like:
```typescript
resolveUri(uri: string | Uri) => ResolveUriDertails[]
```
where `ResolveUriDertails` could take the form of:
```typescript
interface ResolveUriDertails {
  // The uri-resolver used
  uriResolver: Uri;

  // Input & results of the `tryResulveUri` method
  tryResolveUri: {
    input: {
      authority: string;
      path: string;
    };
    result: {
      uri?: string | null;
      manifest?: string | null;
    };
  }
}
```

The array of details structures represents the "resolution chain" that occurs when the resolve URI process is undergone.

Additional information can be added to this details structure as needed, such as:
- timings
- errors

nerfZael:
Closes #583 , closes #581 , closes #326 , closes #431


packages/js/core/src/algorithms/uri-resolution/resolvers/UriResolutionResult.ts
```diff
@@ -0,0 +1,6 @@
+import { Api, Uri } from "../../..";
+
+export type UriResolutionResult = {
```
Niraj-Kamdar: This should be moved to types

packages/js/core/src/algorithms/uri-resolution/resolvers/UriToApiResolver.ts
```diff
@@ -0,0 +1,12 @@
+import { UriResolutionResult } from ".";
+import { UriResolutionStack } from "..";
+import { Client, Uri } from "../../../types";
+
+export interface UriToApiResolver {
```
Niraj-Kamdar: This should also be moved to types. I think we should only keep resolvers inside the resolvers directory

Niraj-Kamdar (dismissed):
Apart from these 2 quick fixes, this looks good! 
It would be nice if we can soon support `local` or `fs` authority so that we can specify plugins as local authority to avoid confusions.

packages/js/client/src/Web3ApiClient.ts
```diff
@@ -104,6 +109,10 @@ export class Web3ApiClient implements Client {
     }
   }
 
+  public getApiCache(): ApiCache {
```
nerfZael: Note: I am exposing cache so that the CacheResolver can use it via the client

Niraj-Kamdar: We should also trace this.

packages/js/client/src/Web3ApiClient.ts
```diff
@@ -419,6 +435,60 @@ export class Web3ApiClient implements Client {
     return subscription;
   }
 
+  @Tracer.traceMethod("Web3ApiClient: resolveUri")
```
Niraj-Kamdar, line 434: Why this haven't been added to contextualizeClient?

packages/js/core/src/types/Client.ts
```diff
@@ -49,6 +54,8 @@ export interface Client
   extends QueryHandler,
     SubscriptionHandler,
     InvokeHandler {
+  getApiCache(): ApiCache;
```
Niraj-Kamdar: Shouldn't we also have getResolvers?

packages/js/core/src/uri-resolution/resolvers/plugin/PluginResolver.ts
```diff
@@ -0,0 +1,47 @@
...
+
+      const api = this.createPluginApi(uri, plugin, environment);
+
+      return {
+        uri,
+        api,
+      };
+    }
+
+    return Promise.resolve({
```
Niraj-Kamdar: Why are we returning Promise.resolve instead of plain return? I maybe wrong but since the function is already async it may not be needed.

Niraj-Kamdar (dismissed):
LGTM! Please look at the comments.

packages/js/core/src/types/Client.ts
nerfZael: Yes we should!

packages/js/client/src/Web3ApiClient.ts
nerfZael: Will add, nice catch

packages/js/client/src/Web3ApiClient.ts
nerfZael: Yep

packages/js/core/src/uri-resolution/resolvers/plugin/PluginResolver.ts
nerfZael: True!

packages/js/core/src/types/Client.ts
```diff
@@ -49,6 +60,8 @@ export interface Client
...
```
dOrgJelli: I don't think we should expose the ApiCache. This is a sensitive part of the client that has yet to be fully fleshed out: https://github.com/polywrap/monorepo/issues/327

I feel like we should instead pass it internally to the resolvers directly. This is a better ownership relationship IMO (client owns cache, gives reference  resolvers [special permission]).

This way we don't have to worry about every-day users of the client having access to the cache. Also when we make changes to the `ApiCache` interface in the future, we don't have to worry about breaking users.

Will push a proposed change.

dOrgJelli (changes requested):
Requested changes here: https://github.com/polywrap/monorepo/pull/680

Issue #26, opened 2020-08-05
CLI Validate Build Files


dOrgJelli:
Run validation over all build files (schema, wasm modules, manifest).

dOrgJelli:
This is being handled by:
- Schema Validation
- Manifest Validation
- WASM Build File Validation

PR #617, opened 2022-01-20
Use https://ipfs.wrappers.io IPFS Provider

dOrgJelli:
- Use https://ipfs.wrappers.io provider
- Reduce hard-coded IPFS providers in the rest of the codebase by using the exported `defaultIpfsProviders` from the `@web3api/client-js` package.

### DIFF
--- a/packages/js/react/src/__tests__/plugins.ts
+++ b/packages/js/react/src/__tests__/plugins.ts
@@ -1,4 +1,5 @@
 import { PluginRegistration } from "@web3api/core-js";
+import { defaultIpfsProviders } from "@web3api/client-js";
 import { plugin as ensPlugin } from "@web3api/ens-plugin-js";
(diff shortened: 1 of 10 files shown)


packages/js/react/src/__tests__/plugins.ts
```diff
@@ -23,7 +24,7 @@ export function createPlugins(
       uri: "w3://ens/ipfs.web3api.eth",
```
nerfZael, line 24: Why isn't the IPFS plugin included by default, why do we need to specify it here?

nerfZael (approved):

packages/js/react/src/__tests__/plugins.ts
dOrgJelli: It's so that we can add our own custom `ipfsProvider` for the tests. If line #26 wasn't present, then this would not be needed.

PR #620, opened 2022-01-24
feat: improve plugin development experience 

### LINKED ISSUE #603: Plugin should have module-wise resolvers similar to the wasm
Currently, when we create a plugin, we will just create a single schema file, a Plugin class and resolvers for both query and mutation module. This will create issues while porting some features from wasm polywrapper to plugin that are module dependent Ex: env 

In wasm polywrapper, we can import `env` directly from the `w3` and use it since the both query and mutation modules will always be decoupled in the separate files and we can correctly have env of either `QueryEnv` or `MutationEnv` type depending on the module but in case of `plugin`, since both modules are in same file and we only generate one single `w3`, we can have `env` with both the `QueryEnv` and `MutationEnv` types so we can't assign it either one of the type, though we can workaround this by creating 2 separate variables: `queryEnv`, `mutationEnv` but that wouldn't be a good and consistent experience. Not to mention in the future we may have more than 2 modules. 

So a better solution would be to create separate directories for every module with its own w3 generated codes in plugin just like we have it in wasm polywrappers. 

Ex: In plugin we can have classes for all the available modules: 
```ts
// Ethereum_Query
import { Input_getGasPrice } from "w3";
class Query {
  constructor(private readonly configs: QueryConfigs) // configs if needed 
  ...
  getGasPrice: async (input: Input_getGasPrice): Promise<string>
  ...
}
```
This will replace current resolvers with module wise classes, we can pass state in constructor from the Plugin class, this state can even be shared or isolated depending on how plugin devs want it to be implemented. 

Ex: Plugin class would just be implementation of the core Plugin interface without any other methods.
```ts
export class EthereumPlugin extends Plugin {
  constructor(config: EthereumConfig) 
  ...
  public static manifest(): PluginPackageManifest {
  ...
  }

  public getModules(
    _client: Client
  ): {
    query: Query;
    mutation: Mutation;
  } {
    return {
      query: () => new Query(this.config),
      mutation: () new Mutation(this.config),
    };
  }
```

We may not even need configs for most of the cases after env support since we can pass it using env from the client, although we would still need it in some cases where user wants to pass an object instance instead of string serializable env so for that having it is still necessary.

Niraj-Kamdar:
- Closes: #603 

This PR implements new Plugin dev exp spec as discussed in the above issue.

**Open questions:**
- Should we rename loadEnv -> setEnv ?
- Should we move name from metaManifest to Manifest ( for both Plugin and Wasm) ?


packages/cli/src/__tests__/plugin/expected-schema/schema.graphql
```diff
@@ -54,25 +51,25 @@ type Query @imports(
...
-    "Ethereum_Access"
+    "Ethereum_TxRequest"
   ]
 ) {
   method(
     arg: UInt32!
   ): String!
 }
 
+type QueryEnv {
```
nerfZael: What about the MutationEnv?

packages/js/client/src/plugin/PluginWeb3Api.ts
```diff
@@ -88,7 +88,13 @@ export class PluginWeb3Api extends Api {
           )) as Record<string, unknown>;
         }
 
-        this._getInstance().loadEnv(env, module);
+        const plugin = this._getInstance();
+        if (plugin && plugin.loadEnvByModule) {
+          plugin.loadEnvByModule(module, env);
+        } else {
+          // Deprecated function
```
nerfZael: We should create an issue to remove this once all plugins are migrated

packages/js/plugins/ethereum/web3api.plugin.yaml
```diff
@@ -1,3 +1,6 @@
 format: 0.0.1-prealpha.1
 language: plugin/typescript
 schema: ./schema.graphql
+import_redirects:
+  - uri: "ens/logger.core.web3api.eth"
```
nerfZael: We dont have the logger interface deployed?

packages/js/plugins/ethereum/schema.graphql
```diff
@@ -1,3 +1,6 @@
+# import { Query, LogLevelEnum } into Logger from "ens/logger.core.web3api.eth"
+# import { Query, Ping } into Coingecko from "ens/coingecko.defiwrapper.eth"
```
nerfZael: Why has this dependency been added?

packages/schema/bind/src/bindings/plugin-ts/createPluginContext.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export type PluginContext = {
+  className: string;
+  funcName: string;
+  modules: PluginContextModules;
+};
+
+export function createPluginContext(opts: PluginContextOptions): PluginContext {
+  const funcName = camelCase(
+    !opts.metaManifest ? generateName() : opts.metaManifest.name
```
nerfZael: I think we should just use a default name instead of generating a random one

packages/js/core/src/algorithms/index.ts
```diff
@@ -5,3 +5,4 @@ export * from "./find-plugin-package";
 export * from "./get-implementations";
 export * from "./parse-query";
 export * from "./resolve-uri";
+export * from "./generate-name";
```
nerfZael: I think this belongs back in test-env, the core shouldn't really have random name generation

nerfZael (changes requested):

packages/cli/src/__tests__/plugin/expected-schema/schema.graphql
Niraj-Kamdar: It's just e2e tests. Both QueryEnv and MutationEnv  are used in schema-bind sanity unit-tests, we should have more `schema-bind` tests though.

packages/js/client/src/plugin/PluginWeb3Api.ts
Niraj-Kamdar: @dOrgJelli already created one for migrating plugins and we will be removing this in the same PR.

packages/js/plugins/ethereum/web3api.plugin.yaml
Niraj-Kamdar: Yes we do but we use redirects instead for faster testing,
Also this is just debug leftover - nice catch.

packages/schema/bind/src/bindings/plugin-ts/createPluginContext.ts
Niraj-Kamdar: I think it would work just a side note that if he creates 2 plugins without meta-manifest and want to test it using client he need to use alias or something while importing this plugins. Not a big deal though and having meta-manifest while building plugins should always recommended anyways.

packages/cli/src/__tests__/plugin/expected-schema/schema.graphql
Niraj-Kamdar: Let me go ahead and create an issue for adding more tests for `schema-bind`

packages/schema/bind/src/bindings/plugin-ts/createPluginContext.ts
nerfZael: True, and we could even move the name to the main manifest, I think I saw a suggestion like that somewhere

packages/cli/src/__tests__/plugin/expected-schema/schema.graphql
Niraj-Kamdar: created issue here: #649

packages/schema/bind/src/bindings/plugin-ts/createPluginContext.ts
Niraj-Kamdar: I am not so sure on this, because I think we can normalise all API type (WASM and Plugin) Manifest to share same required configs and we can have a single CodeGenerator for both Wasm and Plugin, since both have so much similarity after we added support for modules in plugin.

dOrgJelli (approved):

PR #621, opened 2022-01-25
IPFS resolver fixes and parallelization

nerfZael:


packages/js/plugins/ipfs/src/exec-ipfs/CancelablePromise.ts
```diff
@@ -0,0 +1,5 @@
+export type CancelablePromise<TReturn> = {
```
Niraj-Kamdar: Move this to types

packages/js/plugins/ipfs/src/exec-ipfs/cancelableExecIpfs.ts
```diff
@@ -0,0 +1,74 @@
...
+    provider: string,
+    options: unknown
+  ) => Promise<TReturn>,
+): CancelablePromise<TReturn> => {
+  let controller = new AbortController();
+  
+  let error: any = undefined;
+  
+  //If timer is not 0 then set a timeout to abort the execution
+  const timer: NodeJS.Timeout | undefined = timeout
```
Niraj-Kamdar: Will this type anotation work on browser environment?

Niraj-Kamdar: I think letting it infer type would be better

packages/js/plugins/ipfs/src/exec-ipfs/execIpfsWithProviders.ts
```diff
@@ -0,0 +1,194 @@
...
+    cancelAllRequests(requests);
+  } else {
+    //Throw all aggregated errors
+    throw new Error(errors.map(x => x.message).join("\n"));
+  }
+
+  return response.result as TReturn;
+};
+
+const gatherSuccessPromises = async <TReturn>(
```
Niraj-Kamdar: This is a neat trick!

Niraj-Kamdar (dismissed):
LGTM! Really like the clever use of promises!

packages/js/plugins/ipfs/src/exec-ipfs/CancelablePromise.ts
nerfZael: I can add a new dir here in exec-ipfs called types, since I don't intend on exporting this out

packages/js/plugins/ipfs/src/exec-ipfs/cancelableExecIpfs.ts
nerfZael: cant infer because clearTimeout won't work then, but I'll try and work around it

packages/js/plugins/ipfs/src/exec-ipfs/execIpfsWithProviders.ts
nerfZael: Was fun coding it :)

packages/js/plugins/ipfs/schema.graphql
```diff
@@ -24,10 +24,25 @@ type ResolveResult {
 }
 
 type Options {
```
Niraj-Kamdar, line 26: It'd be nice if we can have inside the ENV since all the call can share this vars, We also have option to override env at query time so I don't see any downside in using ENV here.

Niraj-Kamdar (changes requested):

Niraj-Kamdar (changes requested):
Check the comment

packages/js/plugins/ipfs/schema.graphql
nerfZael: I'd agree if we had the IPFS resolver and IPFS sdk plugins separated, but right now it gets confusing.
The Options here are used for the IPFS sdk part and is configurable per request (while using a single provider) while the QueryEnv down below is used for IPFS resolver part.

When we separate them we can have Env for both and configure them without the overlapping mess. 
TLDR: I'd hold off on this until we separate the plugins and leave this as is right now
Thoughts?

Niraj-Kamdar: That makes sense! And TBH it will only be this single release for which we will have this. In next release we need to refactor plugins anyway!

dOrgJelli (approved):

PR #650, opened 2022-01-27
feat (cli query): add yaml support for recipe parsing

### LINKED ISSUE #629: Support YAML for Recipe Scripts
Currently recipe scripts are written in JSON, but I think for all user-defined POD files the CLI uses we should support both JSON & YAML. This is because YAML has many user-friendly features, and we would simply convert to JSON after it gets loaded from disk.

arcbjorn:


packages/cli/package.json
```diff
@@ -57,7 +57,8 @@
     "regex-parser": "2.2.11",
     "rimraf": "3.0.2",
     "typescript": "4.0.7",
-    "ws": "7.3.1"
+    "ws": "7.3.1",
+    "yaml": "^1.10.2"
```
dOrgJelli: Be sure to always lock dependency versions. Please remove the `^`.

packages/cli/yarn.lock
```diff
@@ -0,0 +1,6884 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
dOrgJelli: yarn.lock files should not be generated / committed within package directories. Please read up on how "lerna + yarn monorepos" work here: https://leerob.io/blog/monorepo-lerna-yarn-workspaces

The steps to install dependencies, build packages, and run tests can be found in the root README of the monorepo. Additionally after you install deps + build packages, you can run tests within a single package by opening a terminal there and running `yarn test` or `yarn test:ci`. Feel free to send messages in the discord if you'd like some help getting adjusted to this setup.

packages/cli/package.json
dOrgJelli: There's no need to add another yaml package, we use the `js-yaml` package in the CLI package, as well as in the rest of the packages in the monorepo that need to serialize / deserialize YAML.

dOrgJelli (changes requested):

dOrgJelli (approved):

dOrgJelli:
Amazing! Thanks so much for the thorough testing!

PR #666, opened 2022-02-07
Added Tezos based assemblyscript template to Polywrap's Monorepo

von-dee:
We added an assemblyscript template that supports development and testing of poly wrappers on the Tezos Blockchain. 

Create your project
You can create a Tezos AssemblyScript template project by running the following command:

npx @web3api/cli create api tezos <project-name>

Where <project-name> is replaced with a custom name of your choice. For example my-tezos-wrapper.

Blockwatch.cc


dOrgJelli:
Thanks so much for this PR! We have a feature planned which allows anyone to publish new project templates, and use them with the CLI. You can read more about it here: https://github.com/polywrap/monorepo/issues/585

The idea is to at first support publishing templates to NPM or git repositories.

The syntax would probably look something like `w3 create --template git:https://github.com/organization/repository` or `w3 create --template npm:@org/package:1.2.3`

If this PR is not a blocker for you right now, I think it might be best to wait to add this Tezos specific template using the above feature. If it is blocking, then we can probably just merge this now, and remove it in the future (which would be a breaking change to users who use this).

Thoughts?

von-dee:
Thank you. Will wait for the --template feature then.

Thanks for the response

PR #671, opened 2022-02-08
653 | ref (cli codegen): custom template naming

### LINKED ISSUE #653: Rename customScript to customTemplate in cli codegen
customScript is really confusing name in this context and should be renamed to customTemplate: 
https://github.com/polywrap/monorepo/blob/prealpha/packages/cli/src/commands/codegen.ts#L125

arcbjorn:


krisbitney:
What is the reason for this?

krisbitney:
This may be confusing because we have another thing called a `template`, used in the cli `create` command

arcbjorn:
> This may be confusing because we have another thing called a `template`, used in the cli `create` command

@Niraj-Kamdar what do you think ?

Niraj-Kamdar:
This is subcommand for the `codegen`, and I think it would be obvious in this context , and having it named custom-script would be more confusing since we expect mustache template here. Also, we use `--custom` option here.

dOrgJelli:
This has been recently changed to `--script` here: https://github.com/polywrap/monorepo/blob/6d1f0983d305bcc6c763d11c8e35d73b37418f3c/packages/cli/src/commands/codegen.ts#L39

Let me know if you feel this is not a good name. Going to close this PR for now.

PR #673, opened 2022-02-08
661 | feat (cli query): improve error handling

### LINKED ISSUE #661: CLI: `w3 query ...` Should Return Non-Zero Error Code When Query Fails
Currently when an error throws an error, the CLI still returns a 0 exit code, meaning that automated CI steps think it succeeded successfully.

arcbjorn:


packages/cli/src/commands/query.ts
```diff
@@ -226,6 +226,10 @@ export default {
             print.error("-----------------------------------");
           }
         }
+
+        if (errors) {
+          process.exitCode = 1;
+        }
       }
```
Niraj-Kamdar: Why another condition? We should add this after printing errors

Niraj-Kamdar (changes requested):

packages/cli/src/commands/query.ts
arcbjorn: Yes, did some experimenting with types. Fixed now, thank you

dOrgJelli (approved):

PR #674, opened 2022-02-10
Wasm Sub-Invoke Error Handling

### LINKED ISSUE #643: Support Better Error Handling in Wasm
Currently when a wasm module performs a "subinvoke" and calls into another wasm module, there's no way of programatically handling the error that's returned. Instead an exception is thrown.

This is not suitable for situations where you want to call a function, see if it fails, and then call another function. For example: https://github.com/polywrap/monorepo/blob/8d659a9101e6097c659c56385871a980af9dcacc/packages/js/plugins/ipfs/src/resolvers.ts#L25-L50

dOrgJelli:
closes: https://github.com/polywrap/monorepo/issues/643

# Features
## No-Throw Sub-Invocations (Custom Error Handling)
Instead of calls like this throwing if there is an error:
```typescript
// This throws an exception...
const res = Ethereum_Query.callContractView({
  method: "function badFunctionCall() view returns (uint256)",
  address: input.address,
  args: null,
  connection: input.connection
});

// AssemblyScript / Wasm doesn't support try / catch,
// so code below this line will never get run,
// meaning we cannot handle errors if they occur. 
```

Callers will now be given a `Result<TData, TError>` structure, allowing developers to handle errors themselves:
```typescript
const res = Ethereum_Query.callContractView({
  method: "function badFunctionCall() view returns (uint256)",
  address: input.address,
  args: null,
  connection: input.connection
});

res.isErr; // true
res.unwrap(); // throws
res.unwrapErr(); // error string "VM Exception Occured: ..."
```

## Result<TData, TError> (like Rust's) Added to AssemblyScript Wasm Runtime
I've found the following repository that implements a few useful Rust collections (`Result`, `Option`, `Box`) in AssemblyScript:
https://github.com/yjhmelody/as-container

We are using the `Result` type from this library. It works like a charm, and shares the same interface as Rust's `Result` structure:
https://doc.rust-lang.org/stable/std/result/

Users can use commonly understood functions like `unwrap()`, `expect("...")`, and so on.

# Breaking Change
This is a breaking change. All calls to external wrappers as shown in the above code snippets will now require the caller to get the return value by going through the `Result<T, E>` object's interface (unwrap, expect, etc).


dOrgJelli:
Merging since we discussed this thoroughly in the Dev Sync on Feb 10th 2022:
https://discord.com/channels/796821176743362611/916046622742446100/941387981137252402

PR #677, opened 2022-02-10
docs (schema bind readme): fix typo

arcbjorn:


packages/schema/bind/README.md
```diff
@@ -4,29 +4,29 @@ Generates Web3API schema bindings for supported guest languages.
...
-| Int8 | int 8 | 8-bit signed integer. |
-| Int16 | int 16 | 16-bit signed integer. |
-| Int32 | int 32 | 32-bit signed integer. |
-| String | fixstr or str 8/16/32 | UTF-8 string. |
-| Boolean | bool | true or false stored as 1 byte. |
-| Bytes | bin 8/16/32 | array of 8-bit unsigned integer. |
-| BigInt | fixstr or str 8/16/32 | UTF-8 string. |
-| JSON | fixstr or str 8/16/32 | UTF-8 string. |
-| [Type] | fixarray or array 16/32 | Array of elements. |
-| type CustomObject {<br/>&nbsp;&nbsp;prop: Type<br/>} | fixmap or map 16/32 | sdf |
```
nerfZael, line 29: what is "sdf"?

arcbjorn: not sure, but looks like Semantic Definition File, probably related to GQL specific types

dOrgJelli: :rofl: "sdf" are 3 letters that are next to eachother on my keyboard, and I type them when I don't know what to put. This is something that shouldn't have been pushed :smile: effectively a bad "TODO"

packages/schema/bind/README.md
```diff
@@ -4,29 +4,29 @@ Generates Web3API schema bindings for supported guest languages.
...
+| GraphQL Schema Type                                  | [MessagePack Type](https://github.com/msgpack/msgpack/blob/master/spec.md#formats) | Description                      |
```
Niraj-Kamdar: Personally I don't like this formatted, since it is going to be rendered correctly anyway and editing with all these extra spaces would be harder.

Niraj-Kamdar (changes requested):

packages/schema/bind/README.md
nerfZael: I agree

dOrgJelli (approved):

PR #680, opened 2022-02-13
resolveUri - Minor Edits

dOrgJelli:
Don't expose the ApiCache in the client interface. Reasoning: https://github.com/polywrap/monorepo/pull/610#discussion_r805296344

### DIFF
--- a/packages/js/client/src/Web3ApiClient.ts
+++ b/packages/js/client/src/Web3ApiClient.ts
@@ -464,7 +464,7 @@ export class Web3ApiClient implements Client {
       this._toUri(uri),
       resolvers,
       client,
-      this._apiCache,
+      this._apiCache
     );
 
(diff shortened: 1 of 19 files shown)


dOrgJelli:
Also made some minor interface changes:
* `IUriToApiResolver` to `UriToApiResolver`
* Use `interface`s instead of `type`s
* Use the class constructor's name

Plus improved the sanity test-case, making `getUris()` only return the unique URIs in the array.

nerfZael:
All of the changes seem good to me, except for the cache.
I'm definitely hesitant to expose the cache through the client, but I think I'm even more opposed to having the resolvers access to the cache. I do however think we could merge this as is, for now, since it avoids the former and we can talk about the latter through a separate issue.

nerfZael (approved):

PR #682, opened 2022-02-14
649 | test (Schema Bind): plugin-ts & wasm-as

### LINKED ISSUE #649: Add more unit tests for schema-bind
Currently we just have simple sanity test in schema-bind which may miss a lot of things, we should have fine-grained tests for all the code binding to better cover this part of the code.

arcbjorn:


Niraj-Kamdar:
@dOrgJelli Can/Should we add a formatter with our own configs so that we don't get our git changes polluted by default formatter of the devs?

packages/schema/bind/src/__tests__/plugin-ts.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    it("nullable type", () => {
+      const type = "Type";
+      const formattedType = "Type | null";
+      const result = applyNullable(type, true);
+
+      expect(result).toBe(formattedType);
+    });
+  });
+
+  it("generateBinding", async () => {
```
Niraj-Kamdar, line 99: Won't this get covered by test-cases.spec.ts?

Niraj-Kamdar (commented):
Thanks for adding these tests 🙂

Unit tests looks good to me but we also need more functional tests apart from the default sanity one. It'd be nice to have it similar to what we have in tests for client-js. 

Ex: Binding tests for:
- JSON-type
- BigInt-type
- use capability (getImplementations)
- Array-type
- Imported type
- Interfaces, etc...

This will be under `/packages/test-cases/cases/bind`.

dOrgJelli:
Going to close these as they're a bit out of date, and we're adding more granular test cases within other PRs.

Issue #655, opened 2022-02-01
http plugin: wasm header deserializing fail 


ramilexe:
Error message:

```
Message: __w3_abort: Property must be of type 'map'. Found 'int'.
  Context: Deserializing imported query-type: post
    at array[7]: 
      at headers: Array<Types.HTTP_Header> | null >> type found, reading property
        at headers: unknown >> searching for property type
          at post: Types.HTTP_Response | null >> reading function output
File: ~lib/@web3api/wasm-as/msgpack/ReadDecoder.ts
Location: [269,5].

```

Http plugin expects header `value` field to be a string: https://github.com/polywrap/monorepo/blob/prealpha/packages/js/plugins/http/schema.graphql#L3

But axios could have a string array for `set-cookie` header: https://github.com/axios/axios/blob/4fbf61dc751ad69c7df9977d6edf7cf7ee87a5fb/index.d.ts#L6
https://github.com/axios/axios/blob/d99d5faac29899eba68ce671e6b3cbc9832e9ad8/test/specs/helpers/parseHeaders.spec.js#L31

Here are response headers I get:
```
[
  { key: 'date', value: 'Tue, 01 Feb 2022 23:23:47 GMT' },
  { key: 'content-type', value: 'application/json; charset=utf-8' },
  { key: 'content-length', value: '147' },
  { key: 'connection', value: 'close' },
  { key: 'cf-ray', value: '6d6eedf08d917b57-DME' },
  { key: 'cache-control', value: 'private, no-cache' },
  {
    key: 'set-cookie',
    value: [
      'AWSALB=SR...; Expires=Tue, 08 Feb 2022 23:23:47 GMT; Path=/',
      'AWSALBCORS=SR...; Expires=Tue, 08 Feb 2022 23:23:47 GMT; Path=/; SameSite=None; Secure',
      '__cflb=04...; SameSite=Lax; path=/; expires=Wed, 02-Feb-22 23:23:47 GMT; HttpOnly'
    ]
  },

```

PR #690, opened 2022-02-20
http plugin: response header as a list

ramilexe:
Axios could return value in response header as a list (array) instead of string. We need to handle it and convert it to string

Closes: #655


Niraj-Kamdar (approved):
LGTM! 

But we should research on creating more robust request-response type, since as we can see in the case of `set-cookie`, we need to parse the key, value manually on wrapper side which can be tedious task and definitely be automated.

We can look into http libraries of different strongly typed languages like Rust, go, c++ to know how they are handling it.

nerfZael (approved):

dOrgJelli (approved):

Issue #558, opened 2021-11-30
Base Schema Types: BigDecimal, BigNumber, Fraction


krisbitney:
dApps commonly require math with large numbers. We currently have the BigInt base schema type for math with large integers, but we do not have a base schema type to facilitate arbitrary-precision math with large decimal numbers.

BigNumber and BigDecimal classes are desirable when precision matters, such as for chained operations wherein rounding errors could accrue with each operation. It is sometimes possible to use a Fraction class as a workaround. Fractions may be especially useful when it is necessary to emulate on-chain arithmetic where decimal numbers are not supported.

A BigNumber base schema type should be able to handle arbitrary-precision math with large decimal or integer numbers, including support for algebraic operations like exponentiation and square root. Other desirable operations include those found in standard Math libraries, such as `ceil`, `floor` and `log` operations. The type should also be capable of returning formatted string representations with functions that support printing to a fixed number of decimal places or a fixed number of significant digits. Importantly, BigNumber operations should be based on performant algorithms.

It would be desirable (but not necessary) for BigNumber, Fraction, and BigInt classes to have similar syntax and to have means for typecasting between them. For example, a BigNumber class might have a `toBigInt()` function and BigInt might have `toBigNumber()`.

**Example Fraction implementations**
Uniswap SDK AssemblyScript port: https://github.com/polywrap/integrations/blob/main/uniswapv3/wrapper/src/utils/Fraction.ts
Python standard library: https://docs.python.org/3/library/fractions.html

**Example BigDecimal and BigNumber implementations in JavaScript**
bn.js: https://github.com/indutny/bn.js/
bignumber.js: https://github.com/MikeMcl/bignumber.js
big.js: https://github.com/MikeMcl/big.js/
decimal.js: https://github.com/MikeMcl/decimal.js/
js-big-decimal: https://github.com/royNiladri/js-big-decimal

**Example BigDecimal and BigNumber implementations in AssemblyScript**
as-big: https://github.com/ttulka/as-big
as-bigfloat: https://github.com/polywrap/as-bigfloat/blob/main/assembly/BigFloat.ts

###Tasks

- [x] BigNumber: See https://github.com/polywrap/monorepo/pull/697
- [ ] Fraction: In progress

krisbitney:
BigNumber is complete. See https://github.com/polywrap/monorepo/pull/697

Fraction type is in progress. It needs to be made generic and implemented as a generic schema type.

PR #697, opened 2022-02-26
Add BigNumber to base schema types

krisbitney:
This PR adds a base schema type, Bignumber.

BigNumber is a class for math with arbitrarily large decimal numbers. It is optimized with blockchain applications in mind. By default, it represents numbers up to 512 bits with full precision. Users can adjust that global setting or specify the desired precision for any specific operation by providing optional function arguments.

The class use efficient algorithms for standard arithmetic operations, exponentiation, and finding square roots. It also uses generics to allow most methods to be called with arguments of several possible types.

This PR also includes an update to BigInt with faster exponentiation, faster squaring, faster square roots, and use of generics throughout.

The two classes have similar interfaces so users don't have to learn anything new to get started.

A Fraction base schema type will be added in another PR.

See issue #558 

**LINKS**
[BigInt Repo](https://github.com/polywrap/as-bigint)
[BigNumber Repo](https://github.com/polywrap/as-bignumber)
[Fraction Repo](https://github.com/polywrap/as-fraction)


Niraj-Kamdar:
I think Fraction should be a generic type `Fraction<T>` where T can be `u8`, `i8` to `u64`, `i64` or BigInt.

dOrgJelli:
Hey @krisbitney, thoughts on @Niraj-Kamdar's comment about making Fraction generic? An example of a generic schema type can be found by looking at the "Map Type Support" PR that Niraj recently worked on here: https://github.com/polywrap/monorepo/pull/669

krisbitney:
> Hey @krisbitney, thoughts on @Niraj-Kamdar's comment about making Fraction generic? An example of a generic schema type can be found by looking at the "Map Type Support" PR that Niraj recently worked on here: #669

Yes, it's a good idea. I'll update it soon. Thanks!

dOrgJelli:
Okay, maybe we move Fraction<T> into its own PR, ,and just merge BigNumber here?

krisbitney:
> Okay, maybe we move Fraction into its own PR, ,and just merge BigNumber here?

That would be best. I'll split them into separate PRs soon.

Niraj-Kamdar (dismissed):

krisbitney:
> I think Fraction should be a generic type `Fraction<T>` where T can be `u8`, `i8` to `u64`, `i64` or BigInt.

Hey, so I'm thinking about implementation issues with this and I'm not sure how to handle return types.

Which type do you think should be returned from operations on Fractions of different generic types? For example, what should I get if I multiply a Fraction<i32> by a Fraction<BigInt>?. There are no union types in AS, so I have to choose between always returning the type of the left operand, the right operand, or a static type. The decision has to be made a priori, at compile time. Should return values always be the type of the left operand? Or maybe always return a value of type Fraction<BigInt>?

Do you think it matters if the Fraction operates with generic types internally? An easy solution would be to present the Fraction as generic in the schema, so that users can supply arguments of any accepted type, but the constructed Fractions would not be generic.

Niraj-Kamdar:
@krisbitney 

Ah, gotcha! For starters, We should only allow operations between the similar fraction generic type (Ex: Fraction<i32> can only do arithmetic operations with Fraction<i32> but can't do with Fraction<i8>), this will make it easy for us to implement and put the burden of type conversion on users wherever needed. 

We can in the future add the ability to auto-detect appropriate types for the operations using pre-compile transformation/optimization steps.

Here's a project which uses this kind of method to improve UX: https://github.com/JairusSW/as-json

krisbitney:
> Ah, gotcha! For starters, We should only allow operations between the similar fraction generic type (Ex: Fraction can only do arithmetic operations with Fraction but can't do with Fraction), this will make it easy for us to implement and put the burden of type conversion on users wherever needed.
> 
> We can in the future add the ability to auto-detect appropriate types for the operations using pre-compile transformation/optimization steps.
> 
> Here's a project which uses this kind of method to improve UX: https://github.com/JairusSW/as-json

Super cool. Thanks!

dOrgJelli (approved):

PR #699, opened 2022-03-01
Rust support

### LINKED ISSUE #217: Support Rust Based Web3APIs

namesty:
Closes #217

The current PR aims to provide support for Rust-based Polywrappers. Some points worth highlighting:

* The code generation for Rust transforms all function, method and variable names to snake_case; but serialized data like schema field names are not transformed.
* Just like the Assemblyscript's msgpack implementation, positive signed integers are not cast to unsigned integers when encoded. More info on this: https://ludocode.github.io/mpack/md_docs_protocol.html
* When running the Web3Api Client's tests, the large-types one makes the JS heap to run out of memory when testing from Jest (running manual scripts, it works fine). So I reduced the size of these arguments a bit and will look further into this.


packages/cli/src/lib/Compiler.ts
```diff
@@ -329,6 +364,16 @@ export class Compiler {
   }
 
   private _getGenerationDirectory(entryPoint: string): string {
+    const state = this._state;
+
+    if (
+      state &&
```
nerfZael: You can use `if(state?.compilerOverrides?.getGenerationDirectory)`

packages/cli/src/lib/project/Web3ApiProject.ts
```diff
@@ -239,6 +239,27 @@ export class Web3ApiProject extends Project {
         ),
       };
 
+      // If there exists a query & mutation module, add the query
```
nerfZael: "//TODO:" is a good convention to follow for comments like these to make them more discoverable.
Additionally we could create a github issue and link it here

packages/wasm/as/assembly/msgpack/ReadDecoder.ts
```diff
@@ -131,6 +131,8 @@ export class ReadDecoder extends Read {
     const prefix = this._view.getUint8();
     if (isFloat64(prefix)) {
       return <f64>this._view.getFloat64();
+    } else if (isFloat32(prefix)) {
+      return <f64>this._view.getFloat32();
```
nerfZael, line 136: Shouldn't this be `<f32>this._view.getFloat32();` ?

packages/wasm/rs/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "polywrap-wasm-rs"
```
nerfZael: I don't think crates usually have an "rs" suffix

packages/wasm/rs/src/invoke.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub method: String,
+    pub args: Vec<u8>,
+}
+
+/// Helper for fetching invoke args
+pub fn w3_invoke_args(method_size: u32, args_size: u32) -> InvokeArgs {
+    let method_size_ptr = alloc(method_size as usize);
+    let args_size_ptr = alloc(args_size as usize);
+
+    unsafe { __w3_invoke_args(method_size_ptr as u32, args_size_ptr as u32) };
```
nerfZael, line 31: Since __w3_invoke_args calls into the client and it's in an unsafe block, I wonder if there are any security concerns here. But, I don't know enough to say

packages/wasm/rs/src/invoke.rs
```diff
@@ -0,0 +1,59 @@
...
+        unsafe { Vec::from_raw_parts(args_size_ptr, args_size as usize, args_size as usize) };
+
+    InvokeArgs { method, args }
+}
+
+/// Helper for handling `_w3_invoke`
+pub fn w3_invoke(args: InvokeArgs, opt_invoke_func: Option<InvokeFunction>) -> bool {
+    match opt_invoke_func {
+        Some(func) => {
+            let result = func(args.args.as_slice());
```
nerfZael, line 46: We should probably use `&args.args[]` here instead of `args.args.as_slice()`. It seems like `as_slice()` is on it's way to being deprecated. https://github.com/rust-lang/rfcs/blob/master/text/0198-slice-notation.md

packages/wasm/rs/src/get_implementations.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    // deserialize the `msgpack` buffer,
+    // which contains a `Vec<String>`
+    let mut decoder = ReadDecoder::new(&result_buffer, Context::new());
+    decoder.context().push(
+        "__w3_getImplementations_result",
+        "Vec<String>",
+        "__w3_getImplementations successful",
+    );
+    decoder.read_array(|reader| reader.read_string()).unwrap()
```
nerfZael, line 37: Should we be unwrapping/not handling the error here?

packages/wasm/rs/src/invoke.rs
```diff
@@ -0,0 +1,59 @@
...
+
+pub struct InvokeArgs {
+    pub method: String,
+    pub args: Vec<u8>,
+}
+
+/// Helper for fetching invoke args
+pub fn w3_invoke_args(method_size: u32, args_size: u32) -> InvokeArgs {
+    let method_size_ptr = alloc(method_size as usize);
+    let args_size_ptr = alloc(args_size as usize);
```
nerfZael, line 29: I don't think we're ever calling dealloc for this

packages/wasm/rs/src/msgpack/context.rs
```diff
@@ -0,0 +1,105 @@
+//! Context stores debug information in a stack, and
```
nerfZael, line 1: If this is just for debugging, why is it a part of msgpack?

packages/wasm/rs/src/get_implementations.rs
```diff
@@ -0,0 +1,38 @@
...
+    #[link_name = "__w3_getImplementations_result"]
+    pub fn __w3_getImplementations_result(ptr: u32);
+}
+
+pub fn w3_get_implementations(uri: &str) -> Vec<String> {
+    let success =
+        unsafe { __w3_getImplementations(uri.as_bytes().as_ptr() as u32, uri.len() as u32) };
+
+    if !success {
+        return vec![];
```
nerfZael, line 20: Should we panic here instead?

packages/wasm/rs/tests/invoke_spec.rs
```diff
@@ -0,0 +1,40 @@
+use polywrap_wasm_rs::malloc::alloc;
```
nerfZael: We're missing tests for env, get_implementations and subinvoke

packages/js/asyncify/src/AsyncWasmInstance.ts
```diff
@@ -107,9 +107,12 @@ export class AsyncWasmInstance {
     instance._wrappedImports = instance._wrapImports(config.imports);
 
     // Create Wasm module instance
-    instance._instance = (
-      await WebAssembly.instantiate(config.module, instance._wrappedImports)
-    ).instance;
+    const module = await WebAssembly.compile(config.module);
```
nerfZael: So we compile now, but didn't need to before or was it an oversight?

packages/test-cases/cases/apis/wasm-rs/asyncify/query/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+pub fn get_data(input: query::InputGetData) -> u32 {
+    match ethereum_query::EthereumQuery::call_contract_view(
+        &ethereum_query::InputCallContractView {
+            address: input.address,
+            method: "function get() view returns (uint256)".to_string(),
+            args: None,
+            connection: input.connection,
+        },
+    ) {
+        Ok(v) => v.parse::<u32>().unwrap(),
```
nerfZael, line 15: Shouldn't this be bigInt? The return is unsigned 256 bit integer and we're using a 32 bit one

packages/test-cases/cases/apis/wasm-rs/asyncify/mutation/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+    }) {
+        Ok(_v) => true,
+        Err(_e) => false,
+    }
+}
+
+pub fn subsequent_invokes(input: mutation::InputSubsequentInvokes) -> Vec<String> {
+    let mut result: Vec<String> = vec![];
+
+    for i in 0..input.number_of_times as usize {
```
nerfZael: Why are we casting to usize?

packages/test-cases/cases/apis/wasm-rs/implementations/test-use-getImpl/package.json
```diff
@@ -0,0 +1,8 @@
+{
```
nerfZael: I think this file is here by mistake

packages/test-cases/cases/apis/wasm-rs/json-type/query/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+    let json_a_str = JSON::to_string(&input.object.json_a).unwrap();
+    let json_b_str = JSON::to_string(&input.object.json_b).unwrap();
+    
+    result.push_str(&json_a_str);
+    result.push_str(&json_b_str);
+
+    result
+}
+
+pub fn method_j_s_o_n(input: InputMethodJSON) -> JSON::Value {
```
nerfZael: Why is the function named like this? _j_s_o_n?

packages/test-cases/cases/apis/wasm-rs/json-type/query/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+pub fn parse(input: InputParse) -> JSON::Value {
+    JSON::from_str(&input.value).unwrap()
+}
+
+pub fn stringify(input: InputStringify) -> String {
+    let result_vec: Vec<String> = input.values.iter()
+    .map(|value| JSON::to_string(&value).unwrap())
+    .collect();
+
+    polywrap_wasm_rs::w3_debug_log(&format!("{:?}", result_vec));
```
nerfZael: Should we add this to the "use" statement along with JSON?
Then we can just do `w3_debug_log(...);`

packages/test-cases/cases/apis/wasm-rs/number-types/query/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+
+pub fn u16_method(input: InputU16Method) -> u16 {
+    input.first + input.second
+}
+
+pub fn i32_method(input: InputI32Method) -> i32 {
+    input.first + input.second
+}
+
+pub fn u32_method(input: InputU32Method) -> u32 {
```
nerfZael, line 24: Some tests I noticed are missing are floats, 64 bit uints and ints, String and Bytes

packages/test-cases/cases/apis/wasm-rs/object-types/query/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+            nested: Nested {
+                prop: input.arg.nested.prop,
+            },
+        }),
+    ]
+}
+
+pub fn method5(input: InputMethod5) -> Output {
+    Output {
+        // prop: String::from_utf8(input.arg.prop).unwrap(),
```
nerfZael: Why the comment?

nerfZael (changes requested):

packages/wasm/as/assembly/msgpack/ReadDecoder.ts
dOrgJelli: No, because `readFloat64` returns an `f64`.

packages/wasm/rs/tests/test_encode.rs
```diff
@@ -0,0 +1,244 @@
...
+fn test_write_i8() {
+    let mut writer = WriteEncoder::new(&[], Context::new());
+    writer.write_i8(&(-1)).unwrap();
+    assert_eq!([255], writer.get_buffer().as_slice());
+}
+
+#[test]
+fn test_write_i16() {
+    let mut writer = WriteEncoder::new(&[], Context::new());
+    writer.write_i16(&(-123)).unwrap();
```
ramilexe: `-123` is not `i16`. It is still `i8`. You need to use something less than -128

packages/wasm/rs/tests/test_encode.rs
```diff
@@ -0,0 +1,244 @@
...
+
+#[test]
+fn test_write_u32() {
+    let mut writer = WriteEncoder::new(&[], Context::new());
+    writer.write_u32(&4294967295).unwrap();
+    assert_eq!([206, 255, 255, 255, 255], writer.get_buffer().as_slice());
+}
+
+#[test]
+fn test_write_i8() {
```
ramilexe, line 123: I think it makes sense to add additional test cases for positive i8

ramilexe (changes requested):

packages/test-cases/cases/apis/wasm-rs/number-types/query/src/lib.rs
dOrgJelli: We don't currently support floats & 64 bit integers. Also String + Bytes cannot have under/overflows in the same way integers can.

packages/test-cases/cases/apis/wasm-rs/asyncify/query/src/lib.rs
dOrgJelli: technically yes it should, but this is just for testing so it doesn't matter much, we can assume the value in the SimpleStorage contract is `< u32`.

packages/test-cases/cases/apis/wasm-rs/asyncify/mutation/src/lib.rs
dOrgJelli: Nice catch :ok_hand:

packages/js/asyncify/src/AsyncWasmInstance.ts
dOrgJelli: Ahhh thank you for catching this, turns out this was added for debugging, and is no longer needed. Will remove!

packages/wasm/rs/tests/test_encode.rs
dOrgJelli: Agreed, we'll be sure to add more robust msgpack testing in the future, and make it usable for all wasm runtimes

packages/wasm/rs/tests/test_encode.rs
dOrgJelli: Great catch, thank you!

dOrgJelli (approved):

packages/wasm/rs/src/get_implementations.rs
dOrgJelli: Current functionality in AssemblyScript is to return an empty array, so this is "correct" based on that. I'm undecided on what we should do here... open to suggestions & thoughts.

packages/wasm/rs/src/msgpack/context.rs
dOrgJelli: Fixed

packages/wasm/rs/src/invoke.rs
dOrgJelli: The ownership of these member buffers is transferred to `String` and `Vec` when calling the `from_raw_parts` functions: https://doc.rust-lang.org/std/string/struct.String.html#method.from_raw_parts

packages/wasm/rs/src/get_implementations.rs
dOrgJelli: If an error occurs, then we want this to panic and fail, which will happen. The error information form the deserialization error will be present in the error message.

packages/wasm/rs/src/invoke.rs
dOrgJelli: hmmm, what types of attacks are you imagining?

Unsafe in this case just refers to this code being unchecked by the Rust compiler... so if you were worried about decreased security due to this code here, it'd mean that the attack would have to take place when building the wrapper.

In this scenario, the developer's machine is already compromised, so the resulting wasm binary could be rewritten and compromised as well. Nothing this unsafe block can do to protect against that (if I'm understanding correctly).

packages/wasm/rs/Cargo.toml
dOrgJelli: It's just our convention to ensure that viewers of a package name know what language it is intended for. See here for an example: https://www.npmjs.com/package/@web3api/wasm-as

PR #725, opened 2022-03-09
565 change query recipe structure

pwvpwvpwvpwv:
Adds ability to specify multiple query recipes in the same file, and choose which to run from the CLI. Can be written in either JSON or YAML.


dOrgJelli:
Some notes from the CLI WG meeting on 2022-03-21:
```
- configuration for running (a)sync (recipes/menus/etc)
- - --runInBand like Jest, sane default is async
- - --only or in receipe config, just like jest
- more fleshed out example query receipe file
- possibly rename "cookRecipes" to something that's more inline w/ current interface (easier to discover)
```

dOrgJelli:
Closing this as @Niraj-Kamdar has taken this feature over, and created a new version viewable here https://github.com/polywrap/monorepo/pull/878 and here https://github.com/polywrap/monorepo/pull/903

PR #745, opened 2022-03-14
663 | feat (Client Uri resolution): improve ipfs ens error handling

### LINKED ISSUE #663: Surface Resolver Extension Error Messages
Currently the error handling in the IPFS & ENS plugins is very poor when doing URI resolution.

Some conversation around this:

"worst part right now is if ens or ipfs resolver has an exception it will catch it and throw a generic "cant resolve" error" - @nerfZael 

"I think it should be more specific about the reason why it couldn't resolve the wrapper (Ex: NetworkError, TimeoutError, FileNotFoundError, etc)" - @Niraj-Kamdar 

"I'd even go a bit further and show the underlying error. We already catch it but then ignore/obfuscate it. Recently, while trying to figure out what signer to use for the ethereum plugin, I had to console log the error in the ens plugin just so I could see the actual error instead of the "Unable to resolve uri" error message" - @nerfZael

arcbjorn:


dOrgJelli:
Closing this as this has been addressed here: https://github.com/polywrap/toolchain/pull/1169

PR #747, opened 2022-03-14
change chainId from Int/UInt to BigInt type Ethereum plugin

krisbitney:


packages/js/plugins/ethereum/src/index.ts
```diff
@@ -154,7 +154,7 @@ export class EthereumPlugin extends Plugin {
     const network = await provider.getNetwork();
     return {
       name: network.name,
-      chainId: network.chainId,
+      chainId: network.chainId.toString(),
```
Niraj-Kamdar, line 157: Aren't we serializing this automatically?

krisbitney: That would be true if this were a wrapper, but in plugins the `BigInt` type is an alias for `string`. Here I'm changing the type from `number` to `string`.

dOrgJelli (dismissed):
Great catch!

PR #769, opened 2022-03-22
feat: add support for getImplementations capability in plugin

Niraj-Kamdar:


nerfZael:
What's the context of this issue? Is this needed to make the plugin dev exp more similar to the wrapper one?

EDIT: Yes, I think that's the case. This is the wrapper exp PR: https://github.com/polywrap/monorepo/pull/516

packages/test-cases/cases/bind/sanity/output/plugin-ts/types.ts
```diff
@@ -149,32 +149,40 @@ interface TestImport_Query_Input_anotherMethod extends Record<string, unknown> {
 }
 
 /* URI: "testimport.uri.eth" */
-export const TestImport_Query = {
-  importedMethod: async (
+export class TestImport_Query {
+  public static interfaceUri: string = "testimport.uri.eth";
+  public uri: string;
+
+  constructor(uri: string) {
```
nerfZael: You can use: 
```
constructor(public uri: string) {
}
```

Niraj-Kamdar:
> What's the context of this issue? Is this needed to make the plugin dev exp more similar to the wrapper one?

@nerfZael  Yes, that's the exact reason!

nerfZael (approved):

dOrgJelli (approved):

PR #776, opened 2022-03-23
add getBalance function to Ethereum plugin

krisbitney:
The Ethereum plugin has `getSignerBalance`, but does not support getting the Ether balance of arbitrary accounts. This PR adds `getBalance`, which takes an address as an argument and returns the Ether balance of the address. This is useful for defiwrapper.


Niraj-Kamdar (approved):
LGTM

dOrgJelli (approved):

PR #777, opened 2022-03-23
Web3ApiClient - Preload API resolvers

### LINKED ISSUE #715: URI Resolver Wrapper Infinite Loop
From my current understanding of the client URI resolution process, there is an infinite loop if using a URI Resolve Wrapper

Lets say you have an "ens-resolver.eth" URI resolver wrapper
Invoking a URI "test.eth" would work like this:

client.invoke("test.eth")
  resolveUri
    client.getImplementations("uri-resolver") -> returns ["ens-resolver.eth"]
    forEach URI resolver wrapper
      client.invoke("ens-resolver.eth")
        resolveUri
          client.getImplementations("uri-resolver") -> returns ["ens-resolver.eth"]
          forEach URI resolver wrapper
            client.invoke("ens-resolver.eth")
              resolveUri
                client.getImplementations("uri-resolver") -> returns ["ens-resolver.eth"]
                
This is still just a theory and needs to be tested. If there is indeed an infinite loop, we can address it as a separate issue.

nerfZael:
Closes: #715 
Unblocks: https://github.com/polywrap/demos/issues/23


packages/js/core/src/uri-resolution/resolvers/api/ApiAggregatorResolver.ts
```diff
@@ -10,16 +10,24 @@ import { CreateApiFunc } from "./types/CreateApiFunc";
...
+  private hasLoadedAllResolvers: boolean;
+
   constructor(
     private readonly createApi: CreateApiFunc,
-    private deserializeOptions?: DeserializeManifestOptions
-  ) {}
+    private deserializeOptions?: DeserializeManifestOptions,
+    disablePreloadResolvers?: boolean
+  ) {
+    if (disablePreloadResolvers) {
```
nerfZael: This is a sort of hacky solution to the problem of using the aggregator in core-js tests

dOrgJelli: hmmm, should we use a `TEST` environment variable maybe?

dOrgJelli (changes requested):
This all looks really good!

I spent some time restructuring the terminology to something that I think makes more sense to new users. Would love to discuss here: https://github.com/polywrap/monorepo/pull/806

If you're okay with these proposed changes, I'm good with this PR :D

Everything looks great :100:

packages/js/core/src/uri-resolution/resolvers/api/ApiAggregatorResolver.ts
nerfZael: Maybe, but I think this is also has usecases elsewhere, eg a performance increase if you know all of your resolvers are local

dOrgJelli (approved):

PR #790, opened 2022-04-04
chore: client improved env error handling + test cases

### LINKED ISSUE #783: Add ensureEnv function which makes sure env is defined or throw an error

Niraj-Kamdar:
-  Closes: #783

### Changelog
- Adds convenience method `requireEnv` so that users don't need to type-cast whenever they want to use Env. 
- It also raises an error if env is undefined and hasn't been caught in the deserialization step which should never happen in ideal case though
- Refactored env related test cases out from the `Web3ApiClient.spec.ts` file


packages/test-cases/cases/bind/sanity/output/wasm-as/query/env.ts
```diff
@@ -3,3 +3,8 @@ import {
 } from "./QueryEnv";
 
 export let env: QueryEnv | null = null;
+
+export function ensureEnv(): QueryEnv {
```
nerfZael: ensureEnv is a bit unclear
maybe ensureEnvDefined ?

packages/js/client/src/__tests__/Web3ApiClient.spec.ts
```diff
@@ -2458,6 +2458,7 @@ scalar Int32
 scalar Bytes
 scalar BigInt
 scalar JSON
```
nerfZael: This file looks like it was left over from the Map PR by accident, is that correct?

nerfZael (changes requested):

nerfZael:
Shouldn't the new ensureEnv function be called somewhere? or is that up to the user?

packages/js/client/src/__tests__/Web3ApiClient.spec.ts
Niraj-Kamdar: I believe it's because of the way we build plugins. We use the latest released version of the toolchain to build the plugin that's why some changes are only reflected after the release.

packages/test-cases/cases/bind/sanity/output/wasm-as/query/env.ts
Niraj-Kamdar: I find it too verbose though! open for other suggestions :)

nerfZael: requireEnv? That's a common pattern

packages/js/client/src/__tests__/Web3ApiClient.spec.ts
nerfZael: Ah, ok

Niraj-Kamdar:
> Shouldn't the new ensureEnv function be called somewhere? or is that up to the user?

It's up to user whether they'd like to use it or not

packages/test-cases/cases/bind/sanity/output/wasm-as/query/env.ts
Niraj-Kamdar: sounds good

packages/js/client/src/__tests__/env.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+  initTestEnvironment,
+  stopTestEnvironment,
+} from "@web3api/test-env-js";
+import { createWeb3ApiClient, Web3ApiClientConfig } from "..";
+import { Client, Plugin, PluginModules } from "@web3api/core-js";
+import { GetPathToTestApis } from "@web3api/test-cases";
+
+jest.setTimeout(200000);
+
+describe("Web3ApiClient", () => {
```
dOrgJelli: This name should be changed to "env" or something else more descriptive.

packages/js/client/src/__tests__/env.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+    return {
+      factory: () => new MockEnvPlugin(),
+      manifest: {
+        schema: ``,
+        implements: [],
+      },
+    };
+  };
+
+  describe("env types", () => {
```
dOrgJelli: The only tests we're running here are "happy paths", and we're not testing the error cases that this PR tries to address.

I would request that we add tests for the following cases:
- No env provided, when one is required
- Incorrect env provided
- Env provided, when wrapper does not ask for one (shouldn't throw an error)

packages/schema/bind/src/bindings/assemblyscript/wasm-as/templates/env-ts.mustache
```diff
@@ -4,11 +4,21 @@ import {
 } from "./QueryEnv";
 
 export let env: QueryEnv | null = null;
+
+export function requireEnv(): QueryEnv {
```
dOrgJelli, line 8: This feels like a concern that the wrapper developer shouldn't have, and instead it should expect the client to ensure an environment is provided if one is specified in the schema.

This type of checking is already being done here in the `WasmWeb3Api` class: https://github.com/polywrap/monorepo/blob/daa694fd5d7e77c907307c38a40ca388ab23b5cc/packages/js/client/src/wasm/WasmWeb3Api.ts#L336-L354

I think instead of adding error handling here, we should instead improve the code posted above to handle errors in a more granular way.

NOTE: The client makes the assumption that the `_w3_load_env` function will be present if the wrapper requires an environment. This is the only logical way to detect this requirement IMO.

dOrgJelli (changes requested):
In general, I think we should re-frame this PR to:
- Improve environment related test-cases in the client, testing for all possible error modes.
- Improve environment related error handling in the client (`WasmWeb3Api`, and `PluginWeb3Api`)

If you agree, let's change the PR title to something like "chore: client improved env error handling + test cases"

packages/schema/bind/src/bindings/assemblyscript/wasm-as/templates/env-ts.mustache
Niraj-Kamdar: Yep, I agree, and when we load the env, It will take care of the type on `serialize <-> deserialize` steps. This is just a convenient method for the wrapper developer to get the env in concrete type (i.e: QueryEnv or MutationEnv), not the  (QueryEnv | null) because handling this on the wrapper side just makes the code more verbose, not a big deal but clutters the code.

Niraj-Kamdar: > I think instead of adding error handling here, we should instead improve the code posted above to handle errors in a more granular way.

Can you elaborate on this?

packages/js/client/src/__tests__/env.spec.ts
Niraj-Kamdar: I agree

dOrgJelli (approved):

PR #794, opened 2022-04-04
feat: add e2e tests in template

### LINKED ISSUE #792: Wrapper project template should come with e2e tests

Niraj-Kamdar:
- Closes: #792


packages/templates/api/assemblyscript/package.json
```diff
@@ -15,11 +15,17 @@
     "test": "npx w3 query ./recipes/e2e.json"
   },
   "devDependencies": {
+    "@types/jest": "^27.0.3",
```
dOrgJelli: Dependencies should be locked. Please remove the `^` version resolution modifier.

packages/templates/api/assemblyscript/src/__tests__/e2e/integration.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+import { getPlugins } from "../utils";
+
+jest.setTimeout(500000);
+
+describe("SimpleStorage", () => {
+  const CONNECTION = { node: "http://localhost:8545" };
+
+  let client: Web3ApiClient;
+  let ensUri: string;
+  let app: any;
```
dOrgJelli: Let's avoid the use of any types. We can simply import the types at the top of the file like so `import * as App from "../types/w3"` which would retain type information.

dOrgJelli: Ahh, apologies, I now see why you chose to do this. It's because you're generating the types in-line in your beforeAll step.

IMO having the types be cast as `any` defeats the purpose of the types, so we should take this out of the test's run-time and instead do this as a pre-processing step, allowing us to import types normally.

packages/templates/api/assemblyscript/src/__tests__/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+            },
+            MAINNET: {
+              provider: "http://localhost:8546",
+            },
+          },
+          defaultNetwork: "testnet",
+        }),
+      },
+    ],
+    interfaces: [
```
dOrgJelli: These are added by default, so no need to set this up yourself.

dOrgJelli (changes requested):

packages/templates/api/assemblyscript/yarn.lock
```diff
@@ -0,0 +1,5888 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
dOrgJelli: yarn.lock files shouldn't be versioned for packages, and also shouldn't be present in template projects as a whole. They should be left to be generated client-side when users create the template project.

packages/templates/api/assemblyscript/package.json
```diff
@@ -15,11 +15,17 @@
     "test": "npx w3 query ./recipes/e2e.json"
```
dOrgJelli: The "test" script needs to be updated to include the e2e jest tests.

dOrgJelli (changes requested):

dOrgJelli (approved):

PR #797, opened 2022-04-04
feat:Implement manifest schemata and add version checking to client

pwvpwvpwvpwv:
Added versioning information to the main and build manifests. Added some pre-build scripts to generate compile-time constants for version numbers, for use across the CLI and client. the client will now reject wrappers which do not target the runtime version provided by the client (this is the default, but the restriction can be lowered to simply warn to `console.error` by changing the client config)


dOrgJelli (commented):
Closing this PR as we've decided to refactor the build artifacts, and support a protocol version in another way: https://github.com/polywrap/monorepo/issues/962

PR #798, opened 2022-04-04
Changed chainId type from Bigint to UInt32

namesty:
This PR changes the Ethereum Plugin's `chainId` property from `BigInt` to `UInt32`

The ENS wrapper, used by the Hub, was deployed with version `0.0.1-prealpha.48` of the Ethereum Plugin, where `chainId` is an `UInt32`. But the hub uses the latest `0.0.1-prealpha.69`, where `chainId` is a `BigInt`, which serializes as string, which is indeed a breaking change.

Chain ID should indeed be a number and not a bigint, ethersjs docs validate this: https://docs.ethers.io/v5/api/providers/types/#providers-TransactionRequest.

### DIFF
--- a/packages/cli/src/__tests__/plugin/expected-build/schema.graphql
+++ b/packages/cli/src/__tests__/plugin/expected-build/schema.graphql
@@ -293,7 +293,7 @@ type Ethereum_TxRequest @imported(
   gasPrice: BigInt
   data: String
   value: BigInt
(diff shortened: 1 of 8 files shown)


krisbitney:
The issue with using u32 is that some chain ids won't fit. I think u64 would work, but we don't yet support it. Ethers uses JavaScript numbers, which can be much large than u32.

PR #807, opened 2022-04-11
chore: make docker filelock project specific

Niraj-Kamdar:
This PR will make the build time docker file lock project-specific instead of system-wide which will allow users to leverage concurrency.


packages/cli/src/commands/build.ts
```diff
@@ -265,7 +268,7 @@ export default {
             (key.name == "c" && key.ctrl)
           ) {
             await watcher.stop();
-            await dockerLock.release();
+            // await dockerLock.release();
```
dOrgJelli: This is needed. If the process is `kill()`ed, then the file lock will still exist, causing subsequent calls to halt indefinitely.

packages/cli/src/lib/project/Web3ApiProject.ts
```diff
@@ -52,7 +52,7 @@ export class Web3ApiProject extends Project<Web3ApiManifest> {
     this._buildManifest = undefined;
     this._metaManifest = undefined;
     this._defaultBuildManifestCached = false;
-    this.resetCache();
+    this.removeCacheDir(cacheLayout.buildEnvDir);
```
dOrgJelli: Nice catch :)

packages/cli/src/lib/system/file-lock.ts
```diff
@@ -1,4 +1,17 @@
 import * as fs from "fs";
+import path from "path";
+
+export function createDirs(dirPath: string): void {
```
dOrgJelli: This can be done via `fs.mkdirSync(dir, { recursive: true });`

dOrgJelli (changes requested):

packages/cli/src/commands/build.ts
Niraj-Kamdar: That's just debugging leftover

Niraj-Kamdar: Thanks for catching it

packages/cli/src/lib/system/file-lock.ts
Niraj-Kamdar: Nice! didn't know about that 😛

krisbitney (dismissed):
LGTM

PR #808, opened 2022-04-12
feat(cli): add support for buildkit options

Niraj-Kamdar:
- Added support for some docker buildkit specific build options
  1. local docker image layer cache
  2. custom image output location
  3. allow removing the image after build
  4. allow removing the builder instance after build


packages/cli/src/__tests__/e2e/build.spec.ts
```diff
@@ -256,4 +258,27 @@ ${HELP}`);
...
+
+    expect(code).toEqual(0);
+    expect(sanitizedOutput).toContain(
+      "Artifacts written to ./build from the image `polywrap-build-env-"
+    );
+    expect(sanitizedOutput).toContain(
+      "Manifest written to ./build/web3api.json"
+    );
+    expect(sanitizedOutput).toContain(manifestPath);
+    expect(fs.existsSync(path.join(projectRoot, ".w3", "web3api", "build", "cache"))).toBeTruthy();
```
dOrgJelli: Are we able to test for the presence of specific contents within the cache directory? This will help us ensure it is working correctly, and not just creating the directory.

packages/cli/src/__tests__/project/web3api.buildx.yaml
```diff
@@ -0,0 +1,16 @@
+format: 0.0.1-prealpha.3
+docker:
+  dockerfile: ./Dockerfile
+  buildkit:
+    enabled: true
```
dOrgJelli: Isn't this redundant? If `buildkit:` is present, you're already saying "I want buildkit enabled", correct?

packages/cli/src/__tests__/project/web3api.buildx.yaml
```diff
@@ -0,0 +1,16 @@
...
+    cacheDir: DEFAULT
```
dOrgJelli: I think this should be `cache: string | boolean`, that way users can simply say `cache: true` which will use the default path, or say `cache: ./path/` to specify a specific path.

packages/cli/src/lib/Compiler.ts
```diff
@@ -393,7 +417,10 @@ export class Compiler {
       project.getManifestDir(),
       imageName,
       dockerfile,
-      project.quiet
+      cacheDir,
+      buildxOutput,
+      project.quiet,
```
dOrgJelli: I think having `quiet: boolean` be the last argument of the function makes sense for both `createBuildImage` and `copyArtifactsFromBuildImage`

packages/cli/src/lib/system/docker.ts
```diff
@@ -76,6 +87,15 @@ export async function copyArtifactsFromBuildImage(
     }
 
     await runCommand(`docker rm -f root-${imageName}`, quiet);
+
+    if (useBuildkit) {
+      if (removeBuilder) {
+        await runCommand(`docker buildx rm ${imageName}`, quiet);
+      }
+      if (removeImage) {
+        await runCommand(`docker rmi ${imageName}`, quiet);
```
dOrgJelli: This does not seem to require buildx, should it be inside the `buildkit` options in the manifest?

packages/cli/src/lib/system/docker.ts
```diff
@@ -17,14 +17,25 @@ export function getDockerFileLock(): FileLock {
   return new FileLock(__dirname + "/DOCKER_LOCK", print.error);
 }
 
+export async function isDockerBuildxInstalled(): Promise<boolean> {
+  const { stdout: version } = await runCommand("docker buildx version");
```
dOrgJelli, line 21: I tried this on my local machine, and it looks like it doesn't work. What are the setup steps to add buildx support to the docker CLI?

If it requires a certain version of docker to be installed, it'd be nice to add  some nice error messaging for the user so they know that their buildkit options are not being used due to their docker version being too old.

dOrgJelli (changes requested):

packages/cli/src/__tests__/project/web3api.buildx.yaml
Niraj-Kamdar: What if the user wants to just enable buildkit without any custom options? (Ex: cache, output)

Niraj-Kamdar: Nvm We can do similar to what you suggested below, `buildkit: Boolean | BuildkitOptions`

packages/cli/src/lib/system/docker.ts
Niraj-Kamdar: Nice catch!

packages/cli/src/lib/system/docker.ts
Niraj-Kamdar: unfortunately, it's not that easy and depends on the OS and the way you installed it. It comes included with the docker desktop but if you have installed docker using other means, you need to install buildkit separately.

here's the guide:
https://docs.docker.com/buildx/working-with-buildx/

dOrgJelli (approved):

PR #814, opened 2022-04-14
CLI deployment pipeline

### LINKED ISSUE #670: CLI: Integrate Wrappers CLI Deployment (IPFS & ENS)
@nerfZael has done a lot of great work creating the `wrappers` CLI which provides:
- IPFS publishing via wrappers.io gateway
- ENS publishing (mainnet & rinkeby)

We should integrate this functionality into the main `w3` CLI.

https://github.com/nerfZael/wrappers

namesty:
Closes #670.

This PR aims to create a modular deployment pipeline for Polywrap's CLI. It introduces a new `deploy` CLI command and a `DeployManifest`.

Users declare deployment "stages" or steps, where each one performs deployment actions. Each stage takes a `Uri` argument outputs a `Uri`. Additionally, stages can be chained, so that the output of a stage can automatically be used as the input of another.

This aims to provide a modular, extensible, way of performing complex deployment/publishing/persistance actions that may depend on each other. An example use case would be:

1. Deploy a wrapper's build directory to IPFS
2. Pin the IPFS hash
3. Publish the IPFS hash to an ENS domain
4. Publish the IPFS hash to a Tezos domain

Each stage has an associated (declared) external package (wrappers in the future) that implements the `Deployer` interface:

```ts
interface Deployer {
  execute(uri: Uri, config?: unknown): Promise<Uri>;
}
```

The CLI determines the dependency chains, executes each stage in order of dependency and properly passes the outputs/inputs from one stage to another.

Sometimes, stages (and its packages) need additional arguments in order to execute the deployment actions. For example, an ENS deployment stage may need an ethereum provider URL. These additional arguments are declared as part of the `config` object of each stage.

Each package may provide a `web3api.deploy.ext.json` manifest extension file, with the necessary rules to validate the `config` properties it needs. The CLI extracts these extensions, links them to the appropriate stage in the "base" manifest and performs manifest validation. If the extension does not exist, the `config` property of that stage is not validated.


packages/js/core/src/manifest/formats/web3api.deploy/deserialize.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { compare } from "semver";
+import YAML from "js-yaml";
+import { Tracer } from "@web3api/tracing-js";
+
+export const deserializeDeployManifest = Tracer.traceFunc(
+  "core: deserializeDeployManifest",
+  (manifest: string, options?: DeserializeManifestOptions): DeployManifest => {
+    let anyDeployManifest: AnyDeployManifest | undefined;
+    try {
```
Niraj-Kamdar, line 25: Why don't we load json with yaml loader? Since YAML is superset of JSON, it should be possible.

packages/js/client/src/__tests__/Web3ApiClient.spec.ts
```diff
@@ -948,7 +948,7 @@ describe("Web3ApiClient", () => {
     ).toBe(true);
   });
 
-  it("simple-storage", async () => {
+  it.only("simple-storage", async () => {
```
Niraj-Kamdar: Is it debug leftover?

packages/cli/src/lib/deployers/ipfs/index.ts
```diff
@@ -0,0 +1,81 @@
...
+        if (formDataEntry.opts.filePath && formDataEntry.opts.filePath != "") {
+          options.filepath = formDataEntry.opts.filePath;
+        }
+      }
+      const elementData =
+        formDataEntry.data == null ? Buffer.alloc(0) : formDataEntry.data;
+      fd.append(formDataEntry.key, elementData, options);
+    });
+
+    const resp = await axios.post(`${ipfsUrl}/api/v0/add`, fd, {
```
Niraj-Kamdar: Will it also pin or we need separate deployer for pinning?

Niraj-Kamdar (dismissed):
LGTM!

dOrgJelli (dismissed):

PR #820, opened 2022-04-22
docgen, a CLI command to auto-generate documentation #428

krisbitney:
This PR replaces https://github.com/polywrap/monorepo/pull/428.

Added CLI command "docgen" to automatically generate API documentation from graphql schemas. The command supports four types of output:
- GraphQL Schema
- Docusaurus React App
- Docusaurus markdown
- JSDoc template

docgen works with Wrapper, Plugin, and App projects. This means app developers can generate documentation for wrappers and plugins when they create an app manifest and list imports in a schema.


krisbitney:
The `react` and `schema` sub-commands can be removed in the future as part of issues https://github.com/polywrap/monorepo/issues/834 and https://github.com/polywrap/monorepo/issues/833

Niraj-Kamdar (commented):
I believe that docs should just have docs regarding the wrapper and docs regarding the polywrap should only be linked. If we really want to supply polywrap docs then there should be an explicit option for the user who wants to embed it.

dOrgJelli (approved):
:fire: :fire: :fire: :fire: @krisbitney

PR #827, opened 2022-04-25
refactor(ethereum-plugin): move connection variable to env 

Niraj-Kamdar:


dOrgJelli (changes requested):
Looks great, by why do we have ~70 files being added for CLI build test-cases? This is totally unrelated.

Also we've solved the "build folder is ignored" problem by naming it "build-cmd", see here: https://github.com/polywrap/monorepo/tree/prealpha/packages/test-cases/cases/cli/api/build-cmd

I'm going to remove these changes from this branch, and get this merged @Niraj-Kamdar

dOrgJelli (approved):

Niraj-Kamdar:
@dOrgJelli when I created PR, the problem of `gitignore` was still there but I forgot to remove it afterward.

PR #830, opened 2022-04-26
fix CLI tests

krisbitney:
Some CLI tests were recently refactored and test cases were moved into the test-cases package. Some of the test cases were in folders named `build` or included folders named `build`. This led to an issue where the `yarn reset` script would delete needed files. This PR fixes the issue by renaming the folders.


dOrgJelli:
I stumbled upon this too and realized what had happened :rofl: the CI/CD process deleted the `test-cases/cases/cli/build` directory because it was named `build` :sob: 

This is fixed in this PR that was just merged: https://github.com/polywrap/monorepo/pull/814

I'm going to close this PR, but thank you for pushing a fix for this! Sorry for the delays this may have caused.

PR #843, opened 2022-05-06
fix: optimise CI to certain extents

Niraj-Kamdar:


.github/workflows/js-ci.yaml
```diff
@@ -45,9 +44,150 @@ jobs:
...
+            ${{ runner.os }}-yarn-
+
+      - name: Install dependencies
+        run: yarn install --nonInteractive --frozen-lockfile --prefer-offline
+
+      - name: Build
+        run: yarn build
+
+      - name: Test
+        run: lerna run test:ci --no-private --ignore @web3api/*-plugin-js --ignore @web3api/cli* --ignore @web3api/client-js
```
dOrgJelli: Could we make these separate scripts in the package.json file? Similar to how we have the separate `build:...` scripts?

Niraj-Kamdar: For sure

dOrgJelli (approved):

PR #817, opened 2022-04-21
feat: improve json serializing <-> deserializing experience

Niraj-Kamdar:
- Closes: #633 
- Closes: #829

### Changelog
- Removed JSON type from scalar
- Added a better JSON serializer for serializing object to string and vice-versa.

PR #846, opened 2022-05-10
feat: AssemblyScript Automatic Object-Type JSON Serialization

### LINKED ISSUE #633: Improved JSON Serialization Bindings For AssemblyScript
AssemblyScript does not have a way of automatically generating JSON serialization bindings (that works in all cases)

Instead, we will create JSON serialization functions alongside the msgpack bindings in the schema-bind package.

In the future, this could become an optional "codegen extension", but for now it's sufficient to add this within the schema bindings for AssemblyScript directly.

### LINKED ISSUE #829: Improve JSON serialization
Currently, we don't support easy JSON serialization-deserialization, which makes the job of the wrapper developer very hard since JSON is the most used data format to be used for passing data on the current web, it's very important for us to support robust JSON serialization<->deserialization.

We initially thought of optionally creating JSON serialization <-> deserialization bindings although it will make it easy to serialize-deserialize but usage won't be similar to the JS or TS and maintenance will also be hard.

The alternative is to use the `serial-as` library which offers a clever way to serialize-deserialize JSON, where they use compiler meta-programming to make serialization - deserialization easy for users.

Niraj-Kamdar:
- Closes: #633 
- Closes: #829

### Changelog
- Removed JSON type from scalar
- Added a better JSON serializer for serializing object to string and vice-versa.
- **Not In this PR**: `Json` type alias

> Note: this will replace PR #817


packages/schema/bind/README.md
```diff
@@ -25,7 +25,6 @@ MessagePack encoded data is sent between module boundaries. Decoding of the mess
 | Bytes | bin 8/16/32 | array of 8-bit unsigned integer. |
 | BigInt | fixstr or str 8/16/32 | UTF-8 string. |
 | BigNumber | fixstr or str 8/16/32 | UTF-8 string. |
-| JSON | fixstr or str 8/16/32 | UTF-8 string. |
```
dOrgJelli: I'm not sure why we're making this change...

packages/test-cases/cases/bind/sanity/imports-local/common.graphql
```diff
@@ -14,8 +14,6 @@ type CustomType {
   optBigint: BigInt
   bignumber: BigNumber!
   optBignumber: BigNumber
-  json: JSON!
```
dOrgJelli: We should be keeping JSON as a base type

dOrgJelli (changes requested):

packages/test-cases/cases/bind/sanity/imports-local/common.graphql
Niraj-Kamdar: This will conflict with the new JSON serializer. We will add it back with the `Json` name in a separate PR which will be the alias for the string.

Niraj-Kamdar: alternative would be to keep the old serializer for now and add new serializer as alias 

`import {JSON as Json} from "@serial-as/json"` and rename it in the next PR.

dOrgJelli: Yes I think this is the best approach, since this change should only affect AssemblyScript specific code.

dOrgJelli: I'm going to go through and revert these changes now.

dOrgJelli (dismissed):

dOrgJelli (approved):

PR #847, opened 2022-05-10
Feat: Script tests to templates

cbrzn:
This PR aims to add a test script for `templates` packages, making sure that they are not broken when a new release happens. 

To achieve this, the following steps are being done:
- Iterate over all the folders inside of `templates` package, except for `scripts` and `node_modules`
- If the folder contains a `package.json` file, we read it and check if has a `build` script, if it does, the command `yarn && yarn build && yarn test` is executed, otherwise, it's only `yarn && yarn test`
- If the `package.json` does not have a `test` script, an error is thrown. This is to make sure that all the templates have tests
- If any command fails, an error is returned


.github/workflows/js-ci.yaml
```diff
@@ -6,6 +6,7 @@ on:
       - master
       - prealpha
       - prealpha-dev
+      - feat/script-tests-to-templates
```
cbrzn: This will be removed

packages/templates/api/assemblyscript/package.json
```diff
@@ -20,9 +20,9 @@
   "devDependencies": {
     "@types/jest": "27.0.3",
     "@types/node": "16.11.11",
-    "@web3api/cli": "0.0.1-prealpha.75",
-    "@web3api/ethereum-plugin-js": "0.0.1-prealpha.75",
-    "@web3api/wasm-as": "0.0.1-prealpha.75",
+    "@web3api/cli": "0.0.1-prealpha.76",
```
cbrzn: Should we change this to `latest`?

dOrgJelli: No that will break users if breaking changes are released.

These versions get automatically updated to the latest version via this script here: https://github.com/polywrap/monorepo/blob/42f0f7b075cf1aa2fefe57dbca5a0f0001fe82de/package.json#L40

dOrgJelli (dismissed):

dOrgJelli (approved):

Issue #748, opened 2022-03-14
CLI-V1: Refactor, Use Commader (Remove GlueGun)


dOrgJelli:
Work has begun here: https://github.com/polywrap/monorepo/tree/commander-refactor

PR #859, opened 2022-05-19
feat: use commander as arg-parser

Niraj-Kamdar:
- Closes: #748 

- Blocked by: #861


packages/cli/src/commands/app.ts
```diff
@@ -1,210 +1,108 @@
...
+      )
+      .option(
+        `-i, --ipfs [<${intlMsg.commands_codegen_options_i_node()}>] `,
+        `${intlMsg.commands_codegen_options_i()}`
+      )
+      .option(
+        `-e, --ens [<${intlMsg.commands_codegen_options_e_address()}>]`,
+        `${intlMsg.commands_codegen_options_e()}`
+      )
+      .action(async (options) => {
```
Niraj-Kamdar, line 57: This manual parsing isn't needed but it seems like there is some issue with `commander` parser handler so until it gets fixed, we will be using manual parsing.

packages/cli/src/__tests__/e2e/app.spec.ts
```diff
@@ -111,6 +105,9 @@ ${HELP}`);
           },
         );
 
+        console.log(output);
```
nerfZael: I think you missed these :D

packages/cli/src/__tests__/e2e/deploy.spec.ts
```diff
@@ -15,13 +15,15 @@ import yaml from "js-yaml";
...
-w3 deploy [options]
+const HELP = `Usage: w3 deploy|d [options]
+
+Deploys/Publishes a Web3API
 
 Options:
-  -h, --help                         Show usage information
-  -m, --manifest-file <path>         Path to the Web3API Deploy manifest file (default: web3api.yaml | web3api.yml)
-  -v, --verbose                      Verbose output (default: false)
+  -m, --manifest-file <path>  Path to the Web3API Plugin manifest file
```
nerfZael: Shouldn't this be "Path to the Web3API wrapper manifest" ?

packages/cli/src/commands/app.ts
```diff
@@ -1,210 +1,108 @@
...
-  fixParameters,
-  resolvePathIfExists,
-  defaultAppManifest,
   getSimpleClient,
   getTestEnvProviders,
 } from "../lib";
+import {
+  parseAppManifestFileOption,
+  parseAppCodegenDirOption,
+  // defaultAppCodegenDirOption,
```
nerfZael: What's up with the comments?

packages/cli/src/commands/create.ts
```diff
@@ -1,182 +1,143 @@
...
+      filesystem.remove(projectDir);
+    } else {
+      process.exit(8);
+    }
   }
 
-  return true;
+  generateProjectTemplate(command, lang, projectDir, filesystem)
+    .then(() => {
+      console.log();
```
nerfZael, line 126: Leftover console log?

nerfZael (changes requested):

nerfZael:
Do we still need gluegun, or could we completely remove it?

packages/cli/src/__tests__/e2e/deploy.spec.ts
Niraj-Kamdar: Nice catch!

packages/cli/src/commands/app.ts
Niraj-Kamdar: That's for fixing it back when commander default arg-parser start working.

Niraj-Kamdar: I can remove it though for now

packages/cli/src/commands/create.ts
Niraj-Kamdar: Nope that's for output formatting

Niraj-Kamdar:
> Do we still need gluegun, or could we completely remove it?

Yep but it would be better to do it in other PR.

packages/cli/src/commands/app.ts
nerfZael: Or just add "//TODO: ..."
Or better yet: "//TODO: <issue number>"

dOrgJelli (approved):

PR #861, opened 2022-05-20
Custom ensName fix for `buildAndDeployApi`

namesty:
The `buildAndDeployApi` function assumed the custom `ensName` parameter was a name without the registrar root node suffix (`.eth`, `.test`, etc.); and simply appended `.eth` after the name.

This is not correct, as the registrar contract specified could have a different root node that's not `.eth`.

This PR eliminates that assumption

### DIFF
--- a/packages/js/test-env/src/index.ts
+++ b/packages/js/test-env/src/index.ts
@@ -163,8 +163,8 @@ export async function buildAndDeployApi({
   );
 
   // create a new ENS domain
-  const domainName = ensName ?? generateName();
-  const apiEns = `${domainName}.eth`;
+  const apiEns = ensName ?? `${generateName()}.eth`;
+  const domainName = apiEns.split(".").slice(0, -1).join(".");
 
   // build API
   const {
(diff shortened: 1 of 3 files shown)


packages/js/test-env/src/index.ts
```diff
@@ -163,8 +163,8 @@ export async function buildAndDeployApi({
   );
 
   // create a new ENS domain
-  const domainName = ensName ?? generateName();
-  const apiEns = `${domainName}.eth`;
+  const apiEns = ensName ?? `${generateName()}.eth`;
+  const domainName = apiEns.split(".").slice(0, -1).join(".");
```
dOrgJelli, line 167: I don't know if this would work with a domain like `coingecko.defiwrapper.eth`... can we make a test with this type of ENS domain to see if it works?

dOrgJelli (changes requested):

packages/js/test-env/src/index.ts
dOrgJelli: In order to unblock this, we can merge this PR now, and implement subdomain support in a new PR @namesty

dOrgJelli (approved):

PR #864, opened 2022-05-26
feat(react): add env support

Niraj-Kamdar:


packages/js/react/src/provider.tsx
```diff
@@ -33,7 +33,7 @@ export function createWeb3ApiProvider(
     ClientContext: React.createContext({} as Web3ApiClient)
   };
 
-  return ({ redirects, plugins, interfaces, children }) => {
+  return ({ envs, redirects, plugins, interfaces, children }) => {
```
dOrgJelli, line 36: Thank you for this. Is there a way to make this "future proof" so we don't have to keep updating this with new ClientConfig properties?

dOrgJelli (approved):

PR #866, opened 2022-05-28
Embedded ENS Wrapper + BuildAndDeploy custom subdomains

namesty:
This PR aims to enable the `buildAndDeployApi` testenv utility function with recursive subdomain registration. It also contains some minor ENS publisher package/tests fixes.


packages/js/test-env/src/wrappers/ens/web3api.json
```diff
@@ -0,0 +1,17 @@
+{
+  "format": "0.0.1-prealpha.7",
```
cbrzn, line 2: I think it would be cool to somehow define which version we are using here from the wrapper - Just to keep a track of which implementation we are interacting with, probably adding the [version from package.json](https://github.com/polywrap/integrations/blob/main/protocol/ethereum/ens/wrapper/package.json#L5) in the name?

cbrzn (approved):

packages/js/test-env/src/wrappers/ens/web3api.json
namesty: I agree with you; however I think there's no indicator of implementation in the wrapper's build artifacts, aside from the docker build image hash. I think this will be solved by the upcoming versioning improvements

dOrgJelli (approved):

PR #870, opened 2022-05-28
CLI: Add support for Environment Variables in projects

cbrzn:
With this PR, now a developer can do the following in their `deploy` manifest
```yaml
format: 0.0.1-prealpha.1
stages:
  ipfs_deploy:
    package: ipfs
    uri: fs/./build
  from_deploy:
    package: ens
    depends_on: [ipfs_deploy, $COOL_STAGE]
    config:
      domainName: $MY_COOL_DOMAIN
      credentials: 
         name: user
         password: $COOL_PASSWORD
```
How it works (edited based on Nestor's feedback):
- `Project.loadEnvironmentVariables` expects an object `Record<string, unknown>`
- It iterates and checks every value, if it is a value or an object we iterate on it recursively until string is received
- On iteration, it checks if the string value starts with `$`, if it does, the environment variable is used as the value
- Loads environment variable using `dotenv` library and accessing it through `process.env[variableName]`.
- Throws if there's an env var declared in the schema but not loaded

The logic was added in the `Project` class with the intention of re-using this in other commands if we want. Probably the logic can improve, I am open to any suggestions :-)


packages/cli/package.json
```diff
@@ -22,7 +22,7 @@
     "prebuild": "ts-node ./scripts/generateIntlTypes.ts",
     "build:fast": "rimraf ./build && tsc --project tsconfig.build.json && yarn build:envs",
     "lint": "eslint --color -c ../../.eslintrc.js .",
-    "test": "cross-env TEST=true jest --passWithNoTests --runInBand --verbose",
+    "test": "cross-env TEST=true DOMAIN_NAME=test1.eth jest --passWithNoTests --runInBand --verbose",
```
cbrzn: This probably can be improved and we can inject the env var in any other way?

edit: let's add the env var through secrets of gh action, this should fix the test (since it's running `test:ci`) cc @dOrgJelli 

edit2: also this is a cool way to see if this really works, right?

Niraj-Kamdar: I don't think we need to use gh secrets for this. We can hardcode the envs in the workflow YAML file. There is a way to set env for jobs in gh action. secrets are only used when we want to keep envs encrypted.

cbrzn: > I don't think we need to use gh secrets for this. We can hardcode the envs in the workflow YAML file. There is a way to set env for jobs in gh action. secrets are only used when we want to keep envs encrypted.

You're totally right ser @Niraj-Kamdar thanks for your thoughts!

.github/workflows/js-ci.yaml
```diff
@@ -186,6 +186,9 @@ jobs:
       - name: Build
         run: yarn build
 
+      - name: Set environment variables for testing purposes
+        run: echo "DOMAIN_NAME=test1.eth STAGE_NAME=stage PACKAGE_NAME=package RANDOM_NAME=name" >> $GITHUB_ENV
```
Niraj-Kamdar: @cbrzn  check this out: https://docs.github.com/en/actions/learn-github-actions/environment-variables

Niraj-Kamdar (changes requested):

.github/workflows/js-ci.yaml
cbrzn: amazing

packages/cli/src/lib/project/Project.ts
```diff
@@ -40,6 +41,62 @@ export abstract class Project<TManifest extends AnyManifest> {
...
+     * @param value value of current object, we check if it is object or array,
+     * if so, we need to iterate again, otherwise just check if it is an
+     * env var and update if is true
+     */
+    const replaceValue = (
+      object: Record<string, unknown>,
+      [key, value]: [string, unknown]
+    ) => {
+      if (isObject(value)) {
+        if (Array.isArray(value)) {
```
namesty: The only nitpick is that I think the order of these conditions should be reversed. If I'm not mistaken, `typeof` of an array will yield `object`. So we should evaluate if it's an object after we are sure the value is not an array

namesty (changes requested):

packages/cli/src/lib/project/Project.ts
cbrzn: Gotcha

namesty (dismissed):

dOrgJelli (dismissed):

dOrgJelli (approved):

PR #871, opened 2022-05-29
CLI, Core: Custom named modules

cbrzn:
This PR aims to introduce the first changes to achieve https://github.com/polywrap/monorepo/issues/835. 

These are:
- Create [new manifest](https://github.com/polywrap/monorepo/pull/871/files#diff-a29f6d76a92f3becbfc5486d98399a79b62cf8644c3724d1d8936917d1b98e07R9) for a wrapper, making it able to add any string as a module (instead of only `query` and `mutation`)
- Created new manifest for plugin
- Implemented migrators for new manifest
- Update `Compiler` class to support any string as a module instead of only `query` and `mutation`
- Added tests for `Web3ApiManifest`, in these we load a manifest with custom modules
-

This touches the manifest and build command points in these notes https://hackmd.io/7MilyOKFSiKOxQXRJVlQBQ#CLIClientCore


cbrzn:
Closing this PR because spec has changed, so will create a new branch implementing the new details

PR #878, opened 2022-05-31
feat(client-js): add functionality to run workflows from client

### LINKED ISSUE #880: Allow users to run namespaced recipes from client

### LINKED ISSUE #881: Allow users to run a particular namespace of the recipe from the client

Niraj-Kamdar:
## Linked Issues
- Closes: #881 
- Closes: #880 

## Changelog
- Added functions to run workflows from client
- Added e2e tests that run workflows using client function
- Added the concept of jobs and steps to workflow
- Added the ability to run a particular job using the job id from the workflow
- Added ability to propagate output from the parent job to the child job using absolute if of a step.

## Next steps
- Add support for running workflow files from w3 cli
- Add support for basic tests by validating output
- Add option to create workflow project (post-alpha)


dOrgJelli (dismissed):
This is :fire:

dOrgJelli (dismissed):

packages/js/client/src/__tests__/e2e/workflow-test-cases.ts
```diff
@@ -0,0 +1,87 @@
+import { Workflow } from "@web3api/core-js";
```
cbrzn, line 1: noob question: would it make sense to move this file to `test-cases` package?

cbrzn (approved):
This amazing :-)

I noticed that there's no need to add the `cases` key, but it can be any other thing like `tasks`, correct? The only thing is that it needs to be consistent with the other subjobs, meaning that you'd need to do `task1`. Am I right on this ser? @Niraj-Kamdar

packages/js/client/src/__tests__/e2e/workflow-test-cases.ts
Niraj-Kamdar: I wanted to keep workflow test-cases separate from e2e test-cases

Niraj-Kamdar:
> This amazing :-)
> 
> I noticed that there's no need to add the `cases` key, but it can be any other thing like `tasks`, correct? The only thing is that it needs to be consistent with the other subjobs, meaning that you'd need to do `task1`. Am I right on this ser? @Niraj-Kamdar

Not really @cbrzn, It can be any name. I just kept those names for the test-cases but you can have any kind job hierarchy. 

Ex: `blockchain -> protocol category -> protocol ` hierarchy can showcased by `ethereum -> defi -> aave`. The only condition here is you can only depend on the output of the parent job. You can't access the output of the siblings and siblings will run in parallel. 

So, here, if there is. `ethereum ->defi->aave` and `ethereum -> nft -> opensea`then you can not access output of opensea in aave but you can access output of any common parents (here ethereum is first common parent)

dOrgJelli (approved):

Issue #714, opened 2022-03-03
I want to have custom URI resolvers


nerfZael:
- [x] #715 
- [x] #622 
    - [x] ENS
    - [x] IPFS 
    - [x] FS
- [x] #760

PR #912, opened 2022-06-10
Filesystem plugin separation of concerns

### LINKED ISSUE #622: Separate Concerns, IPFS/ENS Plugins, URI Resolver Interface
With the IPFS & ENS plugins, they both implement their respective core functionalities, along with implementing the URI Resolver interface methods.

This causes a "confusion of concerns", and leads to issues like the ones described in this thread: https://discord.com/channels/796821176743362611/908544308876050515/935542806418710548

We should create separate plugins for the IPFS & ENS URI-Resolvers, which depend on the IPFS & ENS wrappers within themselves.

pileks:
Partially closes #622 and #714 (filesystem part)

After discussion with @nerfZael the following was done:
- Introduced an interface for `filesystem`-like plugins
- Extracted `uri-resolver` interface implementation from `filesystem` plugin into its own `fs-uri-resolver` plugin
- Implemented basic `filesystem` plugin which implements the `filesystem` interface
- Introduced a new build command for interfaces that follow the `@web3api/*-interface` pattern and added it to the main build command

Advice/feedback would be much appreciated!


packages/core-interfaces/filesystem/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "name": "@webapi3/filesystem-interface",
+  "description": "Web3API Filesystem Interface",
+  "private": true,
+  "version": "0.0.1-prealpha.84",
+  "scripts": {
+    "build": "rimraf ./build && node ../../../dependencies/node_modules/@web3api/cli/bin/w3 build",
+    "lint": "eslint --color -c ../../../.eslintrc.js .",
+    "test:env:up": "npx w3 test-env up",
+    "test:env:down": "npx w3 test-env down",
+    "deploy": "npx w3 build --ipfs http://localhost:5001"
```
nerfZael: there's no --ipfs option of the build command anymore. You can replace this with `npx w3 deploy`

packages/core-interfaces/filesystem/src/mutation.graphql
```diff
@@ -0,0 +1,8 @@
+type Mutation {
+  writeFile(path: String!, data: Bytes!): Boolean
+
+  mkdir(path: String!): Boolean
+
+  rm(path: String!): Boolean
```
nerfZael: we should add optional options for `recursive` and `force`

packages/core-interfaces/filesystem/src/mutation.graphql
```diff
@@ -0,0 +1,8 @@
+type Mutation {
+  writeFile(path: String!, data: Bytes!): Boolean
+
+  mkdir(path: String!): Boolean
```
nerfZael: we should add optional options for `recursive`

packages/js/plugins/filesystem/src/mutation/index.ts
```diff
@@ -0,0 +1,41 @@
...
+import fs from "fs";
+
+export type MutationConfig = Record<string, unknown>;
+
+export class Mutation extends Module<MutationConfig> {
+  async writeFile(
+    input: Input_writeFile,
+    _client: Client
+  ): Promise<boolean | null> {
+    fs.writeFileSync(input.path, Buffer.from(input.data));
```
nerfZael: Even though the call on the wrapper side is blocking, this part happens client-side and could be awaited to give other client-side tasks more processing time. E.g you can use:
```typescript
import { promises as fsPromises } from 'fs';
await fsPromises.writeFile('file.txt', 'data');
```

nerfZael: Same applies to all other functions ofc

packages/js/plugins/filesystem/src/mutation/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    return true;
```
nerfZael: If we're returning a boolean instead of void, we might as well catch the error and return false in that case

packages/js/plugins/filesystem/src/query/index.ts
```diff
@@ -1,59 +1,32 @@
...
-        } catch (e) {
-          // TODO: logging
-        }
-      }
-    }
+  async readFileAsString(
+    input: Input_readFileAsString,
+    _client: Client
+  ): Promise<string> {
+    // TODO: Maybe check if input.encoding is within supported encodings?
```
nerfZael: Maybe use an enum instead of string for encoding?

packages/js/plugins/filesystem/src/query/index.ts
```diff
@@ -1,59 +1,32 @@
...
-      "web3api.yaml",
-      "web3api.yml",
-    ];
-
-    let manifest: string | undefined;
-
-    for (const manifestSearchPattern of manifestSearchPatterns) {
-      const manifestPath = path.resolve(input.path, manifestSearchPattern);
+  async readFile(input: Input_readFile, _client: Client): Promise<ArrayBuffer> {
+    return fs.promises.readFile(input.path);
```
nerfZael: Oh, here you ARE using promises :D

packages/js/plugins/filesystem/src/__tests__/e2e.spec.ts
```diff
@@ -1,121 +1,141 @@
...
+    const encoding = "utf-8";
+
+    const expectedContents = fs.readFileSync(sampleFilePath, {
+      encoding: encoding,
     });
 
-    expect(deploy.errors).toBeFalsy();
-    expect(deploy.data).toBeTruthy();
-    expect(deploy.data?.deployContract.indexOf("0x")).toBeGreaterThan(-1);
+    const fsReadFileAsStringResult = await Filesystem_Query.readFileAsString(
```
nerfZael: This might be a bit too verbose a name for the only "result" in this test case. Personally, I'd rather just use `result`. What do you think?

packages/js/plugins/filesystem/src/__tests__/e2e.spec.ts
```diff
@@ -1,121 +1,141 @@
...
+    const fsExistsResult_fileMissing = await Filesystem_Query.exists(
+      { path: nonExistentFilePath },
+      client
+    );
+
+    expect(fsExistsResult_fileMissing.error).toBeFalsy();
+    expect(fsExistsResult_fileMissing.data).toBe(false);
+  });
+
+  it("should write data to a file and succesfully remove it", async () => {
```
nerfZael: I think it'd be better to have a test case per wrapper method. So, separating writing and removing. This is even more important once you add rm options

packages/js/plugins/uri-resolvers/filesystem-resolver/src/query/index.ts
```diff
@@ -0,0 +1,79 @@
...
+export type QueryConfig = Record<string, unknown>;
+
+import path from "path";
+
+export class Query extends Module<QueryConfig> {
+  async tryResolveUri(
+    input: Input_tryResolveUri,
+    _client: Client
+  ): Promise<UriResolver_MaybeUriOrManifest | null> {
+    if (input.authority !== "fs") {
```
nerfZael: @dOrgJelli Should we start renaming `fs` authority to `file` ?

packages/js/plugins/uri-resolvers/filesystem-resolver/src/query/index.ts
```diff
@@ -0,0 +1,79 @@
...
+        try {
+          const manifestResult = await Filesystem_Query.readFileAsString(
+            { path: manifestPath, encoding: "utf-8" },
+            _client
+          );
+          if (manifestResult.error) {
+            console.warn(manifestResult.error);
+          }
+          manifest = manifestResult.data;
+        } catch (e) {
```
nerfZael: We need better error handling, but I know you just re-implemented what was here originally. We can leave this as is since there's a PR up to improve resolver error handling and it's on my TODO list to get that implemented soon

nerfZael (changes requested):

packages/js/plugins/filesystem/src/__tests__/e2e.spec.ts
pileks: I agree that, especially when there is only one result object in the test, we can name it `result`.
I'd still keep `xResult` when there are more results being loaded - not a fan of reusing variables.
Thoughts?

packages/js/plugins/filesystem/src/__tests__/e2e.spec.ts
pileks: Makes sense. Then we'd use `fs` to create the folder/file being removed.

packages/js/plugins/filesystem/src/mutation/index.ts
pileks: I'd lean on throwing an error and not returning anything in that case.

packages/js/plugins/uri-resolvers/filesystem-resolver/src/query/index.ts
pileks: If I have a vote in this, I'd go with doing that as part of a separate issue simply to keep the scope of the PR at a manageable level.

packages/js/plugins/filesystem/src/mutation/index.ts
nerfZael: After a call with @pileks we decided it'd be better to just let the plugin throw so that the app can catch the error instead of masking it

packages/core-interfaces/filesystem/src/schema.graphql
```diff
@@ -0,0 +1,25 @@
+type Module {
+  readFile(path: String!): Bytes!
+  readFileAsString(path: String!, encoding: Encoding): String!
```
pileks: @nerfZael commented that @dOrgJelli mentioned in the IPFS PR that we should leave decoding to the consumer of the plugin (remove any `asString` stuff).
Should we also remove `readFileAsString` from here in favor of the same approach?
This would also mean significantly simplifying the interface, as we would not need to have the `Encoding` enum below.

packages/core-interfaces/filesystem/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "name": "@polywrap/filesystem-interface",
+  "description": "Polywrap Filesystem Interface",
+  "private": true,
+  "version": "0.0.1-prealpha.88",
+  "scripts": {
+    "build": "node ../../../dependencies/node_modules/polywrap/bin/polywrap build",
+    "lint": "eslint --color -c ../../../.eslintrc.js .",
+    "test:env:up": "npx polywrap infra up --modules=eth-ens-ipfs",
+    "test:env:down": "npx polywrap infra down --modules=eth-ens-ipfs",
+    "deploy": "npx polywrap build --ipfs http://localhost:5001"
```
nerfZael: `npx polywrap deploy`

packages/js/plugins/uri-resolvers/filesystem-resolver/src/index.ts
```diff
@@ -0,0 +1,90 @@
...
+import path from "path";
+
+export type ModuleConfig = Record<string, unknown>;
+
+export class FilesystemResolverPlugin extends Module<ModuleConfig> {
+  async tryResolveUri(
+    input: Input_tryResolveUri,
+    _client: Client
+  ): Promise<UriResolver_MaybeUriOrManifest | null> {
+    if (input.authority !== "fs") {
```
nerfZael: After talking with @dOrgJelli we decided to cover both `fs` and `file` authorities

packages/js/plugins/uri-resolvers/filesystem-resolver/src/index.ts
```diff
@@ -0,0 +1,90 @@
...
+      return null;
+    }
+
+    const manifestSearchPatterns = [
+      "polywrap.json",
```
nerfZael: Only "polywrap.json" is currently valid in the build artifacts, you can remove others

packages/js/plugins/uri-resolvers/filesystem-resolver/src/__tests__/e2e.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+  beforeAll(async () => {
+    await initTestEnvironment();
+
+    const config: Partial<PolywrapClientConfig> = {
+      plugins: [
+        {
+          uri: "wrap://ens/fs-resolver.polywrap.eth",
+          plugin: filesystemResolverPlugin({ query: {} }),
+        },
+        // IPFS is required for downloading Web3API packages
```
nerfZael: You can remove these comments

packages/js/plugins/filesystem/src/utils/encodingUtils.ts
```diff
@@ -0,0 +1,46 @@
+import { Filesystem_Encoding, Filesystem_EncodingEnum } from "../wrap";
+
+export const filesystemEncodingToBufferEncoding = (
+  encoding: Filesystem_Encoding | null | undefined
+): BufferEncoding => {
+  switch (encoding) {
+    case Filesystem_EncodingEnum.ASCII:
+    case "ASCII":
```
nerfZael: This can be a string?

packages/js/plugins/filesystem/src/utils/encodingUtils.ts
```diff
@@ -0,0 +1,46 @@
+import { Filesystem_Encoding, Filesystem_EncodingEnum } from "../wrap";
+
+export const filesystemEncodingToBufferEncoding = (
```
nerfZael: I'd name this file after this function: `filesystemEncodingToBufferEncoding.ts`

packages/js/plugins/filesystem/src/__tests__/e2e.spec.ts
```diff
@@ -1,119 +1,228 @@
...
     };
     client = new PolywrapClient(config);
   });
 
-  afterAll(async () => {
-    await stopTestEnvironment();
+  afterEach(async () => {
+    // Clean up temp files/folders in case test failed
+    if (fs.existsSync(tempFilePath)) {
+      await fs.promises.rm(tempFilePath, { force: true, recursive: true });
```
nerfZael: What does recursive do when deleting a file?

packages/js/plugins/filesystem/src/__tests__/e2e.spec.ts
```diff
@@ -1,119 +1,228 @@
...
+    }
+
+    if (fs.existsSync(tempFolderPath)) {
+      await fs.promises.rm(tempFolderPath, { force: true, recursive: true });
+    }
+  });
+
+  it("should read a file", async () => {
```
nerfZael: `can read a file` or `should be able to read a file`
I prefer the first one

packages/js/plugins/filesystem/src/__tests__/e2e.spec.ts
```diff
@@ -1,119 +1,228 @@
...
+      { path: sampleFilePath },
+      client
+    );
+
+    expect(result.error).toBeFalsy();
+    expect(result.data).toEqual(expectedContents);
+  });
+
+  it("should fail reading a nonexistent file", async () => {
+    const nonExistentFilePath = `${sampleFilePath}nonexistent`;
```
nerfZael: Maybe just `nonexistent` without the ${sampleFilePath} prefix ?

packages/js/plugins/filesystem/src/__tests__/e2e.spec.ts
nerfZael: You could just delete the whole sample dir and create needed files when a test needs them

packages/js/plugins/filesystem/src/__tests__/e2e.spec.ts
```diff
@@ -1,119 +1,228 @@
...
+    );
+
+    expect(result.data).toBe(true);
+
+    let directoryExists = fs.existsSync(tempFolderPath);
+
+    expect(directoryExists).toBe(true);
+  });
+
+  it("should create a folder recursively", async () => {
```
nerfZael: Either use dirs everywhere or folders when naming
I prefer directory

packages/js/plugins/filesystem/src/__tests__/e2e.spec.ts
```diff
@@ -1,119 +1,228 @@
...
-    const expectedSchema = await fs.promises.readFile(`${fsPath}/schema.graphql`, "utf-8");
+    expect(result.error).toBeFalsy();
+    expect(result.data).toBe(true);
+
+    const fileExists = fs.existsSync(tempFilePath);
+
+    expect(fileExists).toBe(false);
+  });
+
+  it("should remove a file recursively", async () => {
```
nerfZael: Don't you mean  `remove a directory recursively` ?

packages/js/plugins/filesystem/src/__tests__/e2e.spec.ts
nerfZael: I'd also add this removal logic in beforeAll in case someone kills the test process in the middle of execution

nerfZael (changes requested):

nerfZael:
After talking with @dOrgJelli we decided file system should be two words, so: `FileSystem`, `fileSystem`, `file-system` etc

packages/js/plugins/uri-resolvers/filesystem-resolver/src/__tests__/e2e.spec.ts
pileks: I'd only delete these in the setup phase - the ones in the actual test seem useful enough.

packages/js/plugins/filesystem/src/utils/encodingUtils.ts
pileks: AFAIK, Enums in wrappers are actually of type `Something_Enum | Something_EnumString`:
This is the otuput in `types.ts` for an enum:
```javascript
export enum Filesystem_EncodingEnum {
  ASCII,
  UTF8,
  UTF16LE,
  UCS2,
  BASE64,
  BASE64URL,
  LATIN1,
  BINARY,
  HEX,
}

export type Filesystem_EncodingString =
  | "ASCII"
  | "UTF8"
  | "UTF16LE"
  | "UCS2"
  | "BASE64"
  | "BASE64URL"
  | "LATIN1"
  | "BINARY"
  | "HEX"

export type Filesystem_Encoding = Filesystem_EncodingEnum | Filesystem_EncodingString;
```

packages/js/plugins/filesystem/src/utils/encodingUtils.ts
pileks: Would we then want to `export default`?

packages/js/plugins/filesystem/src/__tests__/e2e.spec.ts
pileks: Nothing, really. Copy-paste :grin:

packages/js/plugins/uri-resolvers/filesystem-resolver/src/query/index.ts
nerfZael: After talking with @dOrgJelli  we decided to support both

nerfZael (dismissed):

dOrgJelli (dismissed):

PR #913, opened 2022-06-11
Compose refactor

namesty:


packages/test-cases/cases/compose/shared-env/output/module.graphql
```diff
@@ -35,13 +35,13 @@ directive @annotate(type: String!) on FIELD
...
 
-type Query {
+type Module {
   method(
     str: String!
   ): String!
 }
 
-type QueryEnv {
+type ModuleEnv {
```
dOrgJelli: This test can be removed entirely

dOrgJelli (approved):

dOrgJelli (approved):

PR #922, opened 2022-06-15
Added Tezos Infra Module

### LINKED ISSUE #52: create tezos test-env as an infra module

namesty:
Closes polywrap/integrations#52

This PR adds a Tezos blockchain environment, as a default infra module


packages/cli/src/lib/defaults/infra-modules/tezos/cmds.js
```diff
@@ -0,0 +1,7 @@
+const CMDS = {
```
dOrgJelli, line 1: @namesty I'm confused by how these js modules get used, as I don't see them referenced in the docker-compose.yaml or Dockerfile.

Are they useful to users in some way? If so, how can they access them? Is this something common you think we should do (bundle helper scripts w/ the infra modules)?

dOrgJelli (changes requested):

packages/cli/src/lib/defaults/infra-modules/tezos/cmds.js
namesty: Yes, they do seem unrelated, I'll remove them

Niraj-Kamdar (commented):
Do tezos have test-env-js like we have for default test-env?

This will be out of scope for this PR but would like to suggest that we should have better way to check whether all the services are started or stopped instead of sleeping for arbitrary time and hope it will be started

dOrgJelli:
Going to close this inactivity. Additionally we should publish this within another repo instead of adding it as a baked-in default within the CLI.

PR #923, opened 2022-06-16
Rename Web3API to Polywrap

dOrgJelli:
closes: https://github.com/polywrap/monorepo/issues/732


package.json
```diff
@@ -23,25 +22,24 @@
...
     "dependencies:install": "cd dependencies && yarn",
-    "preinstall": "yarn dependencies:install",
     "build": "yarn build:core && yarn build:plugins && yarn build:client && yarn build:test-env && yarn build:cli && yarn build:plugins:patch",
-    "build:core": "lerna run build --no-private --ignore @web3api/*-plugin-js --ignore @web3api/cli* --ignore @web3api/react --ignore @web3api/test-env-js",
-    "build:plugins": "lerna run build --scope @web3api/*-plugin-js --concurrency 1",
-    "build:client": "lerna run build --scope @web3api/client-js --scope @web3api/react",
-    "build:test-env": "lerna run build --scope @web3api/test-env-js",
-    "build:cli": "lerna run build --scope @web3api/cli",
-    "build:plugins:patch": "lerna run codegen:patch --scope @web3api/*-plugin-js --concurrency 1",
+    "build:core": "lerna run build --no-private --ignore @polywrap/*-plugin-js --ignore polywrap --ignore @polywrap/clie* --ignore @polywrap/react --ignore @polywrap/test-env-js",
```
cbrzn: Is the `@polywrap/clie*` on purpose?

packages/cli/package.json
```diff
@@ -1,24 +1,24 @@
 {
-  "name": "@web3api/cli",
-  "description": "Web3API CLI",
+  "name": "polywrap",
```
cbrzn, line 2: What's the reason for removing the `/cli` here?

.gitignore
```diff
@@ -5,14 +5,15 @@ yarn-error.log
 coverage
 .vscode
 *.log
-w3
-.w3
+wrap
```
cbrzn, line 8: Noob question: What's the reason for having different generated folders being named `wrap` and `.polywrap`?  My first thought is that we should be consistent with the generated folders, but I am almost sure that I am missing something

packages/cli/package.json
nerfZael: Seems odd

packages/cli/src/__tests__/e2e/help.spec.ts
```diff
@@ -1,22 +1,22 @@
...
 Options:
   -h, --help                  display help for command
 
 Commands:
   app|a                       Build/generate types for your app
-  build|b [options]           Builds a Web3API
-  codegen|g [options]         Auto-generate API Types
-  create|c                    Create a new project with w3 CLI
-  deploy|d [options]          Deploys/Publishes a Web3API
+  build|b [options]           Builds a Polywrap
```
nerfZael: What's a "Polywrap"?

packages/cli/package.json
dOrgJelli: It's so that users can simply run `npx polywrap ...` instead of `npx @polywrap/cli ...`

.gitignore
dOrgJelli: They are at 2 different levels of "abstraction" when thinking about Polywrap vs WRAP.

`.polywrap` is the CLI's cache directory. This is "Polywrap Toolchain Specific".

`wrap` is the wrapper's codegen folder, which contains "WRAP Standard Specific" bindings for encoding + host imports (`__wrap_subinvoke` for example).

dOrgJelli: An analogy is having a `.github` folder that contains GitHub specific files, and also having a `.git` folder that has standardized git protocol data.

package.json
dOrgJelli: Nice catch, thank you!

packages/cli/src/__tests__/e2e/create.spec.ts
```diff
@@ -1,29 +1,30 @@
...
 
-Create a new project with w3 CLI
+Create a new project with polywrap CLI
 
 Options:
   -h, --help                          display help for command
 
 Commands:
-  api [options] <language> <name>     Create a Web3API project langs:
+  wasm [options] <language> <name>    Create a Polywrap project langs:
```
namesty: Shouldn't it be `wrapper` here? Or maybe `wasm wrapper`? Just `wasm` seems a bit odd to me

packages/cli/src/__tests__/e2e/help.spec.ts
```diff
@@ -1,22 +1,22 @@
...
+  codegen|g [options]         Auto-generate Wrapper Types
+  create|c                    Create a new project with polywrap CLI
+  deploy|d [options]          Deploys/Publishes a Polywrap
   plugin|p                    Build/generate types for the plugin
-  query|q [options] <recipe>  Query Web3APIs using recipe scripts
-  infra|i [options] <action>  Manage infrastructure for your Web3API
+  query|q [options] <recipe>  Query Polywraps using recipe scripts
+  infra|i [options] <action>  Manage infrastructure for your Polywrap
```
namesty: I think the appropriate noun here would be `wrapper`. "Query wrappers using recipe..." etc.

packages/cli/src/__tests__/e2e/infra.spec.ts
```diff
@@ -14,21 +14,21 @@ const testCaseRoot = path.join(GetPathToCliTestFiles(), "infra");
   const getTestCaseDir = (index: number) =>
     path.join(testCaseRoot, testCases[index]);
 
-const HELP = `Usage: w3 infra|i <action> [options]
+const HELP = `Usage: polywrap infra|i <action> [options]
 
-Manage infrastructure for your Web3API
+Manage infrastructure for your Polywrap
```
namesty: Same here, shouldn't it be wrapper? In the messages below, `Polywrap's infra` makes sense to me but here it should probably be wrapper

packages/cli/src/__tests__/e2e/no-command.spec.ts
```diff
@@ -1,22 +1,22 @@
...
-  deploy|d [options]          Deploys/Publishes a Web3API
+  build|b [options]           Builds a Polywrap
+  codegen|g [options]         Auto-generate Wrapper Types
+  create|c                    Create a new project with polywrap CLI
+  deploy|d [options]          Deploys/Publishes a Polywrap
   plugin|p                    Build/generate types for the plugin
-  query|q [options] <recipe>  Query Web3APIs using recipe scripts
-  infra|i [options] <action>  Manage infrastructure for your Web3API
+  query|q [options] <recipe>  Query Polywraps using recipe scripts
+  infra|i [options] <action>  Manage infrastructure for your Polywrap
```
namesty: Same as above

packages/core-interfaces/logger/README.md
```diff
@@ -9,5 +9,5 @@
 ## Start Test Environment
 `yarn test:env:up`  
 
-## Build & Deploy Web3API
+## Build & Deploy Polywrap
```
namesty: Same as above

packages/cli/src/__tests__/e2e/help.spec.ts
dOrgJelli: Nice catch, fixed!

packages/cli/src/__tests__/e2e/help.spec.ts
dOrgJelli: Nice catch, fixed!

packages/cli/src/__tests__/e2e/infra.spec.ts
dOrgJelli: Nice catch, fixed!

packages/cli/src/__tests__/e2e/create.spec.ts
dOrgJelli: Fixed!

PR #924, opened 2022-06-16
fix(bind): map of array weren't generating invalid bindings

Niraj-Kamdar:
- Closes: #883


dOrgJelli (approved):
:fire: :fire: :fire: @Niraj-Kamdar

PR #930, opened 2022-06-18
IPFS plugin separation of concerns

### LINKED ISSUE #622: Separate Concerns, IPFS/ENS Plugins, URI Resolver Interface

pileks:
Partially closes #622 and #714 (IPFS part)

After discussion with @nerfZael the following was done:

Introduced an interface for IPFS-like plugins
Extracted uri-resolver interface implementation from IPFS plugin into its own `ipfs-uri-resolver-plugin-js` plugin
`ipfs-plugin-js` now contains only basic IPFS-related functions, with the `ipfs-uri-resolver-plugin-js` depending on the `ipfs-plugin-js`.

Introduced a new build command for interfaces that follow the @web3api/*-interface pattern and added it to the main build command - same as #912

**Possible circular dependency noticed:**
Interfaces require CLI to be built, which results in a `CLI -> Plugins -> Interfaces -> CLI` circular dependency. Advice on how to proceed is very much appreciated.

Remaining work:

- [x] Add tests for `ipfs-uri-resolver-plugin-js`
- [x] Add tests for `ipfs-plugin-js` <- I noticed these were missing

@dOrgJelli We commented on moving the resolver plugins into a separate folder - I would like to do this once we finish up the separation of concerns work so that we move all 3 resolver plugins as part of a single PR. Hope that's okay.

Advice/feedback would be much appreciated!


packages/core-interfaces/ipfs/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "name": "@polywrap/ipfs-interface",
+  "description": "Polywrap Ipfs Interface",
+  "private": true,
+  "version": "0.0.1-prealpha.86",
+  "scripts": {
+    "build": "node ../../../dependencies/node_modules/polywrap/bin/polywrap build",
+    "lint": "eslint --color -c ../../../.eslintrc.js .",
+    "test:env:up": "npx polywrap infra up --modules=eth-ens-ipfs",
+    "test:env:down": "npx polywrap infra down --modules=eth-ens-ipfs",
+    "deploy": "npx polywrap build --ipfs http://localhost:5001"
```
nerfZael: Should be `npx polywrap deploy` or `node ../../../dependencies/node_modules/polywrap/bin/polywrap deploy`

packages/js/plugins/ipfs/package.json
```diff
@@ -12,12 +12,12 @@
     "build"
   ],
   "scripts": {
-    "build": "rimraf ./build && tsc --project tsconfig.build.json",
+    "build": "rimraf ./build && yarn codegen && tsc --project tsconfig.build.json",
     "codegen": "node ../../../../dependencies/node_modules/polywrap/bin/polywrap plugin codegen",
     "codegen:patch": "node ../../../cli/bin/polywrap plugin codegen && rimraf ./src/wrap",
     "lint": "eslint --color -c ../../../../.eslintrc.js src/",
     "test": "jest --passWithNoTests --runInBand --verbose",
-    "test:ci": "jest --passWithNoTests --runInBand --verbose",
```
nerfZael: Seems like a weird change. Why does this call `yarn codegen` ?

packages/core-interfaces/ipfs/.gitignore
```diff
@@ -0,0 +1,3 @@
+build
+node_modules
+w3
```
nerfZael: There's no `w3` directory, it's `wrap` now

packages/core-interfaces/ipfs/polywrap.yaml
```diff
@@ -0,0 +1,5 @@
+format: 0.0.1-prealpha.9
+name: Filesystem
```
nerfZael: IPFS

packages/js/plugins/uri-resolvers/ipfs-resolver/src/index.ts
```diff
@@ -0,0 +1,133 @@
...
+// eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports
+const isIPFS = require("is-ipfs");
+
+export interface IpfsResolverPluginConfig extends Record<string, unknown> {
+  provider: string;
+  fallbackProviders?: string[];
+}
+
+export class IpfsResolverPlugin extends Module<IpfsResolverPluginConfig> {
+  public static isCID(cid: string): boolean {
```
nerfZael: I don't think this needs to be public

packages/js/plugins/uri-resolvers/ipfs-resolver/src/index.ts
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    if (!IpfsResolverPlugin.isCID(input.path)) {
+      // Not a valid CID
+      return { manifest: null, uri: null };
+    }
+
+    const manifestSearchPatterns = [
+      "polywrap.json",
+      "polywrap.yaml",
```
nerfZael: You can remove .yaml and .yml, they're no longer valid

packages/js/plugins/ipfs/src/index.ts
```diff
@@ -54,27 +53,20 @@ export class IpfsPlugin extends Module<IpfsPluginConfig> {
     return isIPFS.cid(cid) || isIPFS.cidPath(cid) || isIPFS.ipfsPath(cid);
```
nerfZael: I dont think `public static isCID(cid: string): boolean` is being used anywhere in this plugin

packages/js/plugins/uri-resolvers/ipfs-resolver/src/__tests__/e2e.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  initTestEnvironment,
+  providers,
+  stopTestEnvironment,
+  buildAndDeployWrapper,
+} from "@polywrap/test-env-js";
+
+import { ipfsResolverPlugin } from "..";
+import { ipfsPlugin } from "@polywrap/ipfs-plugin-js";
+
+import { IpfsClient } from "../../../../ipfs/src/utils/IpfsClient";
```
nerfZael: I'd rather we copy the IpfsClient type into the tests/utils or somewhere, even though that's duplicating code than relying on the IPFS plugin.

packages/js/plugins/ipfs/src/__tests__/e2e.spec.ts
```diff
@@ -1,5 +1,83 @@
...
+    expect(result.error).toBeFalsy();
+
+    expect(result.data).toEqual({
+      cid: `/ipfs/${sampleFileIpfsInfo.hash.toString()}`,
+      provider: providers.ipfs,
+    });
+  });
+
+  it("Should add a file successfully", async () => {
+    let result = await Ipfs_Module.addFile({ data: sampleFileBuffer }, client);
```
nerfZael: Technically, this won't add the file since it was added in beforeAll. Lets use a different file here

nerfZael (changes requested):

packages/js/plugins/ipfs/package.json
pileks: My initial build step here didn't have codegen, so tests would fail after build. Now that it has codegen, it's not required before running tests, you're right.

packages/js/client/src/__tests__/core/plugin-wrapper.spec.ts
```diff
@@ -17,6 +17,7 @@ const defaultPlugins = [
   "wrap://ens/sha3.polywrap.eth",
   "wrap://ens/graph-node.polywrap.eth",
   "wrap://ens/fs.polywrap.eth",
+  "wrap://ens/ipfs-resolver.polywrap.eth",
```
cbrzn, line 21: Hey @nerfZael would it make sense to remove `ipfs.polywrap.eth` since the only purpose of it was to help us to resolve and now this uri takes care of that?

packages/js/plugins/uri-resolvers/ipfs-resolver/polywrap.plugin.yaml
```diff
@@ -0,0 +1,10 @@
+format: 0.0.1-prealpha.3
+name: Ipfs
+language: plugin/typescript
+module: ./src/index.ts
+schema: ./src/schema.graphql
+import_redirects:
+  - uri: "ens/uri-resolver.core.polywrap.eth"
+    schema: ../../../../core-interfaces/uri-resolver/src/query.graphql
```
cbrzn: Let's remove the `query` word and rename the file to `schema.graphql`

packages/js/plugins/uri-resolvers/ipfs-resolver/src/__tests__/e2e.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+        },
+      ],
+    });
+  });
+
+  afterAll(async () => {
+    await stopTestEnvironment();
+  });
+
+  it("Should successfully resolve a deployed wrapper - e2e", async () => {
```
cbrzn, line 59: I am so happy for these tests, thank you :pray:

packages/js/plugins/uri-resolvers/ipfs-resolver/src/index.ts
```diff
@@ -0,0 +1,129 @@
...
+      }
+    }
+
+    if (manifest) {
+      return { uri: null, manifest };
+    } else {
+      // Nothing found
+      return { uri: null, manifest: null };
+    }
+  }
```
cbrzn:
```suggestion
    return { uri: null, manifest: manifest ?? null }
```

packages/js/plugins/uri-resolvers/ipfs-resolver/src/index.ts
```diff
@@ -0,0 +1,129 @@
...
+          },
+        },
+        client
+      );
+
+      if (catResult.data) {
+        return catResult.data;
+      } else {
+        return null;
+      }
```
cbrzn:
```suggestion
      return catResult?.data ?? null
```

packages/js/plugins/uri-resolvers/ipfs-resolver/src/utils/IpfsClient.ts
```diff
@@ -0,0 +1,19 @@
+import CID from "cids";
+
+export interface IpfsClient {
+  add(data: Uint8Array, options?: unknown): Promise<IpfsFileInfo[]>;
```
cbrzn: Probably this is non sense but should the interface of `IpfsClient` defined in the `ipfs-resolver` have the `add` method? I guess it should just be aware of getting content

cbrzn (commented):
Awesome stuff ser :fire:  

Feel free to ignore my code styles suggestions :-D

packages/js/client/src/__tests__/core/plugin-wrapper.spec.ts
nerfZael: `ipfs-resolver.polywrap.eth` plugin uses the `ipfs.polywrap.eth` plugin to read from IPFS. Without it, it wouldn't be able to resolve.

packages/js/plugins/uri-resolvers/ipfs-resolver/src/utils/IpfsClient.ts
nerfZael: This interface should be only for tests and your comment made me check that and find a larger issue, so very nice catch :D

packages/js/plugins/uri-resolvers/ipfs-resolver/polywrap.plugin.yaml
pileks: This is referenced in multiple PRs right now - If it's okay with you, I'd do this as part of an end-of-PRs "cleanup PR" which would redo some of the folder structure, and change naming. We should also think about building the interface rather than referencing directly from `src` - something I've done in the FS MR (#912) and this one for their respective interfaces.

nerfZael (dismissed):

packages/core-interfaces/ipfs/.gitignore
```diff
@@ -0,0 +1,3 @@
+build
```
dOrgJelli: Hey @pileks, no need to have `.gitignore` or `.nvmrc` files in the packages within the monorepo, since these are typically covered by the root monorepo files.

If there are special needs by the package though, that's when you need to create one at the sub-module level.

dOrgJelli (approved):

PR #931, opened 2022-06-19
Remove wrap-man Folders

dOrgJelli:
The `wrap-man` folders were added to solve the circular dependency we have between the plugins, client, and CLI. Now that we have a version of the CLI published that uses the latest version of the plugin's interface + codegen typings, we can remove these manual codegen folders.


cbrzn (approved):
lgtm :green_apple:

namesty (approved):

PR #936, opened 2022-06-21
Replace Nullable with Option in assemblyscript

### LINKED ISSUE #809: Question: use Option type from as-container instead of Nullable type that we are using currently
Benefits:
- thoroughly tested code
- rust-inspired 
- supports any type to be wrapped in this container primitive unlike our Nullable that just allows the base type to be nullable.
- Similar exp to the Result type since from the same package.

Checkout : https://github.com/yjhmelody/as-container for more info

fetsorn:
Replaces custom Nullable type with Option type from as-container. Resolves #809.

A recap:
We want to represent values which could be "null".
Assemblyscript has nullable reference types, e.g. "string | null", but does not support nullable basic types, e.g. "u32", so we wrap primitives in a custom Nullable type, like "Nullable<u32>".
This PR provides a more consistent approach, wrapping both reference types and basic types in an Option container, like "Option.Some<u32>(10)" and "Option.None<u32>()". The Option container has a robust implementation and a rust-inspired interface, similar to the Result type we are already using.


fetsorn:
Ooops, more changes need to be done, will resubmit.

PR #938, opened 2022-06-21
Chore: Rollback JSON Serialization in AS

cbrzn:
This PR aims to remove the additions in PR https://github.com/polywrap/monorepo/pull/846


packages/test-cases/cases/wrappers/wasm-rs/json-type/src/lib.rs
```diff
@@ -3,10 +3,26 @@ pub use wrap::*;
 use polywrap_wasm_rs::JSON;
 use serde_json::*;
 
-pub fn from_json(input: InputFromJson) -> Pair {
```
krisbitney, line 6: I propose keeping the from_json and to_json tests for the Rust JSON implementation. They should still work even if we've removed the analogous capability from the AS code.

krisbitney (approved):
Looks good!

dOrgJelli (approved):

PR #945, opened 2022-06-22
Chore: Logger core interface to latest changes

cbrzn:
This PR aims to update the logger core interface to make it compliant with single module approach so we can use it in the plugin and unblock https://github.com/polywrap/integrations/issues/65


cbrzn:
After talking with @Niraj-Kamdar this interface will be moved to `integrations` repository

cbrzn:
Reopening because of https://github.com/polywrap/integrations/pull/70#issuecomment-1164433608

dOrgJelli (approved):

PR #947, opened 2022-06-22
fix: allow invoking with Uint8Array

Niraj-Kamdar:
- [x] allow calling invoke with Uint8Array
- [x] use Uint8Array instead of ArrayBuffer for Bytes on Plugin side because msgpack always decode Bytes into Uint8Array


packages/js/client/src/plugin/PluginWrapper.ts
```diff
@@ -80,7 +80,19 @@ export class PluginWrapper extends Wrapper {
       let jsInput: Record<string, unknown>;
 
       // If the input is a msgpack buffer, deserialize it
-      if (input instanceof ArrayBuffer) {
+      if (input instanceof Uint8Array) {
```
krisbitney: What do you think about this? It adds an extra test but reduces repeat code.

```
const isUint8Array = input instanceof  Uint8Array;
if (isUint8Array || input instanceof ArrayBuffer) {
  const result = msgpackDecode(isUint8Array ? input.buffer : input);

  Tracer.addEvent("msgpack-decoded", result);

  if (typeof result !== "object") {
    throw new Error(
      `PluginWrapper: decoded MsgPack input did not result in an object.\nResult: ${result}`
    );
  }

  jsInput = result as Record<string, unknown>;
} else {
  jsInput = input;
}
```

krisbitney (dismissed):

packages/js/client/src/plugin/PluginWrapper.ts
Niraj-Kamdar: Thanks for suggestion I rewrote that part

packages/templates/wasm/assemblyscript/src/workflows/config.ts
```diff
@@ -0,0 +1,77 @@
...
+        implementations: [
+          "wrap://ens/ipfs.polywrap.eth",
+          "wrap://ens/ens.polywrap.eth",
+        ],
+      },
+      {
+        interface: coreInterfaceUris.logger.uri,
+        implementations: ["wrap://ens/js-logger.polywrap.eth"],
+      },
+    ],
```
cbrzn: Can we remove this since they are added by default?

packages/templates/wasm/assemblyscript/yarn.lock
```diff
@@ -0,0 +1,6867 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
cbrzn: I guess this `yarn.lock` was not on purpose? (probably I am wrong tho)

packages/templates/wasm/assemblyscript/src/workflows/e2e.yaml
```diff
@@ -0,0 +1,22 @@
+name: simple-storage
+jobs:
+  cases:
+    steps:
+    - uri: fs/build
+      method: deployContract
+      input:
+        connection: null
```
cbrzn: noob question: do we need to define connection as null here? would it be possible just to ignore it entirely?

packages/js/client/src/__tests__/core/plugin-subinvoke.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+          plugin: mockPlugin(),
+        },
+      ],
+    });
+
+    const result = await client.invoke({
+      uri: "ens/demo.eth",
+      method: "call",
+      input: {
+        input: Uint8Array.from([
```
cbrzn: Would it be possible to have everything as string (or json) and then convert it to `Uint8Array`? This way we understand more what the input is

packages/js/client/src/__tests__/core/plugin-subinvoke.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+describe("plugin-subinvoke", () => {
+  const mockPlugin = () => {
+    class MockPlugin extends PluginModule<Record<string, unknown>> {
+      public async call(
+        input: { input: Uint8Array },
+        client: Client
+      ): Promise<ArrayBuffer> {
+        const res = await client.invoke({
+          uri: "ens/http.polywrap.eth",
```
nerfZael: Since this problem is not specific to HTTP, can we use another plugin to test it? I don't think it's ideal if our tests are making HTTP calls

nerfZael (changes requested):

nerfZael:
> * [x]  allow calling invoke with Uint8Array
> * [x]  use Uint8Array instead of ArrayBuffer for Bytes on Plugin side because msgpack always decode Bytes into ArrayBuffer

doesn't msgpack decode into Uint8Array?

Niraj-Kamdar:
> doesn't msgpack decode into Uint8Array?

That's typo.

packages/js/client/src/plugin/PluginWrapper.ts
```diff
@@ -78,9 +78,7 @@ export class PluginWrapper extends Wrapper {
       await this._sanitizeAndLoadEnv(client, module);
 
       let jsInput: Record<string, unknown>;
-
-      // If the input is a msgpack buffer, deserialize it
-      if (input instanceof ArrayBuffer) {
+      if (input instanceof ArrayBuffer || input instanceof Uint8Array) {
```
nerfZael: What about `if(input instanceof ArrayBufferView)`
Since:
```typescript
  type TypedArray =
        | Uint8Array
        | Uint8ClampedArray
        | Uint16Array
        | Uint32Array
        | Int8Array
        | Int16Array
        | Int32Array
        | BigUint64Array
        | BigInt64Array
        | Float32Array
        | Float64Array;
    type ArrayBufferView = TypedArray | DataView;
```

Niraj-Kamdar: type can't be used as a class, i.e. type don't have any instance

packages/js/client/src/__tests__/core/plugin-subinvoke.spec.ts
```diff
@@ -27,32 +26,43 @@ describe("plugin-subinvoke", () => {
     };
   };
 
+  const isEvenPlugin = () => {
+    class IsEvenPlugin extends PluginModule<Record<string, unknown>> {
+      public isEven(args: {num: number}, client: Client): boolean {
+        return (args.num & 1) ? false : true;
```
cbrzn, line 32:
```suggestion
        return !(args.num & 1);
```

packages/js/client/src/__tests__/core/plugin-subinvoke.spec.ts
```diff
@@ -1,18 +1,17 @@
...
+        args: { input: Uint8Array },
         client: Client
       ): Promise<ArrayBuffer> {
         const res = await client.invoke({
-          uri: "ens/http.polywrap.eth",
-          method: "get",
-          input: input.input,
-          noDecode: true,
+          uri: "ens/is-even.eth",
+          method: "isEven",
```
cbrzn, line 13: Haha this is cool

cbrzn (dismissed):
lgtm :green_apple: 

edit: btw feel free to ignore my suggestion if you prefer the other way

packages/js/client/src/plugin/PluginWrapper.ts
```diff
@@ -78,9 +78,7 @@ export class PluginWrapper extends Wrapper {
       await this._sanitizeAndLoadEnv(client, module);
 
       let jsArgs: Record<string, unknown>;
-
-      // If the args are a msgpack buffer, deserialize it
-      if (args instanceof ArrayBuffer) {
+      if (args instanceof ArrayBuffer || args instanceof Uint8Array) {
```
dOrgJelli: Uint8Array is just one type of `ArrayBufferView`. I propose we change `args instanceof Uint8Array` to `ArrayBuffer.isView(args)`.

packages/js/client/src/wasm/WasmWrapper.ts
```diff
@@ -193,7 +193,10 @@ export class WasmWrapper extends Wrapper {
         invokeResult: {} as InvokeResult,
         method,
         sanitizeEnv: {},
-        args: args instanceof ArrayBuffer ? args : msgpackEncode(args),
+        args:
+          args instanceof ArrayBuffer || args instanceof Uint8Array
```
dOrgJelli, line 197: Same as above, let's use `ArrayBuffer.isView`.

packages/js/core/src/types/Invoke.ts
```diff
@@ -15,7 +15,7 @@ export interface InvokeOptions<
    * Arguments for the method, structured as a map,
    * removing the chance of incorrectly ordering arguments.
    */
-  args?: Record<string, unknown> | ArrayBuffer;
+  args?: Record<string, unknown> | Uint8Array | ArrayBuffer;
```
dOrgJelli: Would change `Uint8Array` to `ArrayBufferView`.

dOrgJelli (changes requested):
In general I think we may be "attacking" the root problem in a wrong way here, let me explain.

## Problem Statement
First, I think we need to be more descriptive of the exact problem we're solving here. I believe it is this...
`schema.graphql`
```graphql
type Module {
  method(arg: Bytes!): Bytes!
}
```

`invoke.ts`
```typescript
// To make it easier for the caller, we should accept any type of arraybuffer
client.invoke({
  method: "method",
  args: {
    arg: ArrayBuffer | Uint8Array | any ArrayBufferView
  }
});

// We should also support passing in pre-encoded arguments
client.invoke({
  method: "method",
  args: argsBuf // ArrayBuffer
});

// We need to differentiate between two types of results:
// (1) A msgpack buffer that's not decoded
// (2) A buffer that's returned from the method
const msgpackBuf = client.invoke({
  method: "method",
  args: { ... },
  noDecode: true
});

const resultBuf = client.invoke({
  method: "method",
  args: { ... },
});
```

Not to mention what happens when sub-invokes are used.

## Solution
I'm not sure, but I don't think this PR addresses all of the cases above. I think we need to look into how we implement support for the `noDecode` option, and see if it covers all cases here.

dOrgJelli:
Hey @Niraj-Kamdar, I implemented a more holistic solution to the problem you're describing here in this PR:
https://github.com/polywrap/monorepo/pull/1000

More details described here: https://discord.com/channels/796821176743362611/908544308876050515/991579335628701696

PR #948, opened 2022-06-23
Chore/remove logger

cbrzn:


cbrzn:
after thinking more about this and talking with @namesty, @nerfZael and @Niraj-Kamdar - we decided to let the interface and plugin (for now) in the monorepo. so will re-open the previous PR again

sorry for the inconvenience

PR #949, opened 2022-06-23
URI resolver test fixes

nerfZael:


packages/js/plugins/uri-resolvers/ipfs-resolver/src/__tests__/helpers/createIpfsClient.ts
```diff
@@ -0,0 +1,5 @@
+import { IpfsClient } from "./IpfsClient";
+
+export const createIpfsClient: (
+  ipfsProvider: string
+) => IpfsClient = require("@dorgjelli-test/ipfs-http-client-lite");
```
cbrzn, line 5: should we move this dependency to `@polywrap/..` ?

cbrzn (approved):
Amazing, thanks for this :pray:

packages/js/plugins/uri-resolvers/ipfs-resolver/src/__tests__/helpers/createIpfsClient.ts
nerfZael: cc @dOrgJelli 

We might want to use the full http-client too, or have a lite and full plugin version

pileks (changes requested):
Was `yarn.lock` updated on purpose?

Other than that, seems good.

packages/js/plugins/uri-resolvers/ipfs-resolver/src/__tests__/helpers/createIpfsClient.ts
dOrgJelli: When using the full version (non-light) we ran into decoding errors which I wasn't able to solve. I was able to solve them however by switching to this package.

The reason for having this package self-published is because of this unmerged PR for the "resolve" feature: https://github.com/ipfs-shipyard/js-ipfs-http-client-lite/pull/5

I think the future plan is the following:
- finalize an IPFS wasm wrapper
- embed wasm IPFS wrapper in client
- have the IPFS URI resolver us the embedded wasm wrapper

Also we could just use the HTTP plugin directly within the IPFS-URI-Resolver since it isn't doing too much w/ IPFS itself except issuing simple HTTP requests.

nerfZael: I wouldn't make the resolver use HTTP directly, but if size and/or functionality is important we can have an IPFS wrapper and IPFS lite wrapper. 
IPFS wrapper implements IPFS interface
IPFS lite wrapper implements IPFS lite interface
IPFS interface implements IPFS lite interface
IPFS resolver uses IPFS lite interface (get implementation)

nerfZael: We can call it something other than `lite`

nerfZael:
> Was `yarn.lock` updated on purpose?
> 
> Other than that, seems good.

Dependencies weren't all locked, I locked them and replaced the deprecated "cids" package as well

dOrgJelli (approved):

PR #951, opened 2022-06-23
Remove Docker Buildx Output Option

dOrgJelli:
closes: https://github.com/polywrap/monorepo/issues/929

Based on the findings [here](https://github.com/polywrap/monorepo/issues/929), the docker buildx `--output` option was not properly implemented. I propose we remove this for now, and add it back when it's necessary.

### DIFF
--- a/packages/cli/src/lib/Compiler.ts
+++ b/packages/cli/src/lib/Compiler.ts
@@ -292,7 +292,6 @@ export class Compiler {
     const useBuildx = !!dockerBuildxConfig;
 
     let cacheDir: string | undefined;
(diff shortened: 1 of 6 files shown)


namesty (dismissed):

fetsorn (commented):
Looks good, i just have one question. Do we want the default exporter to be "docker" or perhaps "image"? I don't know enough about the Compiler to be able to tell.
https://docs.docker.com/engine/reference/commandline/buildx_build/#output

dOrgJelli:
Great point @fetsorn, `image` is what we actually want :)

Issue #907, opened 2022-06-09
Add tests for CLI create command


pileks:
PR https://github.com/polywrap/monorepo/pull/902 fixed a bug where the refactor to Commander introduced issues with the `name` argument in the `create` command.

We should add unit/integration tests to confirm that the CLI commands work as expected in terms of accepting appropriate arguments and produce expected outputs based on the same arguments.

cbrzn:
This task should be converted into unit tests for the entire CLI and not only the create command. I will push this once the alpha sprint is over :-).

More context of my thoughts on this here https://github.com/polywrap/monorepo/pull/959#pullrequestreview-1018552141

dOrgJelli:
Closing this as we now fully test the `create` command, but as @cbrzn has suggested we should create more unit tests for the CLI in general to increase our code coverage.

PR #959, opened 2022-06-24
feat(cli): improve tests for create command

Niraj-Kamdar:
- Closes: #907

### DIFF
--- a/packages/cli/src/__tests__/e2e/create.spec.ts
+++ b/packages/cli/src/__tests__/e2e/create.spec.ts
@@ -2,6 +2,7 @@ import { clearStyle, polywrapCli } from "./utils";
 
 import { runCLI } from "@polywrap/test-env-js";
 import rimraf from "rimraf";
+import { ProjectType, supportedLangs } from "../../commands";
 
 const HELP = `Usage: polywrap create|c [options] [command]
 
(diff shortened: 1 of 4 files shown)


cbrzn (commented):
Hey Niraj thanks for start a tackle on this :pray:

I was thinking that this is not the approach we want to take if we want to close #907 - I think instead we should create a new file in `__tests__/unit/` folder named `create.spec.ts` and basically try to run the functions from there and check if the behavior is the expected.

For example, I thought that we can follow these approaches:
https://github.com/tj/commander.js/issues/438#issuecomment-274285003
https://github.com/tj/commander.js/issues/1565#issuecomment-879661279
and execute every method of instance?

Another approach we can take is to (talking from the context of `create` tests) execute the `run` function and see if it indeed creates a folder for example

Note that the objective of these approaches is to be able to take up our code coverage, as you mentioned before :-D

Niraj-Kamdar:
@cbrzn I'd still keep e2e tests, we just need to add unit tests.

cbrzn:
> @cbrzn I'd still keep e2e tests, we just need to add unit tests.

That's correct! I would remove the `closes ...` from the description and just make this an improvement of the current integration tests. As we talked before, let's tackle the unit tests post alpha

dOrgJelli (approved):
This is really nice @Niraj-Kamdar :clap:

dOrgJelli:
Also IMO @cbrzn @Niraj-Kamdar I think this fully tests the `create` functionality, so I'm not sure we need unit tests if the e2e tests can exercise all functionality.

PR #967, opened 2022-06-24
Env refactor

namesty:
## Overview

This PR aims to change the way environment variables/configuration is passed to wrappers. This was done with 5 main objectives in mind:

- Change the wasm-boundary functions to follow from a a "pull" model, instead of a "push" model
- Remove `ClientEnv` and `sanitizeEnv` concept
- Support environment configuration for Rust wrappers
- Enable/pass environment at the method level
- Prepare implementation for, in the future, overriding environment properties at the method level

Over the course of the development and discussion process, we realized there were additional tasks to implement to be feature complete. So this PR also enables:

- Support for importing external environment types
- Support for conditional environment configs
- Support for environment configuration for subinvoked wrappers
- E2E tests for both subinvocation with env configurations, and Rust wrappers with env configurations.

## Developer experience changes

We're introducing the `@env` directive, which annotates methods that use an environment configuration:

`schema.graphql:`

```graphql
type Env {
  str: String
  requiredInt: Int!
}

type Module {
  getEnv(arg: String!): Env @env(required: false)
}
```

Additionally, now module methods take a second `env` argument, from where the wrapper developer can access the environment variables:

`index.ts:`

```typescript
import { Env, Input_getEnv } from "./wrap";

export function getEnv(input: Input_getEnv, env: Env | null): Env | null {
  return env;
}

```


dependencies/package.json
```diff
@@ -1,8 +0,0 @@
-{
```
dOrgJelli: why did we remove this?

dOrgJelli (approved):

packages/js/core/src/types/Plugin.ts
```diff
@@ -34,25 +33,10 @@ export abstract class PluginModule<
     return this._config;
   }
 
-  public _wrap_load_env(env: TEnv): void {
+  public setEnv(env: TEnv): void {
```
Niraj-Kamdar, line 36: Why rename this?

Niraj-Kamdar (approved):

packages/js/core/src/types/Plugin.ts
dOrgJelli: It's because `_wrap_load_env` used to replicate the export that was coming from WRAP Wasm modules. Now that we've changed the way we load environment in wasm, this export is no longer present, so having it here is a bit confusing.

Instead we can simply add a function to set it on the plugin instance.

In order to make this 1:1 with wasm's usage pattern, we'd be passing in the environment to the plugin functions we call, but this is actually a bit more complicated to implement (requires custom codegen) so this setter will do for now.

Issue #982, opened 2022-06-28
PolywrapClient tests - stop relying on simple storage wrapper and ens/ipfs


nerfZael:
```
core 
  resolveUri.spec.ts
    - can resolve an ENS wrapper
    - can resolve previously cached IPFS URI after redirecting from ENS
    - can resolve wrapper an ENS wrapper whose IPFS URI has been redirected
  wasm-wrapper.spec.ts
    - subscribe
    - subscription early stop
e2e
  wasm-as.spec.ts
    - asyncify
    - simple-storage
  wasm-rs.ts
    - asyncify
    - simple-storage
  workflow.spec.ts
  ```
  - [x] #993
  - [x] #994
  - [x] #995
  - [x] #996
  - [x] #997

PR #983, opened 2022-06-28
Migrating away from test-env in client-js tests

nerfZael:
- Removed the use of `simple storage` wrapper in favor of `simple` wrapper from most test cases in the client
- Calls made with `client.query` that call a single method have been replaced with `client.invoke` 

Issue to track all remaining uses of ens/ipfs in client tests: #982


packages/js/client/src/__tests__/core/plugin-wrapper.spec.ts
```diff
@@ -122,36 +112,28 @@ describe("plugin-wrapper", () => {
       ],
     });
 
-    const queryEnv = await client.query({
+    const queryEnv = await client.invoke({
```
cbrzn: Should we also remove the `query` and `mutation` from variable names in tests? (No strong opinion here)

packages/js/client/src/__tests__/core/resolveUri.spec.ts
```diff
@@ -708,50 +678,28 @@ describe("resolveUri", () => {
...
     ]);
   });
 
-  it("can resolve wrapper with redirects", async () => {
-    await runCLI({
-      args: ["build"],
-      cwd: `${GetPathToTestWrappers()}/wasm-as/interface-invoke/test-interface`,
-    });
-
+  it("can resolve wrapper an ENS wrapper whose IPFS URI has been redirected", async () => {
```
cbrzn: Is this well written?

packages/js/client/src/__tests__/e2e/test-cases.ts
```diff
@@ -299,348 +219,257 @@ export const runBigNumberTypeTest = async (
...
+    method: "method1",
+    args: {
+      en: 1,
+      optEnum: "INVALID",
+    },
   });
 
-  expect(method1c.errors).toBeTruthy();
+  expect(method1c.error).toBeTruthy();
   // @ts-ignore
```
cbrzn: We still need the ts-ignore?

packages/test-cases/cases/wrappers/wasm-as/simple-calculator/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "test-case-simple-calculator",
+  "private": true,
+  "dependencies": {
+    "@polywrap/wasm-as": "0.0.1-prealpha.91",
```
cbrzn:
```suggestion
    "@polywrap/wasm-as": "0.1.0",
```

packages/test-cases/cases/wrappers/wasm-as/simple/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "test-case-simple",
+  "private": true,
+  "dependencies": {
+    "@polywrap/wasm-as": "0.0.1-prealpha.58",
```
cbrzn:
```suggestion
    "@polywrap/wasm-as": "0.1.0",
```

cbrzn (commented):
This is awesome :-D

packages/js/client/src/__tests__/core/plugin-wrapper.spec.ts
nerfZael: yes

packages/js/client/src/__tests__/e2e/test-cases.ts
nerfZael: nope

packages/js/client/src/__tests__/core/resolveUri.spec.ts
nerfZael: Kinda, I'll tackle this with a PR for "remove test-env from resolveUri tests"

packages/js/client/src/__tests__/core/resolveUri.spec.ts
```diff
@@ -1,57 +1,34 @@
...
+const wrapperUri = new Uri(`wrap://file/${wrapperPath}/build`);
 
-  beforeAll(async () => {
-    await initTestEnvironment();
-    ipfsProvider = providers.ipfs;
-    ethProvider = providers.ethereum;
-  });
+const simpleFsResolverWrapperPath = `${GetPathToTestWrappers()}/wasm-as/simple-fs-resolver`;
+const simpleFsResolverWrapperUri = new Uri(
+  `wrap://file/${simpleFsResolverWrapperPath}/build`
```
cbrzn, line 18: Just curious here: Is there any reason to use `file` instead of `fs`? I think that indeed we should change it from `fs` to `file` but probably we should keep consistency for now?

dOrgJelli: I think we should slowly start to faze out `fs` and instead use the standard `file://` scheme name.

dOrgJelli (approved):

packages/js/client/src/__tests__/core/resolveUri.spec.ts
nerfZael: > Just curious here: Is there any reason to use `file` instead of `fs`? I think that indeed we should change it from `fs` to `file` but probably we should keep consistency for now?

`file` is just more commonly used in file URL outside of polywrap. E.g. in the browser, or package.json when pointing to local packages

cbrzn: > I think we should slowly start to faze out fs and instead use the standard file:// scheme name.

I would suggest opening a PR and changing this in every part of the codebase (ASAP). We know that `fs` and `file` are equivalent in our codebase, but this might cause some confusion to people that check our codebase IMO

PR #989, opened 2022-06-28
feat(cli): add job status in job result object

Niraj-Kamdar:
- Closes: #974


krisbitney:
This is really nice! I don't know if you saw, but I made a related issue yesterday https://github.com/polywrap/monorepo/issues/984

What do you think of adding a similar a similar JobResult enum for validation results? If we parse the validation script and output, we can run one validation task at a time to test for pass/fail for individual tasks.

Niraj-Kamdar:
@krisbitney Thanks for creating the issue, I believe some of the issues should already be solved in the current PR but will solve the remaining ones.

krisbitney (dismissed):
LGTM

dOrgJelli (approved):

PR #999, opened 2022-06-29
Core: Wrap package refactor

cbrzn:
# Core
## Top-Level Primitives
- Client
- Uri
- Wrapper

## Complementary Primitives
Client:
- Client Methods
- Client Extensions (Interface Impls)
- Client Configuration

URI:
- URI Resolution
- URI Resolvers

## Secondary Primitives
Invocation:
- Invoker
- Invocable

Workflow

MsgPack


packages/js/core/src/invocation/Invoker.ts
```diff
@@ -0,0 +1,21 @@
+// Why do we need 2 types of "invoke" interfaces?
+// Anser: options & return values are different
```
cbrzn, line 2: Options & return values -> input & output

dOrgJelli:
NOTE: this PR is not intended to be merged, and instead was a bunch of half-done refactoring that will be completed in other PRs.

PR #1002, opened 2022-06-30
JS/Core: Msgpack tests implemented

cbrzn:


packages/js/core/src/__tests__/msgpack.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    expect(decoded).toEqual(customObject);
+  });
+
+  it("Should encode and decode, returning the same map", () => {
+    const customMap = new Map()
+    customMap.set("firstKey", "firstValue")
+    customMap.set("secondKey", "secondValue")
+
+    const encoded = msgpackEncode(customMap)
+    // [199, 43, 1] are being added because of the map structure
```
cbrzn, line 29: can someone please confirm this - I added this because that's I thought it work that way, but if you think it's misleading we can simply remove the comment :-)

dOrgJelli (approved):

PR #1013, opened 2022-07-05
feat: add simple wrapper to test workflows

Niraj-Kamdar:
- Relates: #983
- Closes: #997


packages/js/client/src/__tests__/e2e/workflow-test-cases.ts
```diff
@@ -1,81 +1,189 @@
...
-  jobs: {
-    cases: {
-      steps: [
-        {
-          uri: "ens/testnet/simple-storage.eth",
-          method: "deployContract",
-          args: {
-            connection: null,
-          },
+export interface WorkflowTestCase {
```
nerfZael, line 5: Why do we have test cases as a separate file if we're only using them in one test file?

packages/js/client/src/__tests__/e2e/workflow.spec.ts
```diff
@@ -1,75 +1,39 @@
...
+  let client: PolywrapClient;
+
   beforeAll(async () => {
     await initTestEnvironment();
+
+    await buildWrapper(
+      path.join(GetPathToTestWrappers(), "wasm-as", "simple-calculator")
+    );
+
+    client = await getClientWithEnsAndIpfs();
```
nerfZael: You can just use getClient

packages/js/client/src/__tests__/e2e/workflow.spec.ts
```diff
@@ -1,75 +1,39 @@
...
+import { testCases } from "./workflow-test-cases";
+import path from "path";
 
 jest.setTimeout(200000);
 
 describe("workflow", () => {
+  let client: PolywrapClient;
+
   beforeAll(async () => {
     await initTestEnvironment();
```
nerfZael: No need for test-env, that was the point of all this :D

nerfZael (changes requested):

packages/js/client/src/__tests__/e2e/workflow-test-cases.ts
Niraj-Kamdar: Just for more readability

packages/test-cases/cases/wrappers/wasm-as/simple-calculator/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "test-case-simple-calculator",
+  "private": true,
+  "dependencies": {
+    "@polywrap/wasm-as": "0.0.1-prealpha.91",
```
cbrzn, line 5:
```suggestion
    "@polywrap/wasm-as": "0.1.0",
```

cbrzn (approved):
This is an amazing example that I will use to implement workflows in the integrations repository. Thanks for this :)

Niraj-Kamdar:
@cbrzn You mean demo right?

PR #1015, opened 2022-07-05
Remove simplestorage dependency from asyncify client tests

namesty:
Created a test-side plugin to serve as an in-memory storage for the asyncify test cases in the client's e2e test suites.


packages/test-cases/cases/wrappers/plugins/memory-storage/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@polywrap/storage-plugin-js",
```
dOrgJelli: Since this package.json isn't a "private" package, it will be published, which I'm not sure we want.

Can we allow this plugin to still be in the test-cases, get built, but not be published? Another idea is to restructure the asyncify tests so that we don't need this plugin, but it might be necessary.

dOrgJelli (changes requested):

packages/test-cases/cases/wrappers/plugins/memory-storage/package.json
nerfZael: @namesty  What about just putting the plugin in the client test, it doesn't need to be a whole package, it could just be similar to the mock plugins, either defined inline or maybe a separate file since it's a bit bigger.
I don't know why we're sticking all the wrappers and plugins in the test-cases project. Yea, sure it makes the examples reusable but a lot of them aren't reused.

namesty: Yeah, I ended up going for the in-memory plugin approach, declared as a test util in the client's test directory. Just re-tested it to make sure it all passes.

packages/js/client/src/__tests__/e2e/memory-storage.ts
```diff
@@ -0,0 +1,33 @@
+import { Client, PluginFactory, PluginModule } from "@polywrap/core-js";
+
+type NoConfig = Record<string, never>;
+
+function sleep(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
```
nerfZael, line 7: memory-storage-plugin.ts for file name would be a bit better IMo

nerfZael (approved):

dOrgJelli (approved):
LGTM!

PR #1016, opened 2022-07-05
feat: improve cli log for the workflow command 

Niraj-Kamdar:


packages/cli/src/__tests__/e2e/utils.ts
```diff
@@ -7,20 +7,58 @@ export const clearStyle = (styled: string) => {
 
 export const parseOutput = (
   outputs: string
-): Array<{ id: string; data?: unknown, error?: unknown }> => {
```
dOrgJelli, line 10: Seems like some useful "CLI parsing" functionality we could break-off into a wrapper in the future?

dOrgJelli (dismissed):

krisbitney:
This is great. Thanks so much for adding this! 

I have a suggestion that could be a future PR. Instead of the `quiet` option, in the future we could add a `verbose` parameter. Some packages have a `verbose` option to vaguely indicate how much output to expect. Valid values could be like 0, 1, or 2.
* 0 -> quiet
* 1 - > minimal output. For workflows, I'm thinking it would print the job status and job id for each job, but suppress the output values
* 2 -> full output

The reason this is valuable is that sometimes the job output is a complex object that fills up the terminal. Depending on how people are using the workflows, it can make sense to have something in between printing full output and no output.

krisbitney (dismissed):

Niraj-Kamdar:
@krisbitney  I agree about verbosity level and I believe this shouldn't be limited to the workflow but applied to CLI as a whole and we need a holistic plan for implementing it. I think this should be done post origin.

packages/cli/src/__tests__/e2e/utils.ts
Niraj-Kamdar: Yeah, we could but do we need it other than e2e testing? I believe, For programmability, we should always provide better APIs in the implementation language instead of devs needing to interact with CLI and Shell.

krisbitney:
> @krisbitney I agree about verbosity level and I believe this shouldn't be limited to the workflow but applied to CLI as a whole and we need a holistic plan for implementing it. I think this should be done post origin.

That sounds like a great plan

dOrgJelli (approved):

PR #1034, opened 2022-07-07
feat(manifests): add json schema for abi

Niraj-Kamdar:


packages/js/manifests/wrap/src/formats/wrap.info/deserialize.ts
```diff
@@ -29,7 +29,7 @@ export function deserializeWrapManifest(
   }
 
   if (!options || !options.noValidate) {
```
cbrzn, line 31: Should this be && instead of ||?

packages/js/manifests/wrap/src/formats/wrap.info/index.ts
```diff
@@ -6,26 +6,37 @@
...
+  WrapAbi_0_1_0_0_0_1
 };
 
 export enum WrapManifestVersions {
   "0.1.0" = "0.1.0",
-}
+};
+
+export enum WrapAbiVersions {
+  "0.1.0" = "0.0.1",
```
cbrzn: What's the reason for making the value different? Would it make sense to -for now- keep consistency between these?

packages/templates/wasm/assemblyscript/src/__tests__/e2e/integration.spec.ts
```diff
@@ -18,6 +18,8 @@ describe("Template Wrapper End to End Tests", () => {
   it("calls sampleMethod", async () => {
     const expected: string = "polywrap";
 
+    console.log(JSON.stringify(await client.getManifest(wrapperUri, {})));
```
cbrzn:
```suggestion
```

cbrzn (commented):
This is amazing Niraj. Thanks for pushing this :pray:

packages/js/manifests/wrap/src/formats/wrap.info/deserialize.ts
Niraj-Kamdar: Nope it should be or basically it is if (options is falsy or options.novalidate is falsy)

packages/js/manifests/wrap/src/formats/wrap.info/index.ts
Niraj-Kamdar: I don't know it was like this in your original commit

packages/templates/wasm/assemblyscript/src/__tests__/e2e/integration.spec.ts
Niraj-Kamdar: Nice catch!

packages/js/manifests/wrap/src/formats/wrap.info/index.ts
cbrzn: okkk my bad then - since this is an autogenerated file I think that we need to rename the `abi` format to `0.1.0`

Niraj-Kamdar: I'll update it then.

packages/cli/src/lib/Compiler.ts
```diff
@@ -369,7 +369,7 @@ export class Compiler {
       });
 
       const info: WrapManifest = {
-        abi: filteredAbi,
+        abi: (filteredAbi as unknown) as WrapAbi,
```
dOrgJelli, line 376: This is a bit sketchy, and points out that we have 2 ABI types defined in `schema/parse/abi` and `manifest/wrap.info/abi`.

I looked into using the `manifest/wrap.info/abi` inside of the schema pipeline directly, but it is a bit complicated... I think we should create a new PR after this is merged to consolidate these definitions and only use the manifest's ABI.

dOrgJelli (dismissed):

packages/cli/src/lib/Compiler.ts
dOrgJelli: I added a final validation pass below this to ensure we don't write invalid ABIs.

dOrgJelli (approved):

cbrzn:
> I looked into using the manifest/wrap.info/abi inside of the schema pipeline directly, but it is a bit complicated... I think we should create a new PR after this is merged to consolidate these definitions and only use the manifest's ABI.

I will push this forward

PR #1035, opened 2022-07-07
Chore: Update meta manifest

cbrzn:
This PR aims to remove the `queries` key in the `polywrap.meta.yaml` manifest.


Niraj-Kamdar:
What's the purpose of having this now that we don't have hub?

dOrgJelli:
Yah I'm leaning towards @Niraj-Kamdar's perspective. I think we should re-add this feature when we have a better idea of how it should work. For now let's just remove "queries".

dOrgJelli (approved):

PR #1036, opened 2022-07-07
Allow custom cache object to be passed to client config

namesty:


dOrgJelli (dismissed):

packages/js/core/src/types/Client.ts
```diff
@@ -20,6 +21,7 @@ export interface ClientConfig<TUri extends Uri | string = string> {
   interfaces: InterfaceImplementations<TUri>[];
   envs: Env<TUri>[];
   uriResolvers: UriResolver[];
+  wrapperCache?: WrapperCache;
```
krisbitney, line 24: Thoughts on naming this "cache"?

krisbitney (dismissed):

packages/js/core/src/types/Client.ts
krisbitney: Everything LGTM!

dOrgJelli: I like the explicit nature of "wrapper cache", as there may be other things we want to cache in the future.

PR #1046, opened 2022-07-13
Feat: Post origin build artifact

cbrzn:
***Warning: This PR is being blocked by #1073***

# Post origin build artifact

#### Objective

Removal of schema.graphql in wrappers (wasm, plugins) & interfaces built artifact

### Breaking changes

This introduces breaking changes that makes old wrappers not compatible with the new ones built with this new artifact. The reason is that currently, when a wrapper is imported, it tries to resolve a `schema.graphql` which is a string that needs to be converted to ABI.

- Plugins will have `wrap.info` (encoded w/msgpack) inside of `build` folder and `wrap.info.ts` inside of the `wrap` folder. Also, the `getManifest` method will be implemented in `PluginWrapper`, allowing us to access to this manifest
- Polywrap (WASM & Interface), plugins, and app manifest have been updated to version `2.0`, being the main change in the attribute `import_redirects`, in order to support `info` instead of `schema`, allowing developers to point to `wrap.info` insteaf of the `schema.graphql`
- Previously, we had the type `PluginPackageManifest` for plugins manifests
    ```typescript=
    /** The plugin package's manifest */
    export interface PluginPackageManifest {
      /** The Wrapper's schema */
      schema: string;

      /** All interface schemas implemented by this plugin. */
      implements: Uri[];
    }

    export type PluginPackage<TConfig> = {
      factory: () => PluginModule<TConfig>;
      manifest: PluginPackageManifest;
    };
    ```

    This has been changed to:

    ```typescript=
    export type PluginPackage<TConfig> = {
      factory: () => PluginModule<TConfig>;
      manifest: WrapManifest;
    };
    ```

### Other important changes

- `schema.ts` has been removed from app codegen
- `schema/composer` package doesn't output the schema anymore, but instead, only the ABI

### Hotfixes

Because of the circular dependency in our monorepo, a few hacks needed to be done in order to have a functional build process of the libraries:

- Create `wrap-man` with new structure inside of every plugin and update the files to point to this manually generated folder
- Create `build-man` for interfaces so they have the `wrap.info` file

hack.md of this description: https://hackmd.io/@cbrzn/Bki01E32q


dOrgJelli (approved):

dOrgJelli:
`wrap.info` -> `schema.graphql`

PR #1051, opened 2022-07-16
Feat: CLI `polywrap codegen` Command

### LINKED ISSUE #1153: Feat: CLI polywrap codegen command modification
TODO:
- move `plugin codegen` and `app codegen` to just `codegen`
- move `--codegen-dir`, `--publish-dir`, and `--client-config` into manifest file? probably good for a new PR (needs discussion)

dOrgJelli:
I think the CLI interface should be:
```
polywrap codegen
polywrap build
polywrap deploy
polywrap infra
polywrap create
polywrap run
```

Where each command decides internally what projects it works with.

This PR consolidates the `polywrap plugin codegen` and `polywrap app codegen` commands into a single `polywrap codegen` command. Additionally it refines best-practices about what options should be defined in polywrap manifest files and which are provided on the CLI.

NOTE: see the `todo` file for WIP impl notes

TODO:
- [ ] option to override any property
- [ ] option to specify a override fragment yaml file

closes: #1153


pileks:
`polywrap codegen` is now the **only** `codegen` command in the CLI right now. 😄 

@dorgjelli it seems that `codegen` outputs into `./src/wrap` for all project types currently, so I have kept it at that.

Work that should be split into issues and done separately:

- [x] `codegen` for plugins appears to also output the plugins ABI into a `./build` folder. We should refactor this out into the `polywrap build` command, and extend the `build` command to support plugins as well.
  - This would also move the `--publish-dir` option that the plugin `codegen` uses into `build`'s `--output-dir` option
  - `codegen` would then support `wasm`, `plugin` and `app` project types, while `build` would support `wasm`, `plugin` and `interface` types
- [ ] move `--codegen-dir`, `--publish-dir`, and `--client-config` into manifest files. Separate tasks for each of those might seem fair.
    - [ ] `--codegen-dir` issue created
    - [ ] `--publish-dir` issue created
    - [ ] `--client-config` issue created

pileks:
> * [ ]  move `--codegen-dir`, `--publish-dir`, and `--client-config` into manifest files. Separate tasks for each of those might seem fair.
>   
>   * [ ]  `--codegen-dir` issue created
>   * [ ]  `--publish-dir` issue created
>   * [ ]  `--client-config` issue created

Note: we should probably rethink our manifest migration strategy before messing with manifests. While we can have all of these options set as optional manifest properties, I'd say that we'll likely want to create a proper migration strategy sooner than later, as it will reduce future work.

Current migration strategy:
Auto-generated code requires us to implement certain, fixed migrators.
```
available manifests:
0.1.0
0.2.0
0.3.0
0.4.0

migrators/migration paths:
0.1.0 -> 0.4.0
0.2.0 -> 0.4.0
0.3.0 -> 0.4.0
```

Proposed migration strategy:
Allow developers to declare their own migrators, from whichever version to whichever other version.
A simple search algorithm would then find the appropriate migration path, and could even allow for downgrades if the correct migrator exists.

```
available manifests:
0.1.0
0.2.0
0.3.0

migrators:
0.1.0 -> 0.2.0
0.2.0 -> 0.3.0
0.3.0 -> 0.4.0
0.1.0 -> 0.3.0

migration paths to 0.4.0:
0.1.0 -> 0.2.0 -> 0.3.0 -> 0.4.0
0.1.0 -> 0.3.0 -> 0.4.0
```

Related issue with further explanation:
https://github.com/polywrap/toolchain/issues/1211

pileks:
Added issue for the splitting of plugin `codegen` behavior:
https://github.com/polywrap/toolchain/issues/1210

packages/cli/src/lib/project/helpers/getProjectFromManifest.ts
```diff
@@ -0,0 +1,197 @@
...
+type PolywrapProjectType = "wasm/interface" | "plugin" | "app" | undefined;
+
+export async function getProjectFromManifest(
+  manifestFile: string
+): Promise<Project<AnyProjectManifest> | undefined> {
+  const manifest = filesystem.read(manifestFile) as string;
+
+  const format = getManifestFormat(manifest);
+
+  switch (format) {
```
krisbitney: I think the methods in this file are too prone to breaking

packages/interfaces/file-system/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@polywrap/file-system-interface",
   "description": "Polywrap Filesystem Interface",
-  "version": "0.6.0",
+  "version": "0.4.1",
```
krisbitney: Are the interfaces breaking at version 0.6.0?

krisbitney (changes requested):
This PR is awesome. I think it's a very positive change. Thanks for working on this!

I'm concerned about the maintainability of `getProjectFromManifest.ts`. I refactored some it in another branch so you can see what I'm thinking. I messaged you.

packages/cli/src/lib/project/helpers/getProjectFromManifest.ts
pileks: Can agree.
As an aside, we've decided to go with your refactor, as it's something that I've also done before going with this complicated approach. If 2 devs blindly do it, it's surely worth something 😁

packages/interfaces/file-system/package.json
pileks: This is just a happy little accident 😄

krisbitney (dismissed):

krisbitney:
The dependencies in the other interfaces were also changed to 0.4.1

krisbitney (approved):

dOrgJelli (commented):
Awesome work @pileks!

PR #1054, opened 2022-07-20
Rust build image optimizations

namesty:
This PR aims to improve the time it takes to build a Rust wrapper with Polywrap's CLI, by optimizing Docker's build image. 

This is done, mainly, by copying over the Cargo manifests, building dependencies, and then copying the actual source code and rebuilding. With this trick, the copying of the source files doesn't invalidate Docker's cache at the dependencies building layer. Thus saving time when changing the source code and rebuilding (which is the most common iteration when developing a Rust wrapper).

I've also added a test to measure build times, comparing the previous image with the new one by measuring: the time it takes to build each image for the first time, the time it takes to build them again (with the image cached) and the time they take to build after modying the source code in each. I've set this test to be skipped by default as it would heavily increase the CI and overall testing times.

Below is a chart with the average results I've got:

| Case| Current (seconds) | New (seconds) |
|---|---|---|
| First build (no cache) | 189.77 | 220.26 |
| Second build (with cache - no source modification) | 71.90 | 37.22 |
| Third build (after source modification) | 73.31 | 36.99 |

In general, the first build takes a bit more, subsequent builds take less.

### DIFF
--- a/packages/js/client/src/__tests__/e2e/wasm-rs.spec.ts
+++ b/packages/js/client/src/__tests__/e2e/wasm-rs.spec.ts
@@ -336,8 +336,8 @@ describe("wasm-rs test cases", () => {
   })
 });
 
-describe.only("Wasm-rs benchmarking", () => {
-  const wrapperPath = `${GetPathToTestWrappers()}/wasm-rs/bignumber-type`
+describe.skip("Wasm-rs benchmarking", () => {
+  const wrapperPath = `${GetPathToTestWrappers()}/wasm-rs/benchmarks`
   const wrapperUri = `fs/${wrapperPath}/build`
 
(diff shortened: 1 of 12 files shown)


packages/js/client/src/__tests__/e2e/wasm-rs.spec.ts
```diff
@@ -330,3 +335,95 @@ describe("wasm-rs test cases", () => {
     );
   })
 });
+
+describe.skip("Wasm-rs benchmarking", () => {
```
dOrgJelli, line 339: We can merge this as `skip`ped for now, but we should probably make a benchmarking project that does this in the near future.

dOrgJelli (approved):
:fire:

PR #1056, opened 2022-07-20
Remove test env from resolve uri tests

nerfZael:


packages/js/client/src/__tests__/e2e/test-cases.ts
```diff
@@ -5,28 +5,25 @@ export const runAsyncifyTest = async (
   client: PolywrapClient,
   wrapperUri: string
 ) => {
-  const subsequentInvokes = await client.query<{
+  const subsequentInvokes = await client.invoke<{
     subsequentInvokes: string;
   }>({
```
cbrzn:
```suggestion
  const subsequentInvokes = await client.invoke<string>({
```

packages/js/client/src/__tests__/e2e/test-cases.ts
```diff
@@ -845,128 +844,58 @@ export const runSimpleStorageTest = async (
...
+  expect(getDataResult.error).toBeFalsy();
+  expect(getDataResult.data).toEqual(55);
 };
 
 export const runSimpleEnvTest = async (
   client: PolywrapClient,
   wrapperUri: string
 ) => {
-  const queryGetEnv = await client.query({
+  const queryGetEnv = await client.invoke({
```
cbrzn: `invokeGetEnv` instead of `queryGetEnv` maybe?

cbrzn: I am seeing that this is done in multiple parts, so I am not commenting on all of those but rather making a suggestion here to change it in every part `query` is mentioned :P

packages/test-cases/cases/wrappers/wasm-as/simple-fs-resolver/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "test-case-simple-fs-resolver",
+  "private": true,
+  "dependencies": {
+    "@polywrap/wasm-as": "0.1.0",
```
cbrzn:
```suggestion
    "@polywrap/wasm-as": "0.2.0",
```

packages/test-cases/cases/wrappers/wasm-as/simple-redirect-resolver/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "test-case-simple-redirect-resolver",
+  "private": true,
+  "dependencies": {
+    "@polywrap/wasm-as": "0.1.0",
```
cbrzn:
```suggestion
    "@polywrap/wasm-as": "0.2.0",
```

packages/js/client/src/__tests__/e2e/test-cases.ts
nerfZael: I prefer `getEnvResult`

PR #1060, opened 2022-07-21
IPFS Plugin - move config into envs + method options

### LINKED ISSUE #717: Implement comprehensive IPFS URI resolver configuration options
Revisit resolution options (provider, parallel execution, etc...) for IFPS URI resolver and implement them as env variables

Before this task can be tackled, the IPFS part of the following issue needs to be implemented: https://github.com/polywrap/monorepo/issues/622

pileks:
Want to complete before merge:

- [x] Tests for timeouts
  - [x] IPFS
  - [x] IPFS Resolver
- [x] Determine what to do regarding IPFS Interface `fallbackProviders`
  - Should we have `fallbackProviders` be part of the IPFS interface specification? Seems like the answer would be “no”, because this would mean we are assuming that all implementations of the IPFS plugin should use fallback providers.
  - After discussing with @nerfZael we have decided to, for now, move `fallbackProviders` into the `Options` type of the Ipfs interface
- [x] Determine what to do regarding the naming of `skipCheckIfExistsBeforeGetFile` in IPFS Resolver Plugin env
  - I’m not a big fan of how I named this, though I can’t come up with a better name… any suggestion is welcome!
- [ ] See if we would maybe want to switch to a `providers` system instead of having a `provider` and `fallbackProviders`
  - Probably not part of this PR, but wanted to see people’s thoughts.
- [x] Should we include `Env` in IPFS Interface schema?


packages/js/plugins/uri-resolvers/ipfs-resolver/src/index.ts
```diff
@@ -40,7 +40,8 @@ export class IpfsResolverPlugin extends Module<NoConfig> {
         {
           cid: `${args.path}/${manifestSearchPattern}`,
           options: {
-            timeout: 5000,
```
krisbitney, line 43: By removing the default timeout values, it looks like the timeouts will default to 0. I don't know much about IPFS timeouts. Is 0 unlimited timeout? Is this a good default value? Or should we keep default timeout values somewhere? For example, we could have default timeouts in the ipfs uri resolver like `this.env.timeouts?.tryResolveUri ?? 5000`, or we could add a default `env` in the default config. What do you think?

packages/js/plugins/uri-resolvers/ipfs-resolver/src/schema.graphql
```diff
@@ -1,5 +1,26 @@
...
+  getFile: UInt32,
+  checkIfExists: UInt32,
+  tryResolveUri: UInt32,
+}
+
+type Env {
+    """
+    Determines whether the plugin should try to resolve a file (check for its existence) or immediately get the file
+    """
+    skipCheckIfExistsBeforeGetFile: Boolean,
```
krisbitney: **Brainstorming other names**
skipCheckIfExists
skipFileCheck
skipCheckBeforeGetFile

I think the name you have already is long but clear and informative. If you decide to change it, my favorite alternative is `skipCheckIfExists` because it matches the timeout name.

krisbitney (dismissed):
LGTM! Please review the comments before merging.

krisbitney (dismissed):

packages/js/plugins/uri-resolvers/ipfs-resolver/src/schema.graphql
pileks: I think `skipCheckIfExists` sounds best too! Will go with that one.

packages/js/plugins/uri-resolvers/ipfs-resolver/src/index.ts
pileks: Yes, a zero is an unlimited timeout. We could go with something simple like a single `defaultTimeout` which we could use for all actions? Seems like an okay solution, and we don't necessarily need to have a full default local `env` (especially since we wouldn't be defining all `env` defaults, only certain fields).

krisbitney: That makes sense to me. I don't know the best way to handle this. I just wanted to bring up the question. I think the default behavior should be that it "just works" without having to configure timeouts, and timeout configuration should be optional. If unlimited timeouts won't cause any issues (like an app hanging, waiting for a response that will never come) then that would be fine.

pileks:
>  [x] Should we include Env in IPFS Interface schema?

Including Env in the interface would assume too much regarding the implementation itself - IMO we should not include it in the interface schema.

krisbitney (approved):

dOrgJelli (approved):

PR #1067, opened 2022-07-22
fix: ethereum plugin now tries to obtain signer from provider if signer.connect fails

krisbitney:
Signers produced by some wallets, such as Metamask, do not support changing providers. Ethers can throw an exception when calling `signer.connect`:
> Sub-classes must implement this, however they may simply throw an error if changing providers is not supported.

This creates a situation where explicitly providing a signer to the Ethereum plugin config can cause unexpected issues:
```typescript
// this will throw an exception and fail when changing networks
mainnet:  {
  provider: library,
  signer: library.getSigner()
 },
 
 // this will work without issue
mainnet:  {
  provider: library,
 },
```

This is preventable by allowing the provider to provide a signer if signer.connect fails.

I tested this by using yarn link in a react app.

After implementing this fix, I wonder if it makes more sense to instead leave the responsibility of handling this correctly to the user. Thoughts?

### DIFF
--- a/packages/js/plugins/ethereum/src/Connection.ts
+++ b/packages/js/plugins/ethereum/src/Connection.ts
@@ -134,7 +134,11 @@ export class Connection {
       try {
         this._config.signer = signer.connect(this._client);
       } catch (e) {
-        this._config.signer = this._client.getSigner();
+        if (this._client.getSigner) {
+          this._config.signer = this._client.getSigner();
+        } else {
+          throw e;
+        }
       }
     } else {
       this._config.signer = signer;
--- a/packages/js/plugins/ethereum/src/Connection.ts
+++ b/packages/js/plugins/ethereum/src/Connection.ts
@@ -114,6 +114,7 @@ export class Connection {
     return this._client;
   }
 
+  // If setSigner fails, it will attempt to retrieve a signer from the given provider
   public setSigner(signer: EthereumSigner): void {
     if (typeof signer === "string") {
       this._config.signer = getAddress(signer);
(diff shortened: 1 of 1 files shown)


dOrgJelli (changes requested):
Yah I'm not sure about this...

The case that occurs here that I feel uncomfortable with is where a signer is provided, fails to connect to the provider, and then the signer provided by the user is "forgotten" and replaced by the signer setup within the provider.

I think it is actually most correct to let the user configure it correctly. If the signer doesn't support re-configuring the provider, then it should be setup with the correct provider before being sent to the plugin.

Also you can just provide a signer w/o the provider, and the plugin will get the provider from the signer.

So the 3 supported cases are:
- provider only (get signer from provider)
- signer only (get provider from signer)
- provider + signer (signer must have same provider)

Does this make sense to you?

krisbitney:
> Yah I'm not sure about this...
> 
> The case that occurs here that I feel uncomfortable with is where a signer is provided, fails to connect to the provider, and then the signer provided by the user is "forgotten" and replaced by the signer setup within the provider.
> 
> I think it is actually most correct to let the user configure it correctly. If the signer doesn't support re-configuring the provider, then it should be setup with the correct provider before being sent to the plugin.
> 
> Also you can just provide a signer w/o the provider, and the plugin will get the provider from the signer.
> 
> So the 3 supported cases are:
> 
> * provider only (get signer from provider)
> * signer only (get provider from signer)
> * provider + signer (signer must have same provider)
> 
> Does this make sense to you?

Yes, 100%

PR #1073, opened 2022-07-25
Chore: Remove ABI from schema package

### LINKED ISSUE #973: Generate JSON Schema for TypeInfo
Convert object from ts to json schema so it can be imported as a definition in the wrap manifest https://github.com/polywrap/monorepo/blob/prealpha/packages/schema/parse/src/typeInfo/index.ts#L23-L32

cbrzn:
closes #973 

- [x] use undefined instead of null in abi to losslessly save space while encoding to msgpack.
- [x] replace `Abi` from schema parse with  `WrapAbi` from wrap manifest.
- [x] fix a bug in msgpack binding where it sometimes adds an extra 130 in Uint8Array buffer.


packages/schema/parse/src/abi/definitions.ts
```diff
@@ -477,11 +435,9 @@ export function createImportedEnumDefinition(args: {
   };
 }
 
+// TODO: We don't want this hard coded
```
cbrzn, line 440: Do we?

packages/schema/parse/src/abi/definitions.ts
```diff
@@ -137,14 +132,13 @@ export function createMapKeyDefinition(args: {
   }
   return {
     ...createGenericDefinition(args),
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
cbrzn: The reason for this is that we have `type` attribute in `GenericDefinition` and also in `MapKeyDefinition`, which causes a compilation error. When trying to do `as MapKeyType` like in `createScalarDefinition` function, it is still throwing - Open to any suggestions here

packages/js/manifests/wrap/scripts/templates/index-ts.mustache
```diff
@@ -12,6 +12,7 @@ import {
 } from "./{{version}}";
 {{/formats}}
 
+export * from "./0.1";
```
cbrzn: This has been added because we need to somehow access to the generated definitions so we can use it in `schema/*` packages - Don't really like this line of code because I think we should not export with `*` being the main reason that this can cause to introduce breaking changes without noticing (if we remove a variable or change the name, it would affect the end user if it uses it)

Niraj-Kamdar: Won't that count as a breaking change and update the version to 0.2?

cbrzn: I think that this change + the ones in PR #1046 will introduce a breaking change so there's no problem with that IMO

cbrzn:
If I understand things correctly the change of null to undefined is an improvement of the current structure, right? If that's the case, I would suggest tackling this in two steps, the first entirely removing the Abi definition from `schema/parse` and then making these improvements - What do you think ser @Niraj-Kamdar 

edit: it's worth mentioning that I will follow your lead on this since you're more experienced with this part of the code Niraj, I am just making a suggestion, so feel free to ignore it :P

Niraj-Kamdar:
@cbrzn We could have but I already finished 1/3 of the work

cbrzn:
> @cbrzn We could have but I already finished 1/3 of the work

perfect, go for it then bro

packages/schema/parse/src/__tests__/transforms.spec.ts
```diff
@@ -69,14 +70,12 @@ type AnotherType {
 
 describe("Polywrap Schema Abi Transformations", () => {
   it("addFirstLast", () => {
+    // const parsed = parseSchema(schema1);
+    // console.log(JSON.stringify(parsed));
```
cbrzn:
```suggestion
```

packages/schema/parse/src/abi/index.ts
```diff
@@ -43,139 +16,3 @@ export function createAbi(): Abi {
     importedEnvTypes: [],
   };
 }
-
-type ImportedDefinition = ImportedObjectDefinition | ImportedModuleDefinition;
-
-export function combineAbi(abis: Abi[]): Abi {
```
cbrzn, line 49: Just curious here: Why is this function being removed? I think we might want it back since we are migrating to fetch abi instead of schemas from external wrappers

packages/schema/parse/src/extract/imported-module-types.ts
```diff
@@ -73,7 +77,7 @@ const visitorEnter = (
       map: def
         ? ({ ...def, name: node.name.value } as MapDefinition)
         : undefined,
-      required: def && def.required ? true : false,
+      required: def && def.required ? true : undefined,
```
cbrzn, line 80: It seems kinda weird to me that `required` is being `true` or `undefined` - IMO we should have it true or false as before; no strong opinion on this just sharing my thoughts

cbrzn (commented):
🔥 you rock Niraj - Thanks for pushing this forward

packages/schema/parse/src/abi/definitions.ts
Niraj-Kamdar: Yeah, it's alright because we need to hardcode it and we will be removing the concept of capabilities very soon. I have a few improvements in my mind already and Jordan also agrees with removing this.

packages/schema/parse/src/__tests__/transforms.spec.ts
Niraj-Kamdar: Nice catch!

packages/schema/parse/src/abi/index.ts
Niraj-Kamdar: Why do we need it? I believe we were just using that for combining query and mutation

packages/schema/parse/src/extract/imported-module-types.ts
Niraj-Kamdar: It's much better to keep it undefined then the false because it takes a lot of space to store `required: false`

packages/schema/parse/src/abi/index.ts
cbrzn: After checking the code, you're totally right; sorry for the confusion ser

Niraj-Kamdar (dismissed):

packages/test-cases/cases/compose/sanity/output/module.ts
```diff
@@ -17,226 +17,268 @@ import {
...
-  Abi,
+  WrapAbi,
 } from "@polywrap/schema-parse";
 
-export const abi: Abi = {
-  ...createAbi(),
+export const abi: WrapAbi = {
   envType: createEnvDefinition({
     properties: [
-      createScalarPropertyDefinition({ name: "foo", type: "String", required: true }),
```
dOrgJelli, line 28: @cbrzn can you please disable auto formatting, it bloats the size of PRs and makes it difficult to identify where changes have occurred that are not just styling changes.

packages/schema/parse/src/abi/definitions.ts
dOrgJelli: Agreed, once we treat all wrappers the same, and allow you to `invoke` any wrapper (including interface implementations) the same way we can remove this strange "capability" concept.

The `#use ...` syntax will also be removed from the schema.

packages/js/manifests/wrap/src/formats/wrap.info/0.1.schema.json
```diff
@@ -22,7 +22,8 @@
       "type": "string",
       "enum": [
         "wasm",
-        "interface"
```
dOrgJelli, line 25: These changes should technically be going into a new version of the schema. Since no breaking changes are being made (if I'm correct) the version would be `0.2`.

Since this work is fixing things that should have been done before origin release, we can slide these changes in unnoticed right now IMO, but this should be the last time we do something like this.

dOrgJelli (dismissed):

PR #1083, opened 2022-07-31
HTTP Host/Fetch support: Deploy and infra module + Resolver

### LINKED ISSUE #645: Implement HTTP(S) URI Resolver
If wrappers are being hosted on a central server, addressable using a DNS based domain through HTTP(S), we should support querying them through a new URI resolver + authority.

For example, users could query a wrapper like so:
```typescript
client.invoke({
  uri: "https/wrappers.domain.com/some-wrapper",
  ...
})
```

### LINKED ISSUE #1052: Implement HTTP dev server for wrapper hosting
When developing a web application and using a wrapper that's locally available, or under development, one cannot use filesystem URIs as browsers have no access to the filesystem API. So in order to locally access a wrapper from an application, one needs to resort to spinning up the test environment and pinning the wrapper to a local IPFS node or doing things like wrapping the wrapper in a node module, etc. This was highlighted by ChainSafe as one of the cumbersome things of Polywrap fullstack development.

This could be solved by adding a node server to our "default" infra modules. This node server would expose an endpoint to upload wrappers and would serve them. We could make it have a simple "explorer" UI as well.

We could even host a remote instance of it and make it our standard centralized HTTP hosting solution without IPFS, like a `http.wrappers.io`

This would also entail we'd need to finish #645, which I started but did not finish as it wasn't high priority

namesty:
Closes #645. This PR implements the HTTP resolver and adds a default infra module that spins up an HTTP server (RESTful-like API) to host wrappers and a deploy module that deploys through HTTP to a web server.

The HTTP server, and the deployer module, establish certain conventions that I could formalize into an API doc or similar, after feedback. The HTTP server endpoint to host wrappers accepts FormData, and expects the files to be in a `files[]` FormData key. I also explored encoding the files and sending them in the request's body, and also zipping them and sending a single file; but the FormData approach seemed the most standard and optimal.

After receving a wrapper, the HTTP writes the files to an exposed static directory, namespaced with a user identifier and a wrapper name. Files are automatically available individually through GET requests, but can also be zipped and downloaded in a single bundle.

One thing that's important, is that the HTTP plugin, by default, used to encode the result of a `responseType: binary` request to `base64`. This should be configurable as this makes the HTTP resolver encode and decode each file it resolves. I'll change this in a separate PR as I don't know which integrations that use the HTTP plugin could break because of the change in the schema.


packages/cli/src/lib/defaults/infra-modules/http/server/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@polywrap/infra-http-host",
```
dOrgJelli: Would a better name be "@polywrap/http-wrapper-host"?

packages/cli/src/lib/defaults/infra-modules/http/server/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "version": "1.0.0",
```
dOrgJelli, line 3: Should this package be marked as `"private": true` to avoid our CD pipeline from picking it up and deploying it to npm?

packages/cli/src/lib/defaults/infra-modules/http/server/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "description": "Infra module to host wrappers through HTTP",
```
dOrgJelli: The concept of an "infra module" is external to this package, so I don't think we should be making reference to it here. Would rather just say "Host wrappers through an HTTP endpoint".

packages/cli/src/lib/defaults/infra-modules/http/server/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "main": "src/app.js",
+  "scripts": {
+    "start": "node build/server.js",
+    "dev": "yarn build && yarn start",
+    "build": "rimraf build && tsc -p .",
+    "test": "jest"
+  },
+  "author": "dOrg",
```
dOrgJelli: We don't have this author section filled out in other packages, so I'd say we should leave it out of here.

packages/cli/src/lib/defaults/infra-modules/http/server/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "license": "ISC",
```
dOrgJelli: All other packages are "MIT"

packages/cli/src/lib/defaults/infra-modules/http/server/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "dependencies": {
+    "@polywrap/msgpack-js": "0.3.0",
+    "@types/multer": "^1.4.7",
```
dOrgJelli: We should be fixing our dependencies, so I'd recommend removing the `^`

packages/cli/src/lib/defaults/infra-modules/http/server/src/controllers/wrappers.ts
```diff
@@ -0,0 +1,83 @@
...
+const router = Router();
+
+router.post(
+  "/wrappers/:user/:name",
+  upload.array("file[]", 200),
+  async (req, res) => {
+    try {
+      const user = req.params.user as string;
+      const name = req.params.name as string;
+      const basePath = `${__dirname}/../../wrappers/${user}/${name}`;
```
dOrgJelli: user & name should be sanitized to ensure the caller does not do something like `user = "../../../../some/system/dir"

packages/cli/src/lib/defaults/infra-modules/http/server/src/controllers/wrappers.ts
```diff
@@ -0,0 +1,83 @@
...
+    const user = req.params.user as string;
+    const name = req.params.name as string;
+    const basePath = `${__dirname}/../../wrappers/${user}/${name}`;
+    const zipPath = `${basePath}/${uuidV4()}.zip`;
+
+    const zip = new Zip();
+    await zip.createZip(basePath, zipPath);
+
+    console.log(`Zipped ${basePath} to ${zipPath}`);
+    res.status(200).download(path.resolve(zipPath), "wrappers.zip", (err) => {
```
dOrgJelli: Just checking my own understanding, this is zipping all wrappers that this user has published, correct? I want to make sure we're not zipping a single wrapper, because if so, we can just pre-zip them upon publishing to avoid having to do this for every `get` request.

packages/cli/src/lib/defaults/infra-modules/http/docker-compose.yaml
```diff
@@ -0,0 +1,8 @@
+version: '3'
+services:
+  http-server:
+    build: ./server
+    environment:
+      PORT: '${PORT:-3500}'
```
dOrgJelli: Should this env var name be more specific and less generic? I fear the name `PORT` could easily conflict with other infra modules that get composed together. Or are env vars isolated to a specific "service" within the docker-compose, and we can indeed have multiple infra modules with the same env var name?

packages/cli/src/lib/defaults/infra-modules/http/server/src/utils/uuid.ts
```diff
@@ -0,0 +1,11 @@
+export function uuidV4() {
```
dOrgJelli: Why not use a uuid library for this? We do this here: https://github.com/polywrap/toolchain/blob/0c12092d8609c2254060f626f2219f2271b68b78/packages/js/client/src/PolywrapClient.ts#L1

packages/js/plugins/http/polywrap.plugin.yaml
```diff
@@ -3,3 +3,6 @@ language: plugin/typescript
 name: Http
 module: ./src/index.ts
 schema: ./src/schema.graphql
+import_redirects:
```
dOrgJelli: We should update this format to the latest, since `import_redirects` got renamed to `import_abis`

packages/js/plugins/http/src/__tests__/e2e/integration/polywrap.yaml
```diff
@@ -4,3 +4,6 @@ build: ./polywrap.build.yaml
 language: wasm/assemblyscript
 schema: ./schema.graphql
 module: ./src/index.ts
+import_redirects:
```
dOrgJelli: same comment as above, update to latest manifest format.

packages/js/plugins/http/src/schema.graphql
```diff
@@ -1,33 +1,11 @@
...
+#import { Module } into Http from "ens/http.polywrap.eth"
 
-enum ResponseType {
-  TEXT
-  BINARY
+type Env {
+  """
+  Disable querying providers in parallel when resolving URIs
+  """
+  disableParallelRequests: Boolean
```
dOrgJelli: This doesn't seem to be used anywhere, can we remove it? Not sure it makes sense for this wrapper.

packages/js/plugins/uri-resolvers/http-resolver/README.md
```diff
@@ -0,0 +1 @@
+TODO
```
dOrgJelli: Can we fill this out?

packages/js/plugins/uri-resolvers/http-resolver/polywrap.plugin.yaml
```diff
@@ -0,0 +1,10 @@
+format: 0.1.0
+name: HttpResolver
+language: plugin/typescript
+module: ./src/index.ts
+schema: ./src/schema.graphql
+import_redirects:
```
dOrgJelli: Need to update to latest format.

packages/js/plugins/uri-resolvers/http-resolver/src/__tests__/e2e.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+      args: ["infra", "down", "--modules=http", "--verbose"]
+    });
+
+    if (exitCode !== 0) {
+      throw new Error(`Failed to stop test environment: ${stderr}`);
+    }
+  });
+
+  it("Should successfully resolve a deployed wrapper - e2e", async () => {
+    const wrapperUri = `http/${wrapperHttpUri}`;
```
dOrgJelli, line 60: should also support the `https` authority as well.

packages/js/plugins/uri-resolvers/http-resolver/src/index.ts
```diff
@@ -0,0 +1,89 @@
...
+
+type NoConfig = Record<string, never>;
+
+export class HttpResolverPlugin extends Module<NoConfig> {
+  // uri-resolver.core.polywrap.eth
+  public async tryResolveUri(
+    args: Args_tryResolveUri,
+    _client: Client
+  ): Promise<UriResolver_MaybeUriOrManifest | null> {
+    if (args.authority !== "http") {
```
dOrgJelli: same as above, should also support the `https` authority just in case.

packages/js/test-env/src/index.ts
```diff
@@ -400,3 +401,83 @@ export async function buildAndDeployWrapper({
...
+      ...polywrapManifest,
+      deploy: `./${tempDeployManifestFilename}`,
+    })
+  );
+
+  fs.writeFileSync(
+    tempDeployManifestPath,
+    yaml.dump({
+      format: "0.1.0",
+      stages: {
```
dOrgJelli: is there a way to make this more type safe?

dOrgJelli (changes requested):

dOrgJelli (dismissed):

PR #1085, opened 2022-07-31
Move CLI docgen Templates

dOrgJelli:
I realized that the documentation bindings did not make much sense within the `@polywrap/schema-bind` project's `./bindings` directory. This is because schema-bind currently contains all `BindLanguages` (`wasm-as`, `wasm-rs`, `plugin-ts`, `app-ts`), and documentation does not align well with those. Also the documentation bindings were not being exported from the package in the first place, and instead were being dynamically imported by the CLI.

So I think a more proper solution is to embed the documentation bindings within the CLI, and pass the entrypoint script into the `CodeGenerator` class as it was already being done.

Additional misc cleanup:
* removed redundant code from the CLI's `option-parsers`
* improved the structure of the exports from the `schema-bind` package's `bindings/` directory.
* changed the `docgen` command's default output directory to `./docs` instead of `./wrap`, since `wrap` is used for language bindings (app, plugin, wasm), so there could have been a collision.


dOrgJelli:
# CHANGES
## Patch
* `polywrap` CLI: Added the `./lib/docgen` folder, containing all docgen templates.
* `@polywrap/schema-bind`: Removed the `./bindings/documentation` folder. This was not apart of the package's exports, so it is a non breaking change.

## Breaking Change
* `polywrap` CLI: The `docgen` command's `--docgen-dir` option now defaults to `./docs` instead of `./wrap`.
* `@polywrap/schema-bind`: Language binding modules now export types in the form of `<Language>.<ProjectType>`. For example: `Rust.Wasm`, `TypeScript.App`, etc. The following exports have been removed: `WasmAs`, `WasmRs`, `PluginTs`, `AppTs`.

dOrgJelli:
Merging this because it fixes some bugs present in `origin-dev` that are blocking another PR. @krisbitney please let me know if there are any changes you'd like made to this PR and we can create another.

PR #1091, opened 2022-08-02
Polywrap project manifest schema refactor

### LINKED ISSUE #1057: Refactor: Project + Extension Manifests
Technical Council discussion: https://github.com/polywrap/technical-council/issues/52
# Polywrap Project Manifest Refactor
## Current State
Currently, Polywrap supports the following manifest files:
`polywrap.yaml`: Used for wasm wrapper &amp; interface projects.
`polywrap.plugin.yaml`: Used for plugin wrapper project.
`polywrap.app.yaml`: Used for application projects.
`polywrap.build.yaml`: Customizes the build processed used for wasm wrappers.
`polywrap.deploy.yaml`: Defines the deployment process used for wasm wrappers &amp; interfaces.
`polywrap.meta.yaml`: Defines metadata for a wasm wrapper (description, repository, icon, etc).
`polywrap.infra.yaml`: Defines custom docker containers used by the wrapper.

## Problem
It is not clear which of these manifests are used to define a project, and which ones are “extensions” or “customizations” to that project.

#### Common Properties
```yaml
format: 0.1
project:
  name: foo-bar
  type: wasm/... | interface | plugin/... | app/...
```

#### Wasm
`polywrap.yaml`
```yaml
format: 0.1
project:
  name: foo-bar
  type: wasm/assemblyscript
source:
  schema: ./schema.graphql
  module: ./Cargo.toml
```

#### Interface
`polywrap.yaml`
```yaml
format: 0.1
project:
  name: bizz-buzz
  type: interface
source:
  schema: ./schema.graphql
```

#### Plugin
`polywrap.yaml`
```yaml
format: 0.1
project:
  name: my-plugin
  type: plugin/typescript
source:
  schema: ./schema.graphql
  module: ./src/index.ts
```

#### App
`polywrap.yaml`
```yaml
format: 0.1
project:
  name: my-plugin
  type: app/typescript
source:
  schema: ./schema.graphql
```

### Project Extensions
Every project type (wasm, interface, plugin, app) has a set of supported extension types: build, infra, meta, deploy, codegen.

For example:
```yaml
format: ...
project:
  name: ...
  type: wasm/rust
source:
    schema: ...
    module: ...
extensions:
  build: ./polywrap.build.yaml
  infra: ./polywrap.infra.yaml
  deploy: ./polywrap.deploy.yaml
  meta: ./polywrap/metadata.yaml
```

## Notable Codebase Changes
- We need to decouple the `wasm` & `interface` project types within the codebase
- interface versions = major + minor

### Version Pinning: CLI + Runtime
```yaml
format: 0.1
project:
  ...
  versions:
    polywrap: 1.2.3  #cli version
    wrap: 1.2+origin #wrap version
source:
  schema: ./schema.graphql
```
- `polywrap build`
    - checks `project.versions.polywrap` version
- retarget your codegen/compilation to different wrap runtimes

### Different Extension Manifests
Can now support multiple manifests per extension:
```yaml
...
extensions:
  build:
    debug: ./manifests/polywrap.build.debug.yaml
    prod: ./manifests/polywrap.build.prod.yaml
```

`polywrap build ./polywrap.yaml --type debug`
`polywrap build debug ./polywrap.yaml`

### Inline Extension Manifests
```yaml
...
extensions:
  build:
    format: 0.2
    docker:
      ...
```

pileks:
Partially closes #1057 

This PR performs the 1st steps of the refactor of the Polywrap project manifest files (wasm, interface, plugin, app).

Things to note:
- The loading of `polywrap.app.yaml`, `polywrap.plugin.yaml` etc. have been left as such for backwards compatibility.
  - [x] We might want to consider a CLI command to migrate the YAML files to the latest format version, instead of just reading them and migrating them upwards. Opening an issue seems like a good idea - any thoughts/concerns?
  - [x] Create an issue for a `polywrap manifest` CLI command, with option for migrating the manifest to the latest version
- [x] It has been decided to use `major.minor.patch` versioning (as per semver) instead of just `major.minor` for now - this allows us to circumvent the yaml parser from reading `format: 0.2` as a number and forcing us to use `format: “0.2”`
- [x] *Important:* All manifest files within the project have been left on version 0.1 for the time being in order to keep builds working with the existing polywrap CLI pulled from `npm`. I advise we leave them as such until we publish a version of the CLI wich supports the `0.2.0` project manifest format.
  - [x] Explicitly test migrations
  - [x] Move test manifests to 0.2.0


packages/js/manifests/polywrap/src/formats/polywrap.plugin/0.2.0.ts
```diff
@@ -0,0 +1,75 @@
...
+    deploy?: string;
+    /**
+     * Path to project infra manifest file.
+     */
+    infra?: string;
+  };
+  /**
+   * Redirects for the schema's imports.
+   */
+  import_redirects?: {
```
dOrgJelli: I think it'd actually make a lot of sense to move the import_redirects within the `source` property. `source.import_redirects`. Thoughts?

dOrgJelli (changes requested):

packages/js/manifests/polywrap/src/formats/polywrap.plugin/0.2.0.ts
pileks: Since `import_redirects` realistically only serves the schema, I'd agree with this. I'll make the changes!

packages/cli/src/lib/project/PluginProject.ts
```diff
@@ -12,7 +12,7 @@ import { PluginManifest } from "@polywrap/polywrap-manifest-types-js";
 import { bindSchema, BindOutput, BindOptions } from "@polywrap/schema-bind";
 import { ComposerOutput } from "@polywrap/schema-compose";
 import path from "path";
-import { WrapAbi } from "@polywrap/wrap-manifest-types-js";
+import { Abi } from "@polywrap/wrap-manifest-types-js";
```
cbrzn: I think this would be `WrapAbi` even tho `Abi` is exported

packages/cli/src/lib/project/manifests/app/load.ts
```diff
@@ -6,7 +6,11 @@ import {
 } from "@polywrap/polywrap-manifest-types-js";
 import fs from "fs";
 
-export const defaultAppManifest = ["polywrap.app.yaml", "polywrap.app.yml"];
+export const defaultAppManifest = [
+  "polywrap.yaml",
+  "polywrap.app.yaml",
+  "polywrap.app.yml",
```
cbrzn, line 12: Should we still support this naming? :thinking:

packages/cli/src/lib/project/manifests/plugin/load.ts
```diff
@@ -7,6 +7,7 @@ import {
 import fs from "fs";
 
 export const defaultPluginManifest = [
+  "polywrap.yaml",
   "polywrap.plugin.yaml",
```
cbrzn, line 11: Same here as previous comment

packages/js/manifests/polywrap/scripts/templates/migrate-ts.mustache
```diff
@@ -31,13 +31,18 @@ export function migrate{{#latest}}{{type}}{{/latest}}(
   manifest: Any{{#latest}}{{type}}{{/latest}},
   to: {{#latest}}{{type}}{{/latest}}Formats
 ): {{#latest}}{{type}}{{/latest}} {
-  const from = manifest.format as {{#latest}}{{type}}{{/latest}}Formats;
+  let from = manifest.format as {{#latest}}{{type}}{{/latest}}Formats;
+
+  // HACK: Patch fix for backwards compatability
```
cbrzn, line 36: Is this hack temporary or do u think it will stay for the long term? If is the latter if would remove the hack word and just add the comment to make it cleaner

packages/cli/src/commands/docgen.ts
```diff
@@ -35,9 +35,11 @@ const commandToPathMap: Record<string, string> = {
 
 export type DocType = keyof typeof commandToPathMap;
 
+// A list of UNIQUE possible default filenames for the polywrap manifest
 const defaultManifest = defaultPolywrapManifest
   .concat(defaultAppManifest)
-  .concat(defaultPluginManifest);
+  .concat(defaultPluginManifest)
+  .filter((value, index, self) => self.indexOf(value) === index);
```
cbrzn, line 40: This has been added because now the `app` and `plugin` manifest can be `polywrap.yaml` - Correct?

packages/js/manifests/polywrap/src/formats/polywrap.app/0.2.0.ts
```diff
@@ -0,0 +1,71 @@
...
+   */
+  project: {
+    /**
+     * Name of this project.
+     */
+    name: string;
+    /**
+     * Type of this project.
+     */
+    type: string;
```
cbrzn, line 25: I find it weird to have `type` inside the `app` manifest - Not sure tho...

packages/js/manifests/polywrap/src/formats/polywrap.plugin/0.2.0.ts
```diff
@@ -0,0 +1,75 @@
...
+    type: string;
+  };
+  /**
+   * Project source files.
+   */
+  source: {
+    /**
+     * Path to the project's entry point.
+     */
+    module?: string;
```
cbrzn: `module` should always be defined IMO; otherwise, it would be an interface right?

packages/cli/src/commands/docgen.ts
pileks: Correct!

packages/js/manifests/polywrap/src/formats/polywrap.plugin/0.2.0.ts
pileks: You are correct, I missed that one.

packages/js/manifests/polywrap/src/formats/polywrap.app/0.2.0.ts
pileks: If we take a look at the spec https://github.com/polywrap/monorepo/issues/1057 - the `type` field is a common field for all polywrap manifests (plugin, wasm, app, interface).

If I am to run `polywrap app xyz`, how would the CLI be sure that the current project is an app, unless the `type` explicitly said so? 😄

LMK if there's something I missed.

cbrzn: true! my bad

pileks:
Seems like we've got a conflict as there's a new version of the polywrap manifest (0.2).

I'll move the manifests in this PR to version 0.3 (or maybe even 1.0 as there are breaking changes?) and include anything I missed from 0.2

cbrzn:
I would suggest bumping to 0.3

packages/cli/src/commands/docgen.ts
dOrgJelli: This is a clever way to get a unique list!

packages/cli/src/lib/project/manifests/app/load.ts
dOrgJelli: for legacy sake, sure, but not going forward :P

packages/js/manifests/polywrap/scripts/templates/migrate-ts.mustache
dOrgJelli: Yah I think we can potentially get rid of this in the near future, since the `0.1` format was only "live" for a very short time so if there are no published wrappers with this version we can remove this backwards compatibility check in the future.

Also we'll remove all back-compat when we reach `1.0.0`.

dOrgJelli:
Hey @pileks, we actually want to use version `0.2.0` instead of creating another format version (`0.3.0`). The `0.2.0` you were getting conflicts with was still unreleased (only in `origin-dev`), so if we did `0.3.0` we would have bumped by 2 versions in one release.

PR #1092, opened 2022-08-02
Add WebSocket plugin

### LINKED ISSUE #575: Create WebSocket Plugin
Create a WebSocket plugin to compliment the existing HTTP plugin.

An example interface might be:
```
#import { Query } into WS from "w3://ens/websocket.web3api.eth"
```
```typescript
import { WS_Query } from "./w3";

// Open a new socket connection
const wsId = WS_Query.open(...)

// Pull a response
let resp = WS_Query.pull(wsId)

// Based on some condition, choose to keep pulling, or...
while (resp.body.confirmations < reqConfirmations) {
  sleep(...)
  resp = WS_Query.pull(wsId)
}

// Close the socket connection
WS_Query.close(wsId)
```

fetsorn:
Resolves #575. WebSocket plugin allows Polywrap Client to interact with WebSocket servers.

Every incoming WebSocket message can be passed to a callback function in another wrapper. Use `addCallback` to start passing messages and `removeCallback` to stop. The callback function is expected to have a parameter `data`, i.e. `foo(data: string)`.

Incoming WebSocket messages can also be stored in the plugin and retrieved as an array later. Use `addCache` to start caching messages and `removeCache` to stop. Use `receive` to get an array of cached messages and empty the cache. Optionally, wait for a timeout, or a minimum number of cached messages before retrieving the array. 

See the README.md and src/__tests__/e2e/e2e.spec.ts for more info.


packages/js/plugins/ws/README.md
```diff
@@ -0,0 +1,130 @@
...
+# path to WRAP method
+type Callback {
+    uri: String!,
+    method: String!
+}
+
+# optional fields are `Number | null` instead of `Option<i32>`
+type Number {
+    value: Int!
+}
```
Niraj-Kamdar, line 24: Why are we doing this?

Niraj-Kamdar (dismissed):
LGTM! Can you reply on my comment?

packages/js/plugins/ws/README.md
fetsorn: Optional Int is currently implemented as `Option<i32>` in wasm and schema/bind.
But to omit optional fields from the function parameters we need a ` | null` union type.
So we wrap Int in a reference type and make it optional, `Number | null`.

I mentioned this in #944, but now I think we might want to implement Optional Int as `Option<i32> | null` and always use null instead of `Option.None<i32>()`, to no longer require wrapping such as this one.

Niraj-Kamdar: Why don't we use Box instead in that case?

Niraj-Kamdar: i.e. `Box<i32> | null` instead of `Option<i32> | null`

fetsorn: Exactly, created an issue for that #1132.

dOrgJelli (approved):

PR #1095, opened 2022-08-03
Use map instead of array of pair http plugin

### LINKED ISSUE #863: use map instead of array of pair for urlParams in http plugin

pileks:
Resolves #863

Additional work:

`packages/wasm/as/assembly/msgpack/Write.ts`
- Changed signature of abstract method `writeOptionalExtGenericMap` to allow for `null` maps, putting it in line with its implementations in `WriteEncoder` and `WriteSizer`

`packages/js/plugins/http/src/__tests__/e2e/integration/polywrap.yaml`
- Added `import_redirects` for HTTP plugin - codegen needs this to import the latest HTTP plugin schema.

Formatted test files in line with `.prettierrc`

### DIFF
--- a/packages/js/plugins/http/src/__tests__/unit/index.test.ts
+++ b/packages/js/plugins/http/src/__tests__/unit/index.test.ts
@@ -34,10 +34,10 @@ describe("test http plugin", () => {
         {
           url: "/api/test",
           request: {
-            headers: [
-              { key: "Accept", value: "application/json" },
-              { key: "X-Test-Header", value: "test-header-value" },
-            ],
+            headers: new Map([
+              ["Accept", "application/json"],
+              ["X-Test-Header", "test-header-value"],
+            ]),
             urlParams: new Map([["q", "test-param"]]),
             responseType: ResponseTypeEnum.TEXT,
           },
@@ -56,9 +56,9 @@ describe("test http plugin", () => {
 
       expect(response?.status).toBe(200);
       expect(response?.statusText).toBe("Ok");
-      expect(response?.headers).toStrictEqual([
-        { key: "Content-Type", value: "application/json; charset=utf-8" },
-      ]);
+      expect(response?.headers).toStrictEqual(
+        new Map([["Content-Type", "application/json; charset=utf-8"]])
+      );
       expect(response?.body).toBe("{result: 1001}");
     });
 
@@ -77,10 +77,10 @@ describe("test http plugin", () => {
         {
           url: "/api/test",
           request: {
-            headers: [
-              { key: "Accept", value: "application/json" },
-              { key: "X-Test-Header", value: "test-header-value" },
-            ],
+            headers: new Map([
+              ["Accept", "application/json"],
+              ["X-Test-Header", "test-header-value"],
+            ]),
             urlParams: new Map([["q", "test-param"]]),
             responseType: "BINARY",
           },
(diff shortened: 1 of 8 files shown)


packages/js/plugins/http/src/schema.graphql
```diff
@@ -17,7 +17,7 @@ type Response {
 
 type Request {
   headers: [Header!]
-  urlParams: [UrlParam!]
+  urlParams: Map @annotate(type: "Map<String!, String!>")
```
dOrgJelli, line 10: We can remove the `type UrlParam` above, no?

packages/wasm/as/assembly/msgpack/Write.ts
```diff
@@ -63,7 +63,7 @@ export abstract class Write {
     value_fn: (writer: Write, value: V) => void
   ): void;
   abstract writeOptionalExtGenericMap<K, V>(
-    m: Map<K, V>,
+    m: Map<K, V> | null,
```
dOrgJelli, line 66: Nice catch, thank you :)

packages/js/plugins/http/src/__tests__/unit/index.test.ts
```diff
@@ -158,18 +169,21 @@ describe("test http plugin", () => {
...
-          body: "{request: 1001}",
-          responseType: "BINARY",
-        }
-      }, {} as Client);
+      const response = await plugin.post(
+        {
+          url: "/api/test",
+          request: {
+            headers: [
+              { key: "Accept", value: "application/json" },
```
dOrgJelli: Why not make the headers a `Map` as well?

dOrgJelli (changes requested):

packages/js/plugins/http/src/schema.graphql
```diff
@@ -17,7 +17,7 @@ type Response {
 
 type Request {
   headers: [Header!]
-  urlParams: [UrlParam!]
```
Niraj-Kamdar, line 20: I think We can also do the same for header and replace it with Map

Niraj-Kamdar (changes requested):
I realised after reviewing this that we can also replace header with Map

packages/js/plugins/http/src/schema.graphql
pileks: Yup, forgot about it.

packages/js/plugins/http/src/schema.graphql
pileks: Yup, gonna do that.

packages/js/plugins/http/src/__tests__/unit/index.test.ts
pileks: Yup!

pileks:
@dOrgJelli @Niraj-Kamdar After changing the headers to maps, I've decided that I might as well refactor all instances of `query` in the tests to `invoke` as well. Just an FYI.

Niraj-Kamdar (approved):

cbrzn (approved):
LGTM :green_apple:

cbrzn:
I think this is ready to merge but since you requested changes you need to approve ser @dOrgJelli

pileks:
All requested changes should be addressed now, so yes, we can go forward with merging 😄

Niraj-Kamdar (approved):

dOrgJelli (approved):

PR #1097, opened 2022-08-05
Deploy Manifest Refactor + Recursive ENS Registration Deployer

### LINKED ISSUE #1138: local-dev-ens Deploy Module Doesn't Output `/ens/testnet/...` URI
When using the local-dev-ens deploy module, the URI that is output is a mainnet ENS URI, and does not have the `/testnet/` network in the path.

I also think it would make more sense to use `/local/` potentially? If we make the change to `/local/` we need to change all other occurrences of `/testnet/`.

namesty:
Closes #1138


packages/cli/src/commands/deploy.ts
```diff
@@ -69,104 +72,78 @@ async function run(options: DeployCommandOptions): Promise<void> {
...
   ];
 
   sanitizePackages(packageNames);
 
   await project.cacheDeployModules(packageNames);
 
   const packageMap: Record<string, DeployPackage> = {};
-  const stageToPackageMap: Record<string, DeployPackage> = {};
 
   for await (const packageName of packageNames) {
```
Niraj-Kamdar: Why are we awaiting in this loop?

packages/cli/src/commands/deploy.ts
```diff
@@ -69,104 +72,78 @@ async function run(options: DeployCommandOptions): Promise<void> {
...
 
-  for await (const root of roots) {
-    print.info(`\nExecuting deployment chain: \n`);
-    root.handler.getDependencyTree().printTree();
-    print.info("");
-    await root.handler.handle(root.uri);
-    resultLists.push(root.handler.getResultsList());
+  for await (const sequence of sequences) {
+    const sequenceResult = await sequence.run();
+    sequenceResults.push(sequenceResult);
```
Niraj-Kamdar: Do sequence need to be run in sequence? Why don't we parallelize it?

Niraj-Kamdar (changes requested):
What if we convert the deployers into the wrappers, won't we be able to use workflow directly for defining the deployer manifest?

packages/cli/src/commands/deploy.ts
namesty: Yep, parallelized it

packages/cli/src/commands/deploy.ts
namesty: Yep, done

namesty:
> What if we convert the deployers into the wrappers, won't we be able to use workflow directly for defining the deployer manifest?

@Niraj-Kamdar I discussed this with Jure a week ago, and the conclusion was that when we'll make them wrappers in the near future (after wrapper embedding) and inevitably we'll need to tweak and make changes to the deploy command's workflow; but it shouldn't affect its usage as a user

dOrgJelli (dismissed):

packages/js/manifests/polywrap/src/formats/polywrap.deploy/migrators/0.1.0_to_0.2.0.ts
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+
+import { DeployManifest as OldManifest } from "../0.1.0";
```
dOrgJelli, line 3: Can we make sure we have tests that stress this code?

packages/js/test-env/src/index.ts
```diff
@@ -27,6 +21,12 @@ export const providers = {
   ethereum: "http://localhost:8545",
 };
 
+export const embeddedWrappers = {
+  ens: `wrap://fs/${path.join(__dirname, "wrappers", "ens")}`,
+  uts46: `wrap://fs/${path.join(__dirname, "wrappers", "uts46")}`,
+  sha3: `wrap://fs/${path.join(__dirname, "wrappers", "sha3")}`,
```
dOrgJelli, line 27: Why do we need to embed uts46 & sha3?

dOrgJelli (changes requested):

packages/js/test-env/src/index.ts
cbrzn: I think that the reason is that ENS wrapper needs those to work

krisbitney: Tests sometimes fail to resolve the URIs and fetch the wrappers. It happens enough to make it hard to get all of the CI tests to pass at the same time. The ENS wrapper depends on the uts46 and sha3 wrappers now, so if we are embedding the ENS wrapper then it makes sense to embed uts46 and sha3 for the same reasons.

Niraj-Kamdar (dismissed):
Awesome work! Should be good to merged once @dorgJelli 's comments are addressed.

dOrgJelli (dismissed):

PR #1098, opened 2022-08-05
Use jest for template testing

### LINKED ISSUE #873: use jest for template testing instead of custom script
The current way of running these tests with a js script isn't very effective and making it hard to debug this efficiently.

pileks:
Resolves #873 

Now using jest harness to run the test commands.

Tests are still run using `child_process`, but we're now using the `exec` function within a promise which returns an `Error` object on failure, which we can assert.

The scaffolding has been kept pretty much the same, as it seems to do a decent job of it.

We could move to a more explicit scaffolding solution (specify commands for each template individually) if we'd like to.


packages/templates/tests.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+
+describe("Templates", () => {
+  const rootDir = __dirname;
+
+  // Mapping of all project types (app, plugin, etc) to all supported languages
+  const projectLanguages: Record<string, string[]> = {};
+
+  // Define the commands to run for each language
+  const languageTestCommands: Record<string, string[]> = {
+    typescript: ["yarn build", "yarn test"],
```
Niraj-Kamdar: How about making key-value pair of command name to actual command so that we can use command name instead of whole command as test title in jest.

Ex: {build: "yarn build", "test": "yarn test"}, etc.

Niraj-Kamdar (changes requested):
I believe there shouldn't be any issue regarding workflow tests in CI now. so we can enable it. make sure it's passing locally though.

packages/templates/tests.spec.ts
pileks: Funny thing is that it was passing locally with a smaller timeout... 😄

I agree with the naming strategy.

packages/templates/tests.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    },
+    "typescript-react": {
+      // Uncomment when the helloworld wrapper has been deployed to polywrap.eth
+      // build: "CI=false yarn build"
+    },
+    assemblyscript: {
+      // Workflow tests fail in CI because cuelang is not installed
+      build: "yarn build",
+      test: "yarn test:e2e",
+      // "yarn test:workflow"
```
Niraj-Kamdar: remove this test:workflow, e2e will include it anyway

packages/templates/tests.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  const rootDir = __dirname;
+
+  // Mapping of all project types (app, plugin, etc) to all supported languages
+  const projectLanguages: Record<string, string[]> = {};
+
+  // Define the commands to run for each language
+  const languageTestCommands: Record<string, Record<string, string>> = {
+    typescript: { build: "yarn build", test: "yarn test" },
+    "typescript-node": {
+      // Uncomment when the helloworld wrapper has been deployed to polywrap.eth
```
Niraj-Kamdar: I don't think we need it deployed to polywrap.eth, we can just redirect it to the current uri. (i.e. ens/goerli/hello.as.demos.wraplib.eth)

Niraj-Kamdar (changes requested):

packages/templates/tests.spec.ts
pileks: @Niraj-Kamdar Goerli was giving me some trouble, so I decided to use the hello-world wrapper deployed on `mainnet` via @nerfZael's wrappers.io.
Address is `ens/helloworld.polytest.eth`.

packages/templates/tests.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  // for each project + language
+  for (const projectType of Object.keys(projectLanguages)) {
+    for (const language of projectLanguages[projectType]) {
+      // run the test commands at commands[language]
+      let commands = languageTestCommands[language];
+
+      // if nothing exists, try to match language as a substring
+      if (!commands) {
+        for (const testLanguage of Object.keys(languageTestCommands)) {
```
Niraj-Kamdar: Instead of this pattern:
```ts
for (x) {
  for (y) {
   test (`${x} {y}`) {
}
}
}
```

use following for better test hierarchy:
```ts
for (x) {
 test(x) {
  for(y) {
   test(y) {}
}
}
}
```

Niraj-Kamdar (changes requested):

packages/templates/tests.spec.ts
```diff
@@ -55,46 +55,48 @@ describe("Templates", () => {
   // for each project + language
   for (const projectType of Object.keys(projectLanguages)) {
     for (const language of projectLanguages[projectType]) {
-      // run the test commands at commands[language]
-      let commands = languageTestCommands[language];
+      describe(`${projectType}/${language}`, () => {
+        // run the test commands at commands[language]
+        let commands = languageTestCommands[language];
```
Niraj-Kamdar: sorry for being too pedantic but why don't we have `for...describe...for...describe` instead

Niraj-Kamdar (changes requested):

packages/templates/tests.spec.ts
Niraj-Kamdar: My reasoning is that it would be much easier to execute the selected tests if we have proper hierarchy. 
`jest -t "projectType"` would execute every project type and `jest -t "projectType language"` would execute project with the specified language

pileks: Heheh, it's alright, and we can go with that!

Niraj-Kamdar (approved):
Thanks! LGTM!

cbrzn (approved):

PR #1100, opened 2022-08-09
add manifest schema for workflow file

### LINKED ISSUE #970: add manifest schema for workflow file

fetsorn:
Resolves #970.
The Uri field in the workflow script has type `any` so that it is compatible with core's `TUri extends URI | String` type.


cbrzn (commented):
I think it would make sense not to create a new package for the workflow manifest but rather, to add it to the `wrap` manifests (will live along with Abi). Since IIRC we are making the workflow a part of the WRAP protocol

packages/manifests/workflow/formats/workflow/0.1.json
```diff
@@ -0,0 +1,85 @@
...
+  "id": "PolywrapWorkflow",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "format",
+    "name",
+    "schema",
+    "language"
+  ],
+  "required": ["format", "name", "jobs"],
```
Niraj-Kamdar: Remove redundant properties

packages/js/manifests/workflow/src/formats/workflow/index.ts
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable */
+/**
+ * This file was automatically generated by scripts/manifest/index-ts.mustache.
+ * DO NOT MODIFY IT BY HAND. Instead, modify scripts/manifest/index-ts.mustache,
+ * and run node ./scripts/manifest/generateFormatTypes.js to regenerate this file.
+ */
+
+import {
+  PolywrapWorkflow as PolywrapWorkflow_0_1
```
Niraj-Kamdar: Add PolywrapJobs here as we discussed and export it from here

packages/js/manifests/workflow/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./formats";
+export * from "./deserialize";
+export { Jobs } from "./formats/workflow/0.1"
```
Niraj-Kamdar: Instead of this, it should be exported from the formats

packages/js/manifests/workflow/README.md
```diff
@@ -0,0 +1,3 @@
+# @polywrap/polywrap-manifest-types-js
```
Niraj-Kamdar: this should be `workflow-manifest-types-js`

packages/js/manifests/workflow/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@polywrap/workflow-manifest-types-js",
+  "description": "Polywrap Manifest TypeScript Typings",
```
Niraj-Kamdar: Polywrap -> Workflow

Niraj-Kamdar (changes requested):
LGTM once the requested changes are addressed

Niraj-Kamdar (dismissed):

Niraj-Kamdar:
@fetsorn Can you fix the conflicts so we can get this merged and then would be able to merge the `polywrap.test.yaml` PR.

Niraj-Kamdar (dismissed):

PR #1130, opened 2022-08-10
[DRAFT] Added `setPlugin` method, allowing client users to replace or add a plugin at runtime

krisbitney:
usage:
```typescript
client.setPlugin({
  uri: "wrap://ens/ethereum.polywrap.eth",
  plugin: ethereumPlugin({ ... }),
});
```

closes https://github.com/polywrap/monorepo/issues/1089


krisbitney:
After discussion, we determined it is better to keep the client config immutable for now. Replaced by: https://github.com/polywrap/monorepo/pull/1143

PR #1133, opened 2022-08-11
Allow user to get plugin config from client

### LINKED ISSUE #839: allow user to get plugin configs from the client or Web3ApiClientConfig object

pileks:
Closes #839
The implementation fetches the plugin config by calling its factory function within the PluginPackage and extracting the config from the resulting object.


krisbitney:
It seems expensive to construct a plugin instance to get the config. What are the pros and cons of changing the `PluginPackage` type to include a pointer to the config?

```
export type PluginPackage<TConfig> = {
  factory: () => PluginModule<TConfig>;
  manifest: PluginPackageManifest;
  config: TConfig;
};
```

pileks:
@krisbitney The main con that came to mind when looking at this issue was that we would need to trust the developer that their implementation of the PluginFactory for their plugin actually assigns the config being used by the plugin to that field (or assigns it at all).
This does sound like something we could enforce though, so it could be considered a moot point. 😄

My main issue in general here has to do with the recent work of moving away from plugin configs and placing any configuration into Env. I have actually recently worked on an issue that's related to this: https://github.com/polywrap/monorepo/pull/1060

The question of "how much time we want to spend on this?" kind of comes to mind, and whether we are really moving away from configs into envs, making even my implementation obsolete at one point in the future. And in any case, do we foresee that we'll be using this `getConfig` often enough to warrant us making these significant changes in order to be able to fetch the config of a plugin.

I'm sure that I'm missing something important when considering all of this and I'd love to chat regarding this so feel free to hit me up on Discord and we'll come up with something smart!

krisbitney:
All of that makes sense. Thanks!

packages/js/client/src/PolywrapClient.ts
```diff
@@ -171,6 +172,16 @@ export class PolywrapClient implements Client {
     );
   }
 
+  @Tracer.traceMethod("PolywrapClient: getPluginConfig")
+  public async getPluginConfig<TUri extends Uri | string>(
+    uri: TUri,
+    options: GetPluginConfigOptions = {}
```
cbrzn: Since we are using options to interact with the `getPlugin` function, I think it would make sense to re-use the already existing type `GetPluginOptions`

nerfZael:
I don't think this feature makes sense, since we are trying to phase out plugins from the client
Even if it did, why not do `client.getPlugin()` then get the config from there? 

On your point of moving from configs to envs, that's not completely true, configs can do what envs can not (e.g. pass objects by reference) so they will always have their place.

dOrgJelli:
I agree with Jure, I think we should instead just expose a helper `client.getPlugin(uri)` method.

dOrgJelli (changes requested):
comment above

Niraj-Kamdar (dismissed):

dOrgJelli (dismissed):

PR #1135, opened 2022-08-11
Client config builder methods

### LINKED ISSUE #1122: Client config builder

pileks:
Closes #1122 

Possible future tasks:

- [ ] `PolywrapClient` probably shouldn't know about `ClientConfigBuilder` and shouldn't know about `defaults`. Add issue?
  - Should remove the `noDefaults` option within the constructor
  - PolywrapClient should just apply the config it's provided, not modify/add to it upon construction
  - Good separation of concerns, simpler client code.

Initial set of the ClientConfigBuilder methods.

Methods implemented:

### Plugin methods

```typescript
addPlugin(uri: string, plugin: PluginPackage)
```
Adds specified `plugin` registered under the `uri`. If a plugin already exists, overwrites it.

```typescript
removePlugin(uri: string)
```
Removes the plugin registered under the `uri`.

### Env methods
```typescript
addEnv(uri: string, env: Record<string, unknown>)
```
Adds specified `env` registered under the `uri`. If an env is already registered under the `uri`, it performs a merge (currently using naive implementation - spread operator).

```typescript
setEnv(uri: string, env: Record<string, unknown>)
```
Adds specified `env` registered under the `uri` if none exists. Otherwise, overwrites the `env` that is already registered under the `uri`.

```typescript
removeEnv(uri: string)
```
Removes the `env` registered under the `uri` if one exists. Otherwise, does nothing.

### InterfaceImplementation methods

```typescript
addInterfaceImplementation(interfaceUri: Uri | string, implementationUri: Uri | string)
```
Adds the `interfaceImplementationUri` to the `implementations` collection for the specified `interfaceUri`. Creates the InterfaceImplementation if none exists for `interfaceUri`.

```typescript
addInterfaceImplementations(interfaceUri: Uri | string, implementationUris: Array<Uri | string>)
```
Appends all `interfaceImplementationUris` to the the `implementations` collection for the specified `interfaceUri`. Creates the InterfaceImplementation if none exists for `interfaceUri`.

```typescript
removeInterfaceImplementation(interfaceUri: Uri | string, implementationUri: Uri | string)
```
Removes the `interfaceImplementationUri` from the `implementations` collection for the specified `interfaceUri`. Removes the InterfaceImplementation if the `implementations` collection would remain empty.

### UriRedirect methods

```typescript
addUriRedirect(from: Uri | string, to: Uri | string)
```
Adds the specified UriRedirect. If an existing UriRedirect has the same `from`, overwrites it.

```typescript
removeUriRedirect(from: Uri | string)
```
Removes the specified UriRedirect.

### UriResolver methods
```typescript
addUriResolver(resolver: UriResolver)
```
Adds the specified UriResolver to the end of the `uriResolvers` collection.

```typescript
setUriResolvers(resolvers: UriResolver[])
```
Sets the `uriResolvers` collection to `resolvers`.


packages/js/client-config/src/ClientConfigBuilder.ts
```diff
@@ -67,91 +62,227 @@ export class ClientConfigBuilder {
...
+
+    return this;
+  }
+
+  removePlugin(uri: Uri | string): ClientConfigBuilder {
+    const pluginUri = toUri(uri);
+
+    const idx = this._config.plugins.findIndex((x) =>
+      Uri.equals(x.uri, pluginUri)
+    );
```
cbrzn, line 92: Would it be possible to use `find` instead of `findIndex`? I find it more readable

packages/js/client/src/createPolywrapClient.ts
```diff
@@ -71,10 +72,17 @@ export const createPolywrapClient = Tracer.traceFunc(
     }
 
     if (config) {
-      return new PolywrapClient({
-        ...config,
-        plugins: [...plugins, ...(config.plugins ? config.plugins : [])],
-      });
+      const builder = new ClientConfigBuilder().add(config);
+
+      if (plugins.length) {
```
cbrzn: Do we need to do this validation? :thinking:

pileks: Nah, not really, you're right.

This is also a helper function which I'd honestly like to get rid of 😄
While it seems useful, the config builder kind of defeats its purpose. Tests can be refactored to not use this if we choose to do so.

packages/js/client-config/src/ClientConfigBuilder.ts
pileks: I've taken the liberty of using both across the builder. Here specifically I've decided to go with `findIndex` since it's a `remove` method, and the easiest way to remove an item from an array is `splice` which requires the `index`.
Maybe you have an alternative in mind? 😄

cbrzn: Just thinking out loud, but wouldn't something like
```typescript
this._config.plugins = this._config.plugins.filter(p => !Uri.equals(p.uri, pluginUri))
```
do the job here?

dOrgJelli:
I'm onboard with all "future improvements" mentioned above!

packages/js/client-config/src/ClientConfigBuilder.ts
dOrgJelli: The `findIndex` and `splice` implementation is actually faster, since `filter` include redundant operations.

For example:
```typescript
const arr = ["a", "b", "c", "d"]

const findMe = "b";

// 2 iterations + 1 array shift (optimized)
const idx = arr.findIndex((x) => x === "b");
arr.splice(idx, 1);

// 4 iterations + 1 array reconstruction (not optimized)
arr.filter((x) => x === findMe);
```

It's also important to note that the overhead here isn't necessarily the iterations of the predicate (`===`), but rather the allocation of memory. In the case of `splice(..., 1)`, you can easily optimize this by doing a simple array shift. In the case of `filter((x) => ...)`, you cannot know what `...` is because it is arbitrary, so you cannot optimize the array's reconstruction as easily (not sure hot it's actually implemented).

dOrgJelli (approved):
:fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire:

PR #1136, opened 2022-08-12
refactor workflow from core to cli

### LINKED ISSUE #1131: refactor workflow out of core into cli
Workflow logic should reside in `cli`, and the workflow type should be imported from polywrap-manifest-types-js.

fetsorn:
Based on #1100 
Resolves #1131

 - move workflow from core and client to cli
 - move workflow schema to polywrap manifest schemas


Niraj-Kamdar (commented):
Awesome work! Thanks for pushing this to completion. 

Just a minor chnage: args in wokrflow aren't a required property. for void function it can be undefined.

cbrzn (approved):
Amazing work :-D I really like this approach of `polywrap.test`

Niraj-Kamdar (approved):
Awesome! Thanks for pushing this to completion

PR #1143, opened 2022-08-16
Modify Ethereum plugin config to use Connections store

krisbitney:
This PR modifies the Ethereum plugin's config to use a `Connections` store. The `Connections` class stores `Connection` instances. App developers can keep a reference to the Connections store and use it to update networks and providers without reconstructing the plugin or client.

Original spec: https://hackmd.io/@polywrap-dao/BkQdG7NAq

Implemented UX:
```typescript
const connections = new Connections({
  networks: {
    mainnet: new Connection({ provider }),
    rinkeby: new Connection(...)
  },
  defaultNetowrk: "mainnet"
})

ethereumPlugin({
  connections
});

// assign a Connection to "ropsten" network, or replace existing connection if ropsten exists in store
connections.set("ropsten", new Connection({ provider: newProvider })); // pass a Connection
connections.set("ropsten", newProvider); // or pass a provider directly

// get a Connection and call setProvider to avoid constructing a new Connection instance
connections.get("rinkeby").setProvider(newProvider);
```

Feedback is welcome!


packages/js/plugins/ethereum/src/Connections.ts
```diff
@@ -0,0 +1,125 @@
...
+import { getNetwork } from "@ethersproject/providers";
+
+export interface ConnectionsConfig {
+  networks: {
+    [network: string]: Connection;
+  };
+  defaultNetwork?: string;
+}
+
+// https://www.typescriptlang.org/play?pretty=true&ts=4.7.4#code/LAKAxgNghgzjAEBJAdgEwKYA92oAoCcB7ABwGUoBbYideeAb1DoG1lL0AueGAF3wEtkAc3gAfbgE8KAI0IQAulyjIJAblBN4xAQDcoPWr338w8QRmyoAEstQ18XAoUwSbaewB4UFnE7KVqdAA+eABeBk06IXQeAAoeKHxongAaLSJidHweCQBKCJA6Ivh8GIBXfGR4BKSY5m0SLJz5dUKigF8UyO4Y+MTktIbM7Ik0vQgy9FyuWTl0ZQLi4prk+oymiXkw+HHJ1qWi0p4Kqr497vbNdtbNMEJkXnwysB5CfFj8xjbio5P4ZHQAHd4E4XPEABb8GBpbxYXwZchUGgAOnMcOstnsuX28EuIE0pSgqFhlj821iQy4j0EInEMCksgg03gygkYRCXwOv0q1UhMDWJBaVw0IDxIr0+Gq6F42wBwJJ8JIiMCHxuIAMvGYACIhgBhMq8QgULVbcJa3bofWG41qu4PObIiCEITxaU8bV6g2vY3ybHwAD0-vg5qgE0tXqNWtAdpgDqdLo1PGRhOJaHRfliOoyVu9WtyfsDwYtOcj0fusZR8ddMsERmQYHQhAAZkg06SEQEaAWg2d0GX7ZXnbEAPLSABW6BeyIA1ugJDBqzx86oA0HtSmFXgMlq0lmSCWffBQEA
```
dOrgJelli: can this be removed?

packages/js/plugins/ethereum/src/Connections.ts
```diff
@@ -0,0 +1,125 @@
...
+  /** Returns Connection indexed by network name, or by default network if key is undefined */
+  get(network?: string): Connection {
+    if (!network) {
+      return this._connections[this._defaultNetwork];
+    }
+    return this._connections[network];
+  }
+
+  /** sets Connection to index of network name */
+  set(network: string, connection: Connection): void {
```
dOrgJelli: I'm a bit concerned with how we don't have much testing that exhausts these new functions. Could we add a test file for connections + connection?

dOrgJelli (changes requested):

packages/js/plugins/ethereum/src/Connections.ts
krisbitney: Yes, of course. That was just a note for myself that I forgot to delete. It was related to something I was thinking of doing originally, before I realized it wouldn't work out.

packages/js/plugins/ethereum/src/Connections.ts
krisbitney: Yes, of course. I'll update with tests asap.

krisbitney:
Added support for passing providers directly into `connections.set`, to match the original spec:
```typescript
connections.set("ropsten", newProvider);
```

packages/js/plugins/ethereum/src/Connections.ts
krisbitney: I've added tests for Connections and Connection

dOrgJelli (approved):
:fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire:

PR #1154, opened 2022-08-17
replace Option with Box | null in wasm-as

### LINKED ISSUE #1132: replace Option with Box | null in wasm
Optional base types are currently implemented as an Option container in `wasm-as` and `schema/bind`.
But to omit optional fields from the function parameters we need a `| null` union type.
So the user has to wrap Int in a reference type to be able to omit it, like `Number | null`.

```graphql
# optional fields are `Number | null` instead of `Option<i32>`
type Number {
    value: Int!
}
type Module {
   # opt requires Option.None<i32> if omitted
   foo(opt: Int): Boolean
   # opt can be omitted
   bar(opt: Number): Boolean
}
```
Option was introduced in #944, but now it seems more important to omit function parameters than to provide nice Optional API, like "isSome()". We should replace `Option<i32>` with `Box<i32> | null`.

fetsorn:
tests fail in @polywrap/templates. There, `npx polywrap build` says that `wasm-as` doesn't export the new type. It seems to be using an unchanged version of `@polywrap/wasm-as`. I verified that the same wrapper builds correctly elsewhere.
resolves #1132.


krisbitney (changes requested):
I'm working on debugging it but I want to fix the export issue. Everything looks right. I'm not sure why it's not working.

First, I misspoke when I was saying templates might not build/test correctly. I was talking about their usage in the CLI create command, which downloads the templates from npm. The `@polywrap/templates` package itself should build and pass tests. Sorry about that.

krisbitney:
The AssemblyScript templates do pass tests once we add this to `polywrap.build.yaml`:

```yaml
linked_packages:
  - name: "@polywrap/wasm-as"
    path: ../../../wasm/as
```

dOrgJelli (dismissed):

dOrgJelli (approved):

PR #1158, opened 2022-08-18
Remove sha3, uts46,and graph-node plugins from default client config

krisbitney:
This PR removes the sha3, uts46, and graph-node plugins from the default client configuration. It redirects the plugin URIs to the sha3, uts46, and graph-node Wasm wrappers that have been added to the integrations repo.

I updated the ENS Wasm wrapper in the integrations repo to Origin 0.3.0 and replaced the embedded ENS wrapper in the test-env-js package with the updated build. A PR is up to merge the ENS wrapper update in the integrations repo: https://github.com/polywrap/integrations/pull/112

Also related is a PR to add the sha3, uts46, and graph-node plugins to the integrations repo: https://github.com/polywrap/integrations/pull/110

Closes https://github.com/polywrap/monorepo/issues/1125.


packages/js/client/src/default-client-config.ts
```diff
@@ -134,3 +137,9 @@ export const defaultIpfsProviders = [
   "https://ipfs.wrappers.io",
   "https://ipfs.io",
 ];
+
+export const defaultWrappers = {
+  sha3: "wrap://ipfs/QmYFWh4D91sAiYKf8o37EravLwFKAtUAmut4Xnevnt4QZR",
```
dOrgJelli: Can we use an ENS domain here instead? This will ensure it's always kept online by https://wrappers.io

packages/js/client/src/default-client-config.ts
```diff
@@ -29,8 +26,28 @@ export const getDefaultClientConfig = Tracer.traceFunc(
...
+          uri: new Uri(defaultWrappers.graphNode),
+          env: {
+            provider: "https://api.thegraph.com",
+          },
+        },
+      ],
+      redirects: [
+        {
+          from: new Uri("wrap://ens/sha3.polywrap.eth"),
+          to: new Uri(defaultWrappers.sha3),
```
dOrgJelli: @nerfZael would it be possible in the new URI resolution system to registry a redirect w/ a deprecation warning? This way when invocations are made to the old URI, the developer would know that the old URI is being phased out.

dOrgJelli (changes requested):

packages/js/client/src/default-client-config.ts
krisbitney: Okay, I added the ENS URIs.

```
export const defaultWrappers = {
  sha3: "wrap://ens/goerli/sha3.wrappers.eth",
  uts46: "wrap://ens/goerli/uts46.wrappers.eth",
  graphNode: "wrap://ens/goerli/graph-node.wrappers.eth",
};
```

To improve resolution performance, I added a goerli network using the default infura key we use for mainnet in the etherum plugin config:

```typescript
networks: {
  mainnet: {
    provider:
      "https://mainnet.infura.io/v3/b00b2c2cc09c487685e9fb061256d6a6",
  },
  goerli: {
    provider:
      "https://goerli.infura.io/v3/b00b2c2cc09c487685e9fb061256d6a6",
  },
},
```

The resolution performance without it isn't very good. That's why I used the IPFS hashes. Let me know what you think! And thanks.

packages/js/client/src/default-client-config.ts
nerfZael: Yea, it'd be simple to create a URI resolver that inherits from redirect resolver and console logs a message when resolving.
I'm not sure it'd be THAT useful since I think we should migrate to explicit Client configs instead of the current default/implicit one

nerfZael (dismissed):

packages/js/client/src/default-client-config.ts
cbrzn: I went ahead and added the `goerli` connection into the [Ethereum plugin constructor](https://github.com/polywrap/toolchain/pull/1158/commits/953ce3cc29c61e2a8d6b04199b69cea682e398fd#diff-4f5a730d3d49a8d04f4ee9864b4194af5dcc184be182bb5d1e1d033995ea4f19R32-R36) so we guarantee that the user always has the connection to goerli.

The reason for this is that if a user creates a new instance of the client and passes an Ethereum plugin, it will overwrite the default configuration, removing the goerli connection, hence, not being able to achieve the new wrappers deployed to goerli with ENS Uri resolver

I also think that we should stop setting the configuration in client instantiation (i.e: [in test env](https://github.com/polywrap/toolchain/blob/953ce3cc29c61e2a8d6b04199b69cea682e398fd/packages/js/test-env/src/index.ts#L271-L281)) and just use the config builder, this way things would be cleaner, but this is another topic :yum:

cbrzn (dismissed):

packages/js/client/src/default-client-config.ts
cbrzn: > I'm not sure it'd be THAT useful since I think we should migrate to explicit Client configs instead of the current default/implicit one

Yes! Agreed

dOrgJelli (dismissed):

PR #1167, opened 2022-08-20
chore: manually generated folders removed

cbrzn:


krisbitney (changes requested):
Thank you x1000 for looking at this. Your fix resolves the bug. MVP!

There is an additional bug that is preventing generated Rust bindings from compiling, which I didn't see at first because the build stopped when it encountered the schema issue that this PR fixes. At first I requested additional changes to fix the issue because I didn't know the additional bug was unrelated. But as I was fixing it, I realized it's a totally unrelated issue to the fix in your PR. I created a new PR to patch the additional issue: https://github.com/polywrap/monorepo/pull/1168

krisbitney:
AssemblyScript bindings correctly initialize the value to `null`:

```typescript
let _headers: Map<string, string> | null = null;
let _urlParams: Map<string, string> | null = null;
```

krisbitney (approved):

dOrgJelli (approved):

PR #1168, opened 2022-08-21
fix: imported optional Map type bindings in Rust

krisbitney:
There are two issues affecting imported Map type bindings from compiling in Rust.

**First**
```rust
// expected type option, got map
let mut _headers: Option<Map<String, String>> = Map::<String, String>::new();
let mut _url_params: Option<Map<String, String>> = Map::<String, String>::new();
```

**Second**
In the object serialization logic for optional Map type, we are trying to pass a String instead of a Result<String, Error>. Example:
```
"headers" => {
                reader.context().push(&field, "Option<Map<String, String>>", "type found, reading property");
                _headers = reader.read_optional_ext_generic_map(|reader| {
                    reader.read_string()? //  <---- This `?` should not be here!
                }, |reader| {
                    reader.read_string()
                })?; 
                reader.context().pop();
            }
```

### DIFF
--- a/packages/schema/bind/src/bindings/rust/functions.ts
+++ b/packages/schema/bind/src/bindings/rust/functions.ts
@@ -164,7 +164,7 @@ export const toWasmInit: MustacheFunction = () => {
         .substring(openBracketIdx + 1, closeBracketIdx)
         .split(",")
         .map((x) => toWasm()(x.trim(), render));
-      return `Map::<${key}, ${value}>::new()`;
+      return optionalModifier(`Map::<${key}, ${value}>::new()`);
     }
 
     switch (type) {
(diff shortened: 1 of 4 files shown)


cbrzn:
nice catch ser :pray:, I previously fixed this but only in object types (https://github.com/polywrap/monorepo/pull/1142) and not for imported object types. I guess you also need to update `schema/bind` test and we are ready to go

dOrgJelli (dismissed):
Nice catch! Just need tests to go green

krisbitney:
> nice catch ser 🙏, I previously fixed this but only in object types (#1142) and not for imported object types. I guess you also need to update `schema/bind` test and we are ready to go

Thanks. 😄

krisbitney:
> Nice catch! Just need tests to go green

Thanks! 😄

cbrzn (approved):

Niraj-Kamdar (approved):
@krisbitney Is it also fixed for the methodTypes?

krisbitney:
> @krisbitney Is it also fixed for the methodTypes?

I fixed both the issues mentioned in the PR description. Are you referring to one of those or something else?

cbrzn:
> > @krisbitney Is it also fixed for the methodTypes?
> 
> I fixed both the issues mentioned in the PR description. Are you referring to one of those or something else?

Just checked it and we are good to go ser

PR #1169, opened 2022-08-22
URI resolution refactor

nerfZael:
# Changes and Improvements
- Core of URI resolution is in the core-js package while uri-resolvers and uri-resolver-extensions are separate
- @polywrap/result - Package for the result object to be used by other packages
- Single URI resolver in client greatly simplifies the client and core part of URI resolution
- Nested URI resolvers
  - Enable more flexibility in configuring the resolution process
  - Aggregator resolver responsible for handling many resolvers
- Composable URI resolvers with UriResolverLike
  - Can turn different objects into resolvers (like redirects, plugins, wrappers, resolver arrays, etc)
- URI resolution context tracks the currently resolved URIs
  - Enables infinite loop protection when resolution process reenters the client (through the use of ResolverWithLoopGuard or RecursiveResolver)
  - Enables dynamic resolve-time loading of resolver extensions
  - Tracks URI resolution path
- ENV is calculated invoke time from the URI resolution path in the client (instead of in resolvers)
- URI history can be fetched from the resolution context instead of being directly returned
  - This simplifies the return type and makes history optional
- Nested history (even with resolver extension sub-invokes)
- Inner errors are propagated up
- Resolvers (and the resolution process) can return URIs, packages or wrappers
  - This allows for creating new wrappers from packages every time a URI is resolved (e.g. stateless plugins)
  - It also allows pre-creating a wrapper and having a permanent instance regardless of caching (e.g. not losing plugin state if cache is reset)
- Function to build clean URI history to easily interpret it
- Function to get URI resolution path to only show the parts of history which were directly involved in resolution
- Caching
  - URI resolution path and wrapper cache removed from the resolver tryResolverUri method
  - Cache is per resolver, not client
    - Main cachable resolver still mimicks legacy client cache
    - Can create cache resolvers to cache resolver results
    - Can create caches for
      - Whole response object
      - Only the result object (no history)
      - Only uris, packages and wrappers (no error)
      - Only uris
      - Only packages
      - Only wrappers
      - Convert packages to wrappers and cache them (resolver wil then always return uri or wrapper)
        - This process mimicks the legacy client cache and is implemented in PackageToWrapperCacheResolver
- Removed name from IUriResolver
  - Can use URI history step description to accomplish the same thing in a more detailed way
- WASM changes
  - @polywrap/wasm-js (WasmWrapper.ts and other Wasm relates stuff) is a separate package
  - URI and ENV is passed to the wrapper invoke-time
    - Allows ENV to be changed via reference 
    - Wrapper instance is decoupled from URI and ENV
      - This enables things like reusing a wrapper with different URIs or in different clients
  - Can pass ENV override to the invoke function
  - Resolver URI is no longer a dependency of WasmWrapper, replaced with IFileReader
    - This simplifies the creation of a WasmWrapper
  - WasmPackage and InMemoryFileReader
    - These enable embedded wrappers (though not really usable right now unless the user overrides the client resolver)
      - This will be improved in a follow-up PR


packages/js/client/src/PolywrapClient.ts
```diff
@@ -633,52 +615,59 @@ export class PolywrapClient implements Client {
...
     uri: Uri,
+    resolutionContext?: IUriResolutionContext,
     options?: Contextualized
   ): Promise<Wrapper> {
     Tracer.setAttribute("label", `Wrapper loaded: ${uri}`, TracingLevel.High);
 
-    const { wrapper, uriHistory, error } = await this.resolveUri(uri, {
+    if (!resolutionContext) {
+      resolutionContext = new UriResolutionContext();
+    }
```
cbrzn, line 640: Do we need this validation since the same logic is applied in the `tryResolveUri` method?

packages/js/client/src/PolywrapClient.ts
```diff
@@ -633,52 +615,59 @@ export class PolywrapClient implements Client {
...
+      packageOrWrapper = uriPackageOrWrapper.wrapper;
+    }
+
+    const client = contextualizeClient(this, options?.contextId);
+
+    const resolutionPath: Uri[] = resolutionContext.getResolutionPath();
+
+    const wrapper = await initWrapper(packageOrWrapper, client, resolutionPath);
+
     return wrapper;
```
cbrzn: nit:
```suggestion
    return await initWrapper(packageOrWrapper, client, resolutionPath);
```

packages/js/client/src/__tests__/core/uri-resolution/uri-resolution.spec.ts
```diff
@@ -0,0 +1,440 @@
...
+    `${GetPathToTestWrappers()}/wasm-as`,
+    "$root-wrapper-dir"
+  );
+
+  expect(receivedCleanHistory).toEqual(
+    JSON.stringify(JSON.parse(expectedCleanHistory), null, 2)
+  );
+};
+
+const expectPackageWithHistory = async (
```
cbrzn: This test case is not implemented - If this is expected I'd say we should remove this

dOrgJelli: This function is not being used anywhere, do we need it?

dOrgJelli (dismissed):

dOrgJelli (dismissed):

packages/js/client/src/__tests__/core/uri-resolution/uri-resolution.spec.ts
nerfZael: Not anymore, yes in the future, but I'll remove it for now

packages/js/result/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@polywrap/result",
```
cbrzn, line 2: I noticed that all the packages that import this package as a dependency, also import `core`. Would it make sense this package there? :thinking:

packages/js/uri-resolver-extensions/src/ExtendableUriResolver.ts
```diff
@@ -0,0 +1,68 @@
...
+  UriPackageOrWrapper,
+  UriResolutionResult,
+} from "@polywrap/core-js";
+import { Result, ResultOk } from "@polywrap/result";
+import { UriResolverAggregatorBase } from "@polywrap/uri-resolvers-js";
+
+export class ExtendableUriResolver extends UriResolverAggregatorBase<unknown> {
+  loadingExtensionsMap: Map<number, Map<string, boolean>> = new Map();
+
+  private resolverName: string;
```
cbrzn: nit:
```suggestion
  private readonly resolverName: string;
```

packages/js/uri-resolver-extensions/src/ExtendableUriResolver.ts
```diff
@@ -0,0 +1,68 @@
...
+  coreInterfaceUris,
+  IUriResolutionContext,
+  UriPackageOrWrapper,
+  UriResolutionResult,
+} from "@polywrap/core-js";
+import { Result, ResultOk } from "@polywrap/result";
+import { UriResolverAggregatorBase } from "@polywrap/uri-resolvers-js";
+
+export class ExtendableUriResolver extends UriResolverAggregatorBase<unknown> {
+  loadingExtensionsMap: Map<number, Map<string, boolean>> = new Map();
```
cbrzn: Are we using this variable? :thinking:

packages/js/uri-resolver-extensions/src/UriResolverWrapper.ts
```diff
@@ -0,0 +1,103 @@
...
+import { WasmPackage } from "@polywrap/wasm-js";
+import { ResolverWithHistory } from "@polywrap/uri-resolvers-js";
+
+export class UriResolverWrapper extends ResolverWithHistory<unknown> {
+  constructor(public readonly implementationUri: Uri) {
+    super();
+  }
+
+  protected getStepDescription = (): string =>
+    `ResovlerExtension (${this.implementationUri.uri})`;
```
cbrzn:
```suggestion
    `ResolverExtension (${this.implementationUri.uri})`;
```

packages/js/uri-resolvers/jest.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
cbrzn, line 1: Since we don't really have tests on this package, would it make sense to remove this file?

packages/js/uri-resolvers/src/aggregator/UriResolverAggregatorBase.ts
```diff
@@ -0,0 +1,95 @@
...
+      const typeResolver = resolver as IUriResolver<TResolutionError>;
+
+      const result = await typeResolver.tryResolveUri(uri, client, subContext);
+
+      if (
+        result.ok &&
+        result.value.type === "uri" &&
+        result.value.uri.uri === uri.uri
+      ) {
+        continue;
```
cbrzn: nit: this continue is unnecessary

packages/js/uri-resolvers/src/cache/PackageToWrapperCacheResolver.ts
```diff
@@ -0,0 +1,106 @@
...
+  executeMaybeAsyncFunction,
+  Wrapper,
+  IUriResolutionContext,
+  UriPackageOrWrapper,
+  UriResolutionResult,
+} from "@polywrap/core-js";
+import { DeserializeManifestOptions } from "@polywrap/wrap-manifest-types-js";
+import { Result } from "@polywrap/result";
+
+export type ResolutionCallback<TError> = (
```
cbrzn: Looks like this is not being used

packages/js/uri-resolvers/src/helpers/ResolverWithLoopGuard.ts
```diff
@@ -0,0 +1,46 @@
...
+import {
+  IUriResolver,
+  Uri,
+  Client,
+  IUriResolutionContext,
+  UriPackageOrWrapper,
+  UriResolutionResult,
+} from "@polywrap/core-js";
+
+export class ResolverWithLoopGuard<TError = undefined>
```
cbrzn, line 15: I think this file is not being used but I might be wrong :thinking:

packages/js/uri-resolvers/src/utils/buildUriResolver.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Result } from "@polywrap/result";
+import {
+  IUriResolver,
+  Uri,
+  Client,
+  IWrapPackage,
+  IUriPackage,
+} from "@polywrap/core-js";
+
+export const buildUriResolver = <TError = undefined>(
```
cbrzn, line 14: real nit please feel free to ignore: since all the conditionals are based on `resolvable` variable, it might make sense to change this to a switch statement to make the code easier to read

packages/js/uri-resolvers/src/utils/getUriResolutionPath.ts
```diff
@@ -0,0 +1,42 @@
...
+        if (!x.result.ok) {
+          return true;
+        }
+
+        const uriPackageOrWrapper = x.result.value;
+
+        if (uriPackageOrWrapper.type === "uri") {
+          // If a redirect has occured
+          return uriPackageOrWrapper.uri.uri !== x.sourceUri.uri;
+        } else if (uriPackageOrWrapper.type === "package") {
```
cbrzn, line 20: nit: I am noticing that we are using a lot the strings `package` and `wrapper` - Would it make sense to create an `Enum` somewhere and just use the values of the Enum? Making the code easier to maintain

packages/js/wasm/src/WasmWrapper.ts
```diff
@@ -163,13 +119,17 @@ export class WasmWrapper extends Wrapper {
         },
         invokeResult: {} as InvokeResult,
         method,
-        args: isBuffer(args) ? args : msgpackEncode(args),
-        env: msgpackEncode(this._getClientEnv()),
+        args: args
+          ? isBuffer(args)
+            ? args
+            : msgpackEncode(args)
+          : EMPTY_ENCODED_OBJECT,
```
cbrzn, line 126: nit: I would convert this to a variable, making this line more readable

packages/js/client/src/__tests__/core/uri-resolution/uri-resolution.spec.ts
cbrzn: This is outdated (saw this yesterday) so I am resolving

packages/js/client/src/PolywrapClient.ts
nerfZael: Yes, since the context is an optional argument (and makes sense to keep it that way) and we need it to get history to show in the error message in case of an error

packages/js/result/package.json
nerfZael: No, the result package is useful on it's own even without the core. It could be replaced by another third party if we wished (there is a similar one which this one is based of off)

packages/js/uri-resolvers/src/helpers/ResolverWithLoopGuard.ts
nerfZael: It's not, but it's a useful primitive for users :)

packages/js/uri-resolvers/src/cache/PackageToWrapperCacheResolver.ts
nerfZael: Nice catch!

packages/js/uri-resolvers/src/utils/buildUriResolver.ts
nerfZael: I don't think this could be rewritten as a switch statement, but if you have an idea, please share!

packages/js/uri-resolvers/src/utils/getUriResolutionPath.ts
nerfZael: If it was another language then I'd say yes, definitely. But I find that in TypeScript string unions (not plain strings) are preferable since they don't require another enum type, don't require importing that type, are more readable, can be printed out or logged, and (unlike in other languages) are type safe

packages/js/wasm/src/WasmWrapper.ts
nerfZael: I find this really readable (though if you're not used to ternary operators I can see how it wouldn't be)
If you have a more readable code snippet pls post it :)

packages/js/uri-resolvers/jest.config.js
dOrgJelli: Let's keep it, since we'll be adding tests in the near future :)

packages/js/uri-resolvers/src/aggregator/UriResolverAggregatorBase.ts
dOrgJelli: great catch 🙏

packages/js/uri-resolvers/src/utils/buildUriResolver.ts
dOrgJelli: yah no this can't be written as a switch :P

packages/js/uri-resolvers/src/utils/getUriResolutionPath.ts
dOrgJelli: Yah I agree, enums are a bit tricky in TypeScript. Here's an article that might cover some of the erroneous cases (didn't read myself): https://thoughtbot.com/blog/the-trouble-with-typescript-enums

PR #1189, opened 2022-08-26
schema.graphql should be replaced with wrap.info in the polywrap.app.yaml migrator

### LINKED ISSUE #1185: Handle wrap.info in the polywrap.app.yaml migrator
**Describe the bug**
`polywrap.app.yaml` manifest migrates incorrectly from 0.1.0 to 0.2.0. 
Currently it assigns `0.1.0/import_redirects.schema` to `0.2.0/import_abis.abi`. 
The first used to be `schema: path/to/build/schema.graphql`, so the migration will be `abi: path/to/build/schema.graphql`.
But the latter is expected to be `abi: path/to/build/wrap.info`.

https://github.com/polywrap/toolchain/blob/0a1e6e8e1248dfe3708b511b0e8d26c98ba2e4cf/packages/js/manifests/polywrap/src/formats/polywrap.app/migrators/0.1.0_to_0.2.0.ts#L16

 - go to `demos/simple-storage/wrapper/assemblyscript` 
 - update to `polywrap^0.4.1`
 - run `npx polywrap app codegen -m ./src/__tests__/types/polywrap.app.yaml`

**Expected behavior**
`polywrap.app.yaml` migrator should account for the change from `schema.graphql` to `wrap.info` in `import_abis`.
There should also be tests for the .app migrator in `@polywrap/polywrap-manifest-types-js` so we can catch this.

pileks:
Resolves #1185 

We might want to consider checking for the existence of `schema.graphql` before blindly migrating the `import_redirects` to `wrap.info`. This would require the migrator to "know" about the manifest's path.

Here's some context to my reasoning for this (copied from Discord):

> This is an interesting edge-case here, as the redirects work under the assumption that the file will always exist, which clearly isn't the case when we point to a file within the build folder.
> 
> So the migrator assumes that schema.graphql exists when migrating  import_redirects to import_abis, and I believe that to be the correct 1st step. 
> 
> What should probably happen in the case of that bug would be that the migrator searches for wrap.info within the same folder if the original file could not be found.
> 
> I'd need a bit more context on WHEN we use import_redirects, as my assumption was that we point it to a local file (and that file doesn't necessarily have to be part of a build output).
> 
> In any case I believe this solution would allow for flexibility in this case where we had 2 significant changes happen accross the repo (1st migration to manifest format 0.2, 2nd the introduction of wrap.info and depreciation of schema.graphql for build outputs)

Thoughts? @dOrgJelli @namesty @fetsorn ?

### DIFF
--- a/packages/js/manifests/polywrap/src/__tests__/manifest/polywrap/migrations/polywrap-0.1.0.yaml
+++ b/packages/js/manifests/polywrap/src/__tests__/manifest/polywrap/migrations/polywrap-0.1.0.yaml
@@ -8,4 +8,4 @@ meta: ./polywrap.meta.yaml
 deploy: ./polywrap.deploy.yaml
 import_redirects:
   - uri: "ens/ipfs.polywrap.eth"
-    schema: ../../../../../../../../interfaces/ipfs/build/schema.graphql
+    schema: ../../schema.graphql
--- a/packages/js/manifests/polywrap/src/__tests__/manifest/polywrap/migrations/polywrap-0.2.0.yaml
+++ b/packages/js/manifests/polywrap/src/__tests__/manifest/polywrap/migrations/polywrap-0.2.0.yaml
@@ -7,7 +7,7 @@ source:
   module: ../index.ts
   import_abis:
     - uri: "ens/ipfs.polywrap.eth"
-      abi: ../../../../../../../../interfaces/ipfs/build/schema.graphql
+      abi: ../../wrap.info
 extensions:
   build: ./polywrap.build.yaml
   meta: ./polywrap.meta.yaml
--- a/packages/js/manifests/polywrap/src/formats/polywrap.app/migrators/0.1.0_to_0.2.0.ts
+++ b/packages/js/manifests/polywrap/src/formats/polywrap.app/migrators/0.1.0_to_0.2.0.ts
@@ -1,6 +1,8 @@
 import { AppManifest as OldManifest } from "../0.1.0";
 import { AppManifest as NewManifest } from "../0.2.0";
 
+import path from "path";
+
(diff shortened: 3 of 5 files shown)


dOrgJelli (changes requested):
Hey @pileks, while this is an interesting "fix", I think the real problem / solution is a bit different than what is described (cc @fetsorn). Let me explain.

There are 2 changes that need to be considered.

First, we removed `schema.graphql` from the build artifacts, and replaced it with `wrap.info`. This **is a breaking change** for any code that assumed the existence of `build/schema.graphql`.

Second, we renamed `import_redirects` to `import_abis`, and now support the following `abi: ...` formats: `schema.graphql`, `wrap.info`, `abi.json`, `abi.yaml`

So, when considering the problem reported here https://github.com/polywrap/toolchain/issues/1185 , the problem is actually do the the **first** change described above, which is breaking.

The proper fix is **for the user** to move away from using `build/schema.graphql`.

We could add intelligent migration logic where `if (schemaPath.contains("build/schema.graphql"))` then `replace("schema.graphql", "wrap.info")`, but I do not think this is necessarily the right approach. Instead we should probably have a CLI driven project migrator that can guide users through breaking changes like this, and output the upgraded manifests.

So, to summary, I'm of the opinion we should:
- close the issue https://github.com/polywrap/toolchain/issues/1185
- manually fix the problem within the project in question
- spec out a CLI driven project migrator?

fetsorn:
> we renamed `import_redirects` to `import_abis`, and now support the following `abi: ...` formats: `schema.graphql`, `wrap.info`, `abi.json`, `abi.yaml`

Ah, so the existing migration code works as intended when `schema.graphql` is outside the build folder. And we just need better warnings for other cases, aka cli migrator. Fair enough, closing the original issue.

PR #1192, opened 2022-08-27
feat: client.reconfigure(config) + remove invoke-time configuration

dOrgJelli:


cbrzn (commented):
Would it make sense to think that previously we had the context to ensure that the client configuration was guaranteed but now that we have the config builder we don't need to "contextualize" the client anymore?

pileks:
The one thing that I'm battling with here is the `reconfigure` method name. It kind of sounds like we're reconfiguring the existing client (OO approach) instead of returning a newly configured client (functional approach).

We might want to think about naming the method to something like `getReconfigured`?

In the end, I'm nitpicking here, `reconfigure` is fine too.

nerfZael:
> The one thing that I'm battling with here is the `reconfigure` method name. It kind of sounds like we're reconfiguring the existing client (OO approach) instead of returning a newly configured client (functional approach).
> 
> We might want to think about naming the method to something like `getReconfigured`?
> 
> In the end, I'm nitpicking here, `reconfigure` is fine too.

What about `clone`? That's technically what we're doing here. But I'm also more in favor of `configure` than `reconfigure` or 'getReconfigured' (I think the return type being the Client kinda hints enough that we're not reconfiguring the existing client).
IMO: `clone` > `configure` > `reconfigure` > `getReconfigured`

pileks:
> What about `clone`? That's technically what we're doing here. But I'm also more in favor of `configure` than `reconfigure` or 'getReconfigured' (I think the return type being the Client kinda hints enough that we're not reconfiguring the existing client). IMO: `clone` > `configure` > `reconfigure` > `getReconfigured`

I can get behind `clone` for sure!

dOrgJelli:
In the end @pileks @nerfZael, I decided against the `clone(...)` name because it was strange to have it return a `ClientConfigBuilder` instance instead of a new `Client`. So I still think `reconfigure(...)` makes the most sense. This is the interface:
```typescript
const newClient = new PolywrapClient(
  client
    .reconfigure({ ...partial_config })
    .addPlugin({...})
    .build()
);
```

packages/js/client/src/__tests__/core/uri-resolution/uri-resolution.spec.ts
```diff
@@ -360,23 +360,22 @@ describe("URI resolution", () => {
     const fromUri = new Uri(`ens/from.eth`);
     const redirectUri = new Uri(`ens/to.eth`);
 
-    const client = await getClient();
+    const client = await getClient({
```
nerfZael, line 363: This is an invoke-time reconfiguring test, so we should rename the test and test reconfigure here

pileks:
> In the end @pileks @nerfZael, I decided against the `clone(...)` name because it was strange to have it return a `ClientConfigBuilder` instance instead of a new `Client`. So I still think `reconfigure(...)` makes the most sense. This is the interface:
> 
> ```ts
> const newClient = new PolywrapClient(
>   client
>     .reconfigure({ ...partial_config })
>     .addPlugin({...})
>     .build()
> );
> ```

This approach seems neat, though my ONE nitpick would be that the Client has to know about the ClientConfigBuilder, which I was kind of hoping to separate in the future. TBH, it would be ideal that the only common ground between them be a ClientConfig type found in Core.
This way, the Client is completely agnostic of its default configuration, plugins, etc, and never makes assumptions about how a config should be built. This would also entail removing use of the Client Config Builder from the Client's constructor, and removing the `noDefaults` option, which would essentially always be true with a "purer" version of the Client.

Obviously, this is not within the scope of this PR, I just wanted to make note of this, as I feel it would best fit the boundaries of what the Client should be doing.

packages/js/client/src/__tests__/core/uri-resolution/uri-resolution.spec.ts
pileks: I'm not too keen on the whole `getClient(Partial<ClientConfig>)` thing either... Probably something we should work to remove now that we have the ClientConfigBuilder + reconfigure.

pileks (dismissed):
Other than my non-blocking thoughts, LGTM!

Issue #1193, opened 2022-08-29
fix logic for msgpack-js in case of nested map 


Niraj-Kamdar:
Currently nested msgpack deserializes as follows: 
```
    ExtData {
      type: 1,
      data: Uint8Array(11) [
        129, 163, 117, 115, 100,
        165,  50,  48,  50,  52,
         53
      ]
    }
```

Which is inaccurate, have a better way to deserializes the value of the map in msgpack-js

krisbitney:
Resolved by https://github.com/polywrap/toolchain/pull/1194

PR #1194, opened 2022-08-29
fix(msgpack-js): nested map should de/serialize properly on client side

Niraj-Kamdar:
- Closes: #1193

### DIFF
--- a/packages/js/msgpack/src/__tests__/msgpack.spec.ts
+++ b/packages/js/msgpack/src/__tests__/msgpack.spec.ts
@@ -26,8 +26,6 @@ describe("msgpack", () => {
     customMap.set("secondKey", "secondValue")
 
     const encoded = msgpackEncode(customMap)
-    // [199, 43, 1] are being added because of the map structure
-    expect(encoded).toEqual(Uint8Array.from([ 199, 43, 1, ...expectedArrayLike]))
     const decoded = msgpackDecode(encoded)
     expect(decoded).toEqual(customMap);
   });
@@ -39,8 +37,6 @@ describe("msgpack", () => {
     customMap.set("secondKey", new Map([["second", "2"]]));
 
     const encoded = msgpackEncode(customMap)
-    // [199, 43, 1] are being added because of the map structure
-    expect(encoded).toEqual(Uint8Array.from([ ...expectedArrayLike]))
     const decoded = msgpackDecode(encoded)
     expect(decoded).toEqual(customMap);
   });
--- a/packages/js/msgpack/src/__tests__/msgpack.spec.ts
+++ b/packages/js/msgpack/src/__tests__/msgpack.spec.ts
@@ -9,7 +9,7 @@ describe("msgpack", () => {
     115, 101,  99, 111, 110, 100,  86,  97,
     108, 117, 101
   ]
-  it("Should encode and decode, returning the same object", () => {
+  it("Should encode and decode object", () => {
     const customObject = {
       "firstKey": "firstValue",
       "secondKey": "secondValue",
@@ -20,7 +20,7 @@ describe("msgpack", () => {
     expect(decoded).toEqual(customObject);
   });
 
-  it("Should encode and decode, returning the same map", () => {
+  it("Should encode and decode map", () => {
     const customMap = new Map()
     customMap.set("firstKey", "firstValue")
     customMap.set("secondKey", "secondValue")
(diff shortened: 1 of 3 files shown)


packages/js/msgpack/src/index.ts
```diff
@@ -20,16 +22,23 @@ extensionCodec.register({
...
     }
   },
   decode: (data: Uint8Array) => {
-    const map = decode(data) as Record<string | number, unknown>;
-    return new Map(Object.entries(map));
+    const obj = msgpackDecode(data) as Record<string | number, unknown>;
+    const map = new Map();
+
+    for (const [key, value] of Object.entries(obj)) {
+      if (isBuffer(value)) {
```
nerfZael: What if it's a key => buffer map?

packages/js/msgpack/src/__tests__/msgpack.spec.ts
```diff
@@ -26,8 +26,17 @@ describe("msgpack", () => {
     customMap.set("secondKey", "secondValue")
 
     const encoded = msgpackEncode(customMap)
-    // [199, 43, 1] are being added because of the map structure
-    expect(encoded).toEqual(Uint8Array.from([ 199, 43, 1, ...expectedArrayLike]))
+    const decoded = msgpackDecode(encoded)
+    expect(decoded).toEqual(customMap);
+  });
+
+  it("Should encode and decode nested map, returning the same map", () => {
```
nerfZael: I think `returning the same map` is implied

packages/js/msgpack/src/__tests__/msgpack.spec.ts
```diff
@@ -26,8 +26,17 @@ describe("msgpack", () => {
...
+    const customMap: Map<string, Map<string, string>> = new Map()
+    // @ts-ignore
+    customMap.set("firstKey", new Map([["one", "1"]]));
+    customMap.set("secondKey", new Map([["second", "2"]]));
+
+    const encoded = msgpackEncode(customMap)
     const decoded = msgpackDecode(encoded)
```
nerfZael, line 40: `decodedMap` instead of `decoded` ?

nerfZael, line 40: `decodedMap` instead of `decoded` ?

nerfZael, line 40: `decodedMap` instead of `decoded` ?

packages/js/msgpack/src/index.ts
Niraj-Kamdar: It will just work because it's recursive encoding/decoding. But to make it more clear i suppose I should throw an error if the value isn't buffer

packages/js/msgpack/src/__tests__/msgpack.spec.ts
Niraj-Kamdar: Agree, I just kept it as it is but renaming these test name would be better

packages/js/msgpack/src/__tests__/msgpack.spec.ts
Niraj-Kamdar: Yes we expect decodedMap but if the decoding is wrong then it might not be map.

packages/js/msgpack/src/index.ts
Niraj-Kamdar: Nevermind it was unnecessary in the first place so removed it

packages/js/msgpack/src/index.ts
```diff
@@ -1,10 +1,12 @@
-import {
-  Encoder,
-  Decoder,
-  ExtensionCodec,
-  encode,
-  decode,
-} from "@msgpack/msgpack";
+import { Encoder, Decoder, ExtensionCodec } from "@msgpack/msgpack";
+
+export function isBuffer(maybeBuf: unknown): maybeBuf is BufferSource {
```
cbrzn: Where is this function being used? We have it defined in two places (https://github.com/polywrap/toolchain/blob/origin-dev/packages/js/core/src/utils/is-buffer.ts) I think we should have it in one place

Niraj-Kamdar: Will remove it from here not needed anymore

packages/js/msgpack/src/index.ts
```diff
@@ -22,14 +16,19 @@ extensionCodec.register({
       for (const [key, value] of object) {
         optimized[key] = value;
       }
-      return encode(optimized);
+      return msgpackEncode(optimized);
```
cbrzn, line 19: So now you're calling `msgpackEncode` instead of calling this function recursively, right? Seems simpler than what you thought at first :P

cbrzn (approved):
lgtm :green_apple:

dOrgJelli (approved):

packages/js/msgpack/src/index.ts
Niraj-Kamdar: It just worked out haha

Issue #985, opened 2022-06-28
JavaScript minimization (aka "minification")


krisbitney:
JavaScript minimization (aka "minification") issues have been encountered in both the Uniswap v3 front end demo and the Polywrap Hub. For example, in the Polywrap Hub, class names were being erased during optimization and this was fixed by configuring Webpack. We should ensure our JavaScript packages comply with minimization standards/requirements so that minimization works without requiring users to modify Webpack settings.

dOrgJelli:
Can we do some research on:
- reproducing these minifcation errors (branch links w/ repro steps)
- what are some common issues with minification in production are ([removing class names](https://duckduckgo.com/?q=minification+removing+class+names&t=brave&ia=web) for example)
- where in our code are these issue stemming from?

cbrzn:
> reproducing these minifcation errors (branch links w/ repro steps)
- Go to demos repository & checkout to branch `update-hello-world-to-origin` (PR: https://github.com/polywrap/demos/pull/70)
- In `ts-vanilla` app, go to the `webpack.config.js` file and change (in line 5) mode to `production`
- Run the app and try to run the invocation - it should fail

> where in our code are these issue stemming from?

Based on @Brioline report of the bug, it looks like the `Uri` class might be a problem regarding minification. My first thought is that it is something with the `getters` but not sure...
![image](https://user-images.githubusercontent.com/21031138/179001199-7af0430e-d4fd-4148-ba2e-0cdbf29bf71f.png)

Also, it's worth mentioning that I fixed this with the keep function & class names in minification (https://github.com/polywrap/polyfolio/pull/150/files)

PR #1199, opened 2022-08-31
Fix: JavaScript minimization bug

krisbitney:
The client has a bug that prevents url resolvers from loading correctly after JavaScript minimization (aka "minification"). The bug can be circumvented by configuring weback to keep function names. 

This PR fixes the issue so that webpack configuration is not necessary. It does so by removing the use of reflection from the client and uri resolvers. In particular, the client and uri resolvers were using `X.name` where X is the name of a uri resolver class.

I'm not sure how to write an automated test for this. I tested the fix using the Uniswap v3 front end. After linking the updated toolchain packages, the Uniswap v3 front end works without modifying the webpack configuration.

Closes https://github.com/polywrap/toolchain/issues/985.


dOrgJelli (approved):
👏

PR #1201, opened 2022-09-01
Fix: Use HashMap instead of BTreeMap in wasm/rust

dOrgJelli:
closes: https://github.com/polywrap/documentation/issues/213


dOrgJelli:
`BTreeMap` is actually preferred, see comment here: https://github.com/polywrap/documentation/issues/213#issuecomment-1233652757

PR #1204, opened 2022-09-02
Build pipeline refactor: Image, Local and VM strategies

### LINKED ISSUE #1246: Implement and benchmark a "Docker VM" build approach

namesty:
The current PR closes #1246 and #1119.

# Overview

It introduces a refactor to the build pipeline composed of 3 main changes:

## Strategies

Implementation of different build "strategies", which are abstractions of different source building algorithms. This also implied a decoupling of Docker from the `Compiler` and `Project`. Currently there are 3, and "VM" is the new default:

- `Image`: default approach, prior to this PR. It implies building a Docker image at runtime, where dependencies are installed and sources are copied and built as `Dockerfile` instructions. On subsequent builds, Docker tries to reuse cached image layers and rebuild accordingly. This approach is notably slow but the complete process happens in Docker, and can be reproduced, examined and audited layer by layer (from dependency installation to build artifacts output)

- `Local`: approach that does not use Docker at all. It simply executes a `.sh` file that contains the necessary instructions to install dependencies and build sources. While this is the fastest way of building, it involves requiring the user to install "prequisite" system dependencies and given that sources are built on the host machine and not a reproducible docker environment, "reproducibility" isn't guaranteed ("but it builds on my machine tho!"). This strategy does not concern itself with dependency symbolic links, as all dependencies are handled by the host's dep manager.

- `VM`: approach that uses Docker only for the source building part of the process. At build time, it pulls a pre-built image with all necessary system dependencies, env vars and runtime; and it instantiates a Docker container with it. The Docker container instantiates bind-mounts (volumes) to copy the sources and dependencies from the host, build the sources inside the container, and copy the build artifacts back to the host machine. This approach ensures that the sources will be built in a reproducible environment but it doesn't use Docker for anything else and no image is built at runtime.

## Build Manifest Changes

The build manifest was changed to contemplate configurations for build strategies, and strategy-specific overrides/config. It also validates strategy configurations based on the Polywrap's manifest language. Example: if the project is an Assemblyscript wrapper and the image strategy is going to be used, then `node_version` is a required image-strategy config property; but it is not if the project is a Rust wrapper.

## Different handling of build cache (`.polywrap/build`) directory

Now the defaults (scripts, templates, etc.) used for any strategy are cached in the `.polywrap/build` dir for easy debugging and overriding. We no longer write a default manifest to the cache folder, only to read it later with its manifest extension.

Other misc impl details include: `Compiler` does not generate code but depends on `CodeGenerator`, compiler overrides.

# Short-term future needs

This PR introduces some short-term needs we need to cover in other PRs:

- Build benchmarks repository and automated measurements and stress testing (so far, I've been just locally testing)
- Automated pipeline to change and re-publish base VM images to DockerHub

# Further fleshing-out of VM strategy

https://hackmd.io/HQHG0xXfT4CEc-iHsuMHpg


packages/cli/src/commands/build.ts
```diff
@@ -67,43 +70,76 @@ export const build: Command = {
...
     project,
-    outputDir,
-    schemaComposer,
+    client,
   });
 
   const execute = async (): Promise<boolean> => {
-    compiler.reset();
+    project.reset();
+    schemaComposer.reset();
```
nerfZael: Can't say for sure without diving deeper, but these resets feel sketchy

packages/cli/src/lib/codegen/strategies/CompilerCodegenStrategy.ts
```diff
@@ -0,0 +1,37 @@
...
+    const abi = await this.schemaComposer.getComposedAbis();
+    const binding = await this.project.generateSchemaBindings(
+      abi,
+      generationSubPath
+    );
+
+    return writeDirectorySync(binding.outputDirAbs, binding.output);
+  }
+
+  private async _getGenerationSubpath() {
```
nerfZael: Missing return type

packages/cli/src/commands/build.ts
```diff
@@ -67,43 +70,76 @@ export const build: Command = {
...
 
-  const dockerLock = new FileLock(
-    project.getCachePath("build/DOCKER_LOCK"),
-    print.error
-  );
+  const polywrapManifest = await project.getManifest();
+  await validateManifestModules(polywrapManifest);
 
-  const schemaComposer = new SchemaComposer({
+  const buildStrategy = createSourceBuildStrategy(docker ? "docker" : "local", {
```
nerfZael: It might even be clearer if there was no `createSourceBuildStrategy` func and it's code was just here.
Wdyt?

namesty: Yep, agreed

packages/cli/src/lib/codegen/strategies/CompilerCodegenStrategy.ts
namesty: Woo, nice catch

packages/cli/src/commands/codegen.ts
```diff
@@ -82,25 +82,23 @@ async function run(options: CodegenCommandOptions) {
...
+  const strategy = script
+    ? new ScriptCodegenStrategy({
+        codegenDirAbs: codegenDir,
+        script,
+        schemaComposer,
+        project,
+        omitHeader: false,
+        mustacheView: undefined,
+      })
+    : new CompilerCodegenStrategy({
```
pileks: This is the line I was most interested in.

In my PR, it's all reduced to the following:
https://github.com/polywrap/toolchain/pull/1051/files#diff-87c9d295ac1ee17a0b50c70f922ce1ebc7ad5da4746b6d4f18b45334452ae030R106

It seems to work... 😁

namesty:
After seeing @pileks's changes, indeed the Compiler's codegen does not anything really different to the CodeGenerator's codegen. Additionally, the `generationSubpath` argument may, at some point, be used but right now it is not being used since we're overriding the default `src/wrap` with `src/wrap` if we're codegenerating for Rust. For these reasons, (and the lack of need for changing strategies at runtime) I just removed the strat pattern for codegen, in favor of simpler, classic inheritance.

I noticed we did not have sanity tests for Rust, when using the `build` command or the `codegen` command; so to really make sure this refactor does not cause regressions, I added sanity cases for Rust for both builds and codegens.

I decided to keep this PR smaller by just incluiding the "no docker" build possibility; I'll push the Docker VM approach in a different PR.

packages/cli/src/commands/build.ts
namesty: They were, and were serving no purpose; just removed them

packages/cli/src/commands/codegen.ts
namesty: Yep, you're right. In fact, between your PR and this one; I think we're removing significant dead code

namesty:
Poured my performance analysis on this PR here: https://hackmd.io/IVTtH4rTSa6ud5XwIUkCDw

namesty:
Closes #1119

packages/cli/src/lib/build-strategies/strategies/DockerVMStrategy.ts
```diff
@@ -0,0 +1,254 @@
...
+    defaultIncludes: ["Cargo.toml", "Cargo.lock"],
+    baseImage: "namesty/base-rust",
+  },
+  "wasm/assemblyscript": {
+    defaultIncludes: ["package.json", "package-lock.json", "yarn.lock"],
+    baseImage: "namesty/base-assemblyscript",
+  },
+};
+
+interface BuildManifestConfig {
```
pileks: This seems like a duplicate of [PolywrapProject.ts:42](https://github.com/polywrap/toolchain/pull/1204/files#diff-77f66692da7e2e7f64d7d60655863cba49e4b2fa2071510ae8d1a0e655266510R42)

Any specific reason for the duplication?

packages/cli/src/lib/build-strategies/strategies/ImageStrategy.ts
```diff
@@ -0,0 +1,330 @@
...
+      let cacheDir: string | undefined;
+      let removeBuilder = false;
+
+      if (dockerBuildxConfig && typeof dockerBuildxConfig !== "boolean") {
+        const cache = dockerBuildxConfig.cache;
+
+        if (cache == true) {
+          cacheDir = this.project.getCachePath("build/image/cache");
+        } else if (cache) {
+          if (!path.isAbsolute(cache)) {
```
pileks: Just from a style perspective, maybe omit the invert the if so that we don't have to read the `!` which might be hard to notice at a cursory glance? Nitpicking here...

pileks (changes requested):
All in all looks great! See comments and we're good!

packages/cli/jest.rs.config.js
```diff
@@ -0,0 +1,24 @@
...
+  collectCoverage: true,
+  preset: "ts-jest",
+  testEnvironment: "node",
+  testMatch: ["**/__tests__/**/?(*.)+(spec|test).[jt]s?(x)"],
+  globals: {
+    "ts-jest": {
+      diagnostics: false
+    },
+  },
+  modulePathIgnorePatterns: [
```
cbrzn: We have two definitions of `modulePathIgnores` and `testMatch`- I guess this should change right?

packages/cli/src/__tests__/e2e/build-rs.spec.ts
```diff
@@ -0,0 +1,68 @@
+import { polywrapCli } from "./utils";
+
+import { runCLI } from "@polywrap/test-env-js";
+import { GetPathToCliTestFiles } from "@polywrap/test-cases";
+import fs from "fs";
+import path from "path";
+
+jest.setTimeout(500000);
```
cbrzn: It looks like this is not enough for CI when it comes to building w/docker :(

packages/cli/src/__tests__/e2e/build-rs.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  
+      const buildDir = `./build`;
+  
+      expect(code).toEqual(0);
+      expect(output).toContain(`Artifacts written to ${buildDir}`);
+      expect(output).toContain(`WRAP manifest written in ${buildDir}/wrap.info`);
+    });
+  })
+
+  // NOTE: Skipped because CI needs system prequisites: bindgen, snip, etc.
```
cbrzn, line 36: ignorant q: Would it be possible to install this in the CI script?

packages/cli/src/__tests__/e2e/build.spec.ts
```diff
@@ -1,4 +1,3 @@
-import { PolywrapProject, loadBuildManifest } from "../../lib";
```
cbrzn, line 1: nit: I am noticing that now we have `build` and `build-rs`. Would it make sense to create a folder `build` inside of `__tests__` and inside of this add an implementation test?

My main thought here is that for example if we would like to add tests for `go` build, it seems better to just add `go.spec.ts` inside of the `__tests__/build` folder, rather than creating a `build-go.spec.ts`?

packages/cli/src/commands/build.ts
```diff
@@ -8,19 +8,24 @@ import {
...
-  parseManifestFileOption,
   parseDirOption,
   parseClientConfigOption,
+  parseManifestFileOption,
 } from "../lib";
+import { CodeGenerator } from "../lib/codegen/CodeGenerator";
+import { LocalBuildStrategy } from "../lib/build-strategies/strategies/LocalStrategy";
+import { ImageBuildStrategy } from "../lib/build-strategies/strategies/ImageStrategy";
+import { BuildStrategy, SUPPORTED_STRATEGIES } from "../lib/build-strategies";
+import { DockerVMBuildStrategy } from "../lib/build-strategies/strategies/DockerVMStrategy";
```
cbrzn: nit: add these guys to `lib/build-strategies/index.ts` and just have one `import` statement here? 

feel free to ignore this it's a style thing

packages/cli/src/commands/build.ts
```diff
@@ -8,19 +8,24 @@ import {
...
 
-import { print } from "gluegun";
 import path from "path";
 import readline from "readline";
 import { PolywrapClient, PolywrapClientConfig } from "@polywrap/client-js";
+import { PolywrapManifest } from "@polywrap/polywrap-manifest-types-js";
 
 const defaultOutputDir = "./build";
+const defaultStrategy = "vm";
```
cbrzn: nit: what about converting this (& [supported strategies values](https://github.com/polywrap/toolchain/blob/namesty/docker-build-vm/packages/cli/src/lib/build-strategies/index.ts#L3)) to enum, so if you change the enum, you don't need to change it in any other file? :thinking:

packages/cli/src/commands/build.ts
```diff
@@ -67,56 +79,91 @@ export const build: Command = {
           ),
           clientConfig: await parseClientConfigOption(options.clientConfig),
           outputDir: parseDirOption(options.outputDir, defaultOutputDir),
+          strategy: options.strategy ?? defaultStrategy,
```
cbrzn: nit: I think this can be removed since we have the `default` attribute in strategy option?

packages/cli/src/lib/build-strategies/BuildStrategy.ts
```diff
@@ -0,0 +1,56 @@
...
+      "..",
+      "defaults",
+      "build-strategies",
+      language,
+      this.getStrategyName()
+    );
+
+    if (
+      fse.existsSync(
+        this.project.getCachePath(PolywrapProject.cacheLayout.buildStrategyUsed)
```
cbrzn: real nit feel free to ignore:

what about converting `this.project.getCachePath....` into a variable and use this variables in all the cases, hence, making this code easier to read?

packages/cli/src/lib/codegen/CodeGenerator.ts
```diff
@@ -0,0 +1,107 @@
...
+
+      return true;
+    } catch (e) {
+      gluegun.print.error(e);
+      return false;
+    }
+  }
+
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  protected async runCodegen(_: BindLanguage, __?: Ora): Promise<string[]> {
```
cbrzn, line 74: Probably I am missing something and am wrong here but do we need these arguments in this function?

packages/cli/src/lib/codegen/ScriptCodeGenerator.ts
```diff
@@ -0,0 +1,120 @@
...
+import Mustache from "mustache";
+import { Ora } from "ora";
+import path from "path";
+
+export class ScriptCodegenerator extends CodeGenerator {
+  private _script: string;
+  private _mustacheView: Record<string, unknown> | undefined;
+  private _codegenDirAbs: string;
+  private _omitHeader: boolean;
+  private _schema: string | undefined = "";
```
cbrzn: nit: all these attributes can be `readonly`

packages/cli/src/lib/project/PolywrapProject.ts
```diff
@@ -41,12 +39,28 @@ export interface PolywrapProjectConfig extends ProjectConfig {
   metaManifestPath?: string;
 }
 
+export interface BuildManifestConfig {
```
cbrzn, line 42: Same q: https://github.com/polywrap/toolchain/pull/1204#discussion_r986323786

packages/cli/src/lib/project/manifests/polywrap/languages.ts
```diff
@@ -37,3 +38,21 @@ export function polywrapManifestLanguageToBindLanguage(
       );
   }
 }
+
+export async function getGenerationSubpath(
```
cbrzn: Looks like this function's not being used

packages/js/plugins/ethereum/src/__tests__/integration/polywrap.build.yaml
```diff
@@ -1,6 +1,9 @@
-format: 0.1.0
-config:
-  node_version: "16.13.0"
+format: 0.2.0
+strategies:
+  image:
+    node_version: "14.16.0"
```
cbrzn, line 4: nit: would it be possible to use `16.13.0` in all the custom build manifests (for as) in monorepo?

packages/js/manifests/polywrap/src/formats/polywrap.build/0.2.0.ts
```diff
@@ -0,0 +1,101 @@
...
+ * This file was automatically generated by json-schema-to-typescript.
+ * DO NOT MODIFY IT BY HAND. Instead, modify the source JSONSchema file,
+ * and run json-schema-to-typescript to regenerate this file.
+ */
+
+export interface BuildManifest {
+  /**
+   * Polywrap build manifest format version.
+   */
+  format: "0.2.0" | "0.2";
```
cbrzn, line 13: i think we should not support `0.2`

packages/test-cases/cases/cli/wasm/build-cmd/assemblyscript/001-sanity/yarn.lock
```diff
@@ -0,0 +1,51 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
cbrzn: Remove

cbrzn (commented):
:fire: regarding next steps i can def help with the benchmark tool :)

packages/cli/src/__tests__/e2e/build-rs.spec.ts
cbrzn: now that I am thinking more about this, I think it would make sense to add instructions somehow to download the necessary dependencies if you'd like to run this build locally; if I am not wrong it would mean to run the `local.sh` from rust?

I guess you get the point - lmk if this doesn't make sense and i can try harder to explain myself

cbrzn:
btw just noticed you're fixing 19 alerts (https://github.com/polywrap/toolchain/pull/1204#issuecomment-1267123319) :goat:

packages/js/plugins/ethereum/src/__tests__/integration/polywrap.build.yaml
namesty: I think this should probably be done in a separate PR

packages/cli/src/lib/build-strategies/strategies/DockerVMStrategy.ts
namesty: Nope. Great catch on a merge remnant hehe

packages/cli/src/lib/build-strategies/strategies/ImageStrategy.ts
namesty: Yep, I agree with this

packages/cli/src/lib/codegen/CodeGenerator.ts
namesty: The base class may not use them, but they're still available for classes that inherit from this one. For example: `ScriptCodegen` uses them

packages/cli/src/__tests__/e2e/build.spec.ts
namesty: I agree, but I think we should do this consistently for the other test suites as well, probably in another PR

packages/cli/src/__tests__/e2e/build-rs.spec.ts
namesty: I just double checked, jsut to make sure there was no sneaky deadlock; and indeed, it just takes an awful lot of time. Pumped the timeout by 2 mins

packages/cli/src/__tests__/e2e/build-rs.spec.ts
namesty: Agreed. Yeah I was thinking about including the dependencies in a Readme or an install.sh script; because it shouldn't be the Polywrap's CLI responsibility to install these permanent system dependencies on the user's machine.

namesty: I almost feel like, we could get this merged and open an issue on where to include that in the docs, or maybe add a command flag to output an install script for the user to run. Thoughts? @dOrgJelli

pileks (approved):

packages/cli/src/__tests__/e2e/build-rs.spec.ts
cbrzn: > we could get this merged and open an issue on where to include that in the docs

yah lets do it

cbrzn (approved):

PR #1205, opened 2022-09-02
fix(cli): job Output and validation output having arbitrary order

Niraj-Kamdar:


packages/cli/src/__tests__/e2e/jobrunner-test-cases.ts
```diff
@@ -16,6 +16,7 @@ export const testCases: WorkflowTestCase[] = [
     workflow: {
       name: "simpleCalculator",
       format: "0.1",
+      __type: "PolywrapWorkflow",
```
cbrzn, line 19: Now that I am seeing this, should we call this manifest `PolywrapTest`? :thinking: I remember you had some thoughts regarding separation of concerns when it comes to `test` and `workflow` but not sure

Just bringing it up in case it is important, feel free to ignore this comment :smile:

dOrgJelli (dismissed):
LGTM!

cbrzn (commented):
Would it be possible to fix LGTM alert before merging this?

packages/cli/src/__tests__/e2e/jobrunner-test-cases.ts
Niraj-Kamdar: Agree but will be done in separate PR

PR #1217, opened 2022-09-07
Feat: TypeScript App/Plugin Codegen Type Safety

### LINKED ISSUE #897: TypeScript App/Plugin Codegen Types Aren't Very Type-Safe
Currently, most-all interface types for TypeScript applications are like so:
```typescript
interface ModuleMethodInput extends Record<string, unknown> {
  argument?: Types.String | null;
}

export const Module = {
  method: async (
    input: ModuleMethodInput,
    client: Client
  ): Promise<InvokeApiResult<...>> => {
    ...
  }
}
```

The problem is that our input types extend from `Record<string, unknown>`, which is a generic property-map. This means that code like this will still compile, when it shouldn't:
```typescript
Module.method({ foo: "value" });
// ^ "foo" should be "argument"
```

krisbitney:
This PR enforces type safety for generated TypeScript code. Instead of extending `Record<string, unknown>`, generated object and arg types are cast to `Record<string, unknown>` just before invocations are called in the generated `Module` type. Generated arg types must be typecast to `Record<string, unknown>` because the client's `invoke` method expects a `Record<string, unknown> | UInt8Array | undefined`.

The TypeScript interface for the generated `Env` type still extends `Record<string, unknown>` to ensure it conforms to the generic parameter restriction set by the `PluginModule` definition in `@polywrap/core-js`.

Closes https://github.com/polywrap/toolchain/issues/897

### DIFF
--- a/packages/schema/bind/src/bindings/typescript/app/templates/types-ts.mustache
+++ b/packages/schema/bind/src/bindings/typescript/app/templates/types-ts.mustache
@@ -88,7 +88,7 @@ export type {{#detectKeyword}}{{type}}{{/detectKeyword}} = {{type}}Enum | {{type
 {{#methods}}
 
 /* URI: "{{parent.uri}}" */
-interface {{parent.type}}_Args_{{name}} extends Record<string, unknown> {
+interface {{parent.type}}_Args_{{name}} {
   {{#arguments}}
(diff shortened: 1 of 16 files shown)


pileks (approved):
LGTM!

dOrgJelli (approved):
Awesome!

PR #1218, opened 2022-09-07
fix: update tracer + install latest signoz

dOrgJelli:


packages/cli/src/lib/defaults/infra-modules/tracer/docker-compose.yaml
```diff
@@ -44,27 +68,80 @@ services:
...
+    #   - "1777:1777"     # pprof extension
+    #   - "8888:8888"     # OtelCollector internal metrics
+    #   - "13133:13133"   # Health check extension
+    #   - "55679:55679"   # zPages extension
     restart: on-failure
     depends_on:
       clickhouse:
         condition: service_healthy
+
+  hotrod:
```
fetsorn: We might want to remove `hotrod` and `load-hotrod` services to avoid mock traces in the database.

packages/cli/src/lib/defaults/infra-modules/tracer/docker-compose.yaml
```diff
@@ -44,27 +68,80 @@ services:
...
     container_name: frontend
     restart: on-failure
     depends_on:
+      - alertmanager
       - query-service
     ports:
       - "3301:3301"
     volumes:
-      - ./nginx-config.conf:/etc/nginx/conf.d/default.conf
+      - ../common/nginx-config.conf:/etc/nginx/conf.d/default.conf
```
fetsorn: The `../common/nginx-config.conf` file should be copied as well, otherwise it appears as an empty directory in `.polywrap/infra/modules` when docker fails to access it, and the container fails.
And then this path should be changed back to "./nginx-config.conf:/etc/nginx/conf.d/default.conf"

packages/cli/src/lib/defaults/infra-modules/tracer/docker-compose.yaml
dOrgJelli: Okay sounds good, will do

dOrgJelli:
NOTE: Do not merge this yet, I haven't been able to run the tracer successfully, for some reason docker is hanging.

Issue #1134, opened 2022-08-11
Bug: Support interface invocations in Rust


krisbitney:
In AssemblyScript wrappers, we can get an interface's registered implementation URIs and use one of the URIs to instantiate an interface module. This makes it possible for us to invoke an interface agnostic to its implementation.

The logic in the AssemblyScript and Rust bindings is different. In Rust, it is not possible to invoke an interface because we are generating interface modules as though they were standard modules. The Rust method attempts to invoke the interface URI rather than the URI of an implementation.

For example, contrast the following generated code for Rust and AssemblyScript using the same schema:

Schema
```graphql
#import { Module } into Interface from "wrap://ens/goerli/interface.sleep.wrappers.eth"
#use { getImplementations } for Interface
```

Rust Codegen
```rust
pub struct SleepInterfaceModule {}

impl SleepInterfaceModule {
    pub const URI: &'static str = "wrap://ens/goerli/interface.sleep.wrappers.eth";

    pub fn new() -> SleepInterfaceModule {
        SleepInterfaceModule {}
    }

    pub fn sleep(args: &ArgsSleep) -> Result<Option<bool>, String> {
        let uri = SleepInterfaceModule::URI;
        let args = serialize_sleep_args(args).map_err(|e| e.to_string())?;
        let result = subinvoke::wrap_subinvoke(
            uri,
            "sleep",
            args,
        )?;
        deserialize_sleep_result(result.as_slice()).map_err(|e| e.to_string())
    }
}

```

AssemblyScript Codegen
```typescript
export class SleepInterface_Module {

  public static interfaceUri: string = "wrap://ens/goerli/interface.sleep.wrappers.eth";

  public uri: string;

  constructor(uri: string) {
    this.uri = uri;
  }

  public sleep(
    args: Args_sleep
  ): Result<Option<bool>, string> {
    const argsBuf = serializesleepArgs(args);
    const result = wrap_subinvokeImplementation(
      "wrap://ens/goerli/interface.sleep.wrappers.eth",
      this.uri,
      "sleep",
      argsBuf
    );

    if (result.isErr) {
      return Result.Err<Option<bool>, string>(
        result.unwrapErr()
      );
    }

    return Result.Ok<Option<bool>, string>(
      deserializesleepResult(result.unwrap())
    );
  }
}
```

nerfZael:
How do plugins registered for interfaces work then?
I almost wanna say that the AS implementation is wrong and the rust one is correct, at least ATM
[...]

PR #1219, opened 2022-09-08
fix: interface module bindings

Niraj-Kamdar:
Rust binding for interface-implementation invocation isn't working due to a bug in rust binding template checkout below issue for more details
- Closes: #1134


cbrzn (commented):
Thanks for pushing this 🙏 Probably I am wrong and please correct me if so, but I think this adds the capability of interface invocation, hence, we should add the `interface-invoke` test cases (as in https://github.com/polywrap/toolchain/tree/origin/packages/test-cases/cases/wrappers/wasm-as/interface-invoke)

Niraj-Kamdar:
@cbrzn I agree, the tests should cover this scenario

packages/js/client/src/__tests__/e2e/wasm-as.spec.ts
```diff
@@ -137,12 +137,18 @@ describe("wasm-as test cases", () => {
   });
 
   it("implementations - getImplementations", async () => {
+    const interfacePath = `${GetPathToTestWrappers()}/wasm-rs/implementations/test-interface`;
     const interfaceUri = "wrap://ens/interface.eth";
 
-    const implementationPath = `${GetPathToTestWrappers()}/wasm-as/implementations/test-use-getImpl`;
-    const implementationUri = `wrap://fs/${implementationPath}/build`;
+    const implementationPath = `${GetPathToTestWrappers()}/wasm-rs/implementations/test-wrapper`;
```
krisbitney: It looks like there is a typo here. The wrapper URIs point to the `wasm-rs` folder, but they should point to `wasm-as`.

krisbitney (changes requested):

krisbitney (approved):

cbrzn (approved):

PR #1220, opened 2022-09-08
fix(schema-bind): handle reserved keywords for serde in rust

Niraj-Kamdar:
Currently, Rust JSON serializer doesn't support de/serializing the reserved words. This PR fixes it.

- Close: https://github.com/polywrap/toolchain/issues/1221


cbrzn (commented):
Would it make sense to add a test in the `test-cases/cases/wasm/wasm-rs/json-type` folder to ensure this works?

packages/js/client/src/__tests__/e2e/test-cases.ts
```diff
@@ -473,6 +473,28 @@ export const runJsonTypeTest = async (client: PolywrapClient, uri: string) => {
 
   const methodJSONResult = JSON.stringify(json);
   expect(methodJSONResponse.data).toEqual(methodJSONResult);
+
+  if (testReserved) {
```
cbrzn, line 477: If this is added, then the `wasm-rs` file should be edited to pass `true` for this arg when calling the `runJsonTypeTest` function

packages/js/client/src/__tests__/e2e/test-cases.ts
```diff
@@ -412,7 +412,7 @@ export const runInvalidTypesTest = async (
   );
 };
 
-export const runJsonTypeTest = async (client: PolywrapClient, uri: string) => {
+export const runJsonTypeTest = async (client: PolywrapClient, uri: string, testReserved: boolean = false) => {
```
dOrgJelli, line 415: Can you please pass in `true` for `testReserved` in the wasm-rs tests? This logic is not currently being used.

dOrgJelli (changes requested):
LGTM!

cbrzn (approved):

dOrgJelli (approved):

PR #1225, opened 2022-09-08
Fix: build issues following Origin 0.7.0 release

krisbitney:


nerfZael (dismissed):

pileks (dismissed):
Nice catch with the `codegen` stuff! That was my bad for not thinking ahead...

dOrgJelli (approved):

dOrgJelli:
tests aren't passing, investigating.

packages/js/plugins/uri-resolvers/file-system-resolver/src/index.ts
```diff
@@ -37,7 +36,7 @@ export class FileSystemResolverPlugin extends Module<NoConfig> {
     if (manifestExistsResult.data) {
       try {
         const manifestResult = await FileSystem_Module.readFile(
-          { path: manifestPath, encoding: FileSystem_EncodingEnum.BINARY },
```
dOrgJelli, line 40: Why were these changes made? I'm going to revert them as they're causing tests to fail.

dOrgJelli (changes requested):

packages/js/plugins/uri-resolvers/file-system-resolver/src/index.ts
dOrgJelli: My mistake! I just realized that the `encoding` property was removed from the `readFile` method's arguments, and wasn't causing an error because the arguments used to be a generic object. With the new release this is causing an error because the property does not exist. My bad @krisbitney

cbrzn (approved):

PR #1231, opened 2022-09-13
Feat: Add '--no-codegen' option to CLI build command

### LINKED ISSUE #796: allow user to change generated code by giving option to not run codegen while building

krisbitney:
This PR adds a `--no-codegen` option to the CLI's build command. When the flag is present, bindings are not generated during the build process. Users are expected to generate bindings with the `codegen` command. Bindings should be located at `src/wrap`, where the compiler will look for them.

```bash
polywrap build --no-codegen
polywrap build -n
```

**Questions**
- Should this feature be added or moved to the build manifest?
- Should users be able to specify a path to their bindings directory other than `src/wrap`?

Closes https://github.com/polywrap/toolchain/issues/796

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -28,6 +28,7 @@ type BuildCommandOptions = {
   manifestFile: string;
   outputDir: string;
   clientConfig: Partial<PolywrapClientConfig>;
+  codegen: boolean; // defaults to true
   watch?: boolean;
   verbose?: boolean;
 };
@@ -54,6 +55,7 @@ export const build: Command = {
         `-c, --client-config <${intlMsg.commands_common_options_configPath()}>`,
         `${intlMsg.commands_common_options_config()}`
       )
+      .option(`-n, --no-codegen`, `${intlMsg.commands_build_options_n()}`)
       .option(`-w, --watch`, `${intlMsg.commands_build_options_w()}`)
       .option(`-v, --verbose`, `${intlMsg.commands_build_options_v()}`)
       .action(async (options) => {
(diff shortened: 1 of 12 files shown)


pileks:
> Should this feature be added or moved to the build manifest?

I'd say that I could see a world where this is useful.

> Should users be able to specify a path to their bindings directory other than src/wrap?

This should definitely be a thing, as `codegen` has a `codegenDir` option, so it only makes sense to also have a `codegenDir` or `bindingsDir` option.

packages/cli/src/commands/build.ts
```diff
@@ -54,6 +55,7 @@ export const build: Command = {
         `-c, --client-config <${intlMsg.commands_common_options_configPath()}>`,
         `${intlMsg.commands_common_options_config()}`
       )
+      .option(`-n, --no-codegen`, `${intlMsg.commands_build_options_n()}`)
```
pileks, line 58: While `-n` is okay, maybe we could have `-nc` just so that we don't take up the `-n` flag if we see a future where we could use it for something else?
Not a blocking issue, dismiss at will.

pileks (changes requested):
If you are to additionally specify a `--codegen-dir` option, I feel like we should also add a test for it.

Otherwise, LGTM!

packages/cli/src/commands/build.ts
krisbitney: The short flags can only be one character in the `commander` framework we are using

pileks: We can consider it resolved then.

pileks (approved):

packages/cli/src/commands/build.ts
```diff
@@ -68,7 +70,14 @@ export const build: Command = {
...
 
 async function run(options: BuildCommandOptions) {
-  const { watch, verbose, manifestFile, outputDir, clientConfig } = options;
+  const {
+    watch,
+    verbose,
+    manifestFile,
+    outputDir,
+    clientConfig,
+    codegen,
```
Niraj-Kamdar, line 79: does commander do this negation implicitly?

Niraj-Kamdar (approved):

dOrgJelli (approved):

packages/cli/src/commands/build.ts
krisbitney: Yeah https://github.com/tj/commander.js#other-option-types-negatable-boolean-and-booleanvalue

PR #1236, opened 2022-09-15
Plugin refactor

### LINKED ISSUE #1259: Remove plugin types from client and core
Plugin types and all mention of plugins should be removed from client-js and core-js packages.
After they URI resolution refactor and the addition of PackageResolver they can now be treated as WrapPackages which can be registered in a resolver array.

Instead of this: 
```typescript
 plugins: [
        {  
            uri: "ens/ethereum.polywrap.eth",
            plugin: ethereumPlugin({ connections })
        },
      ]
});
```
We should be able to do this:
```typescript
 resolver: UriResolver.from([
        {  
            uri: "ens/ethereum.polywrap.eth",
            package: ethereumPlugin({ connections })
        },
      ]
  )
});
```

NOTE: `UriResolver.from` is not implemented, currently it is implemented in the form of the `buildUriResolver` func, but `UriResolver.from` is more readable.

Overriding plugins can be moved to the config builder in the form of "packages"

nerfZael:
# Description:

The goal of this refactor is to remove plugins from the core-js, and client-js packages and improve their composability with resolvers
This PR introduces a new plugin-js package where the PluginWrapper and PluginPackage classes are located.

## Features: 
- client-js
  - Polywrap Client now re-exports the config builder and uri-resolvers (in addition to core) packages. This is done to improve dev exp and remove the need for users to import those package themselves.
    - For users who do not need those package and are using `noDefaults` there will be a separate PR that refactor core client functionality into a core-client package that does not depend on the config builder and uri-resolvers packages, but has no defaults.
- config-builder-js
  - Added `addRedirects`, `addWrappers`, `addPackages` methods to the `ClientConfigBuilder`, so users can add many items at once.
  - Added `buildDefault` to the `ClientConfigBuilder` which builds a `ClientConfig` using default resolvers.
- plugin-js 
  - New package for plugins
  - Can create plugin packages with `PluginPackage.from`
    - Accepts `manifest` and a `PluginModule` or an inline `PluginModule` 
- uri-resolvers-js
  - Added `StaticResolver` and `StaticResolver.from` to optimize building resolvers with `IUriRedirect`, `IUriWrapper` and `IUriPackage`.
- schema-bind-js
  - In the module template, `PluginModule` is now imported fron plugin-js instead of core-js

## Breaking changes: 
- client-js
  - The Polywrap Client with `noDefaults: false` no longer accepts a `plugins` field, but it accepts `wrappers` and `packages`.
    - `resolver` field has been replaced with `resolvers`, since with default client the resolver used is the `RecursiveResolver` with the `PackageToWrapperCacheResolver`.
  - The Polywrap Client with `noDefaults: true`, no longer accepts a `plugins` field. It is expected that devs using this option will manually configure their own resolver.
  - removed `getPlugins` and `getPluginByUri`. Will add `getWrapper`, `getWrapperByUri`, `getPackage`, `getPackageByUri`, in a follow up PR.
  - `createPolywrapClient` function has been deleted from the client-js package as it is unnecessary
- config-builder-js
  - Now uses the `CustomClientConfig` which doesn't have `plugins` and a `resolver`, but now has `wrappers`, `packages` and `resolvers`
  - Calling build returns an instance of the `CustomClientConfig`, which can be used with defaults from the `PolywrapClient`, but can not be used if `noDefaults: true` is passed to the `PolywrapClient` constructor.
  - Removed `addPlugin` from the `ClientConfigBuilder`, users can now use `addWrapper` or `addPackage` where appropriate.
  - Renamed `addUriRedirect` to `addRedirect` to keep it inline with `addWrapper` and `addPackage` (IUriRedirect, IUriWrapper, IUriPackage)
- core-js
  - Plugins are no longer a part of this package, they have been moved to the plugin-js package
  - Renamed `UriRedirect` to `IUriRedirect` to match `IUriWrapper` and `IUriPackage`
  - `IUriRedirect`, `IUriWrapper` and `IUriPackage` are now generic and their generic param implements `Uri | string`
  - Removed `options` argument from `client.getManifest` method since all wrappers have a deserialized manifest
- react-js
  - Replaced `plugins` on the `PolywrapProvider` with `wrappers` and `packages`
- uri-resolvers-js
  - Replaced helper func `buildUriResolver` with `UriResolver.from`
  - Constructors of built-in resolvers like `RecursiveResolver` and `PackageToWrapperCacheResolver` now accept a concrete `IUriResolver` while their static `from` methods accept a `UriResolverLike` 
  - Remove `PluginsResolver` and `PluginResolver`, users can now use `WrapperResolver` or `PackageResolver`


packages/cli/src/lib/helpers/client.ts
```diff
@@ -1,29 +1,29 @@
...
+import { PolywrapClient } from "@polywrap/client-js";
 
 interface SimpleClientConfig {
   ensAddress?: string;
   ethProvider?: string;
   ipfsProvider?: string;
 }
 
-export function getSimpleClient(config: SimpleClientConfig): PolywrapClient {
+export function getSimpleClient(config: SimpleClientConfig): Client {
```
cbrzn: it looks like this function its not being used should we remove this file?

packages/cli/src/lib/test-env/client-config.ts
```diff
@@ -10,12 +13,10 @@ import {
...
 import { ipfsPlugin } from "@polywrap/ipfs-plugin-js";
 import { ensAddresses } from "@polywrap/test-env-js";
-import { Env } from "@polywrap/core-js";
+import { Env, IUriPackage, Uri } from "@polywrap/core-js";
 
-export async function getTestEnvClientConfig(): Promise<
-  Partial<PolywrapClientConfig>
-> {
-  const providers = await getTestEnvProviders();
+export function getTestEnvClientConfig(): Partial<PolywrapClientConfig> {
```
cbrzn, line 14: This function is also not being used, let's remove it

packages/cli/package.json
```diff
@@ -56,6 +56,7 @@
     "@polywrap/test-env-js": "0.9.0",
     "@polywrap/wasm-js": "0.9.0",
     "@polywrap/wrap-manifest-types-js": "0.9.0",
+    "@polywrap/uri-resolvers-js": "0.9.0",
```
cbrzn: i think this package is not being used here

cbrzn (commented):
nit feel free to ignore: i am noticing a pattern in plugin tests where we have a `getClient` function - would it makes sense to use the config builder so we start to use this pattern in the toolchain?

packages/cli/src/lib/test-env/client-config.ts
nerfZael: Nice catch, this is old code, I think I can merge these two functions

nerfZael:
> nit feel free to ignore: i am noticing a pattern in plugin tests where we have a `getClient` function - would it makes sense to use the config builder so we start to use this pattern in the toolchain?

@cbrzn 
Only in the places where we want to test the config builder and or the default client config.
Ideally, tests test just a single functionality, so they should depend on the least amount of deps and use a client with only the necessary packages configured.
That's why in most of the tests I'm using `noDefaults: true` on the client ctor.

packages/cli/src/lib/helpers/client.ts
nerfZael: definitely

packages/js/plugin/src/PluginPackage.ts
```diff
@@ -0,0 +1,60 @@
...
+      );
+    }
+  }
+
+  async getManifest(): Promise<Result<WrapManifest, Error>> {
+    return ResultOk(this.manifest);
+  }
+
+  async createWrapper(): Promise<Result<Wrapper, Error>> {
+    return ResultOk(new PluginWrapper(this.manifest, this.pluginModule));
```
dOrgJelli, line 58: Should we support creating `PluginPackage`s with a "module factory function"? This way, if a plugin developer wants their module to be re-created every time an invocation occurs, we can support this?

packages/js/client/src/PolywrapClient.ts
```diff
@@ -1,96 +1,73 @@
...
-          consoleEnabled: !!config?.tracerConfig?.consoleEnabled,
-          consoleDetailed: config?.tracerConfig?.consoleDetailed,
-          httpEnabled: !!config?.tracerConfig?.httpEnabled,
-          httpUrl: config?.tracerConfig?.httpUrl,
-          tracingLevel: config?.tracerConfig?.tracingLevel,
-        },
-      };
-
-      this._validateConfig();
+      this._config = !options?.noDefaults
```
dOrgJelli, line 67: This does not properly handle the `config === undefined` scenario, what should be done when that is the case?

dOrgJelli (changes requested):
I'm good with everything, except for the "PolywrapClient undefined config scenario" I comment about, and the renaming of the config classes we discussed via DM.

packages/js/core/src/types/Client.ts
```diff
@@ -16,8 +14,7 @@ import { WrapManifest } from "@polywrap/wrap-manifest-types-js";
 import { Result } from "@polywrap/result";
 
 export interface ClientConfig<TUri extends Uri | string = string> {
-  readonly redirects: Readonly<UriRedirect<TUri>[]>;
-  readonly plugins: Readonly<PluginRegistration<TUri>[]>;
+  readonly redirects: Readonly<IUriRedirect<TUri>[]>;
```
Niraj-Kamdar: Do we still need to keep redirects in config if it can just be a resolver?

nerfZael: We don't, but this PR is already quite large, so I'm focusing it on plugins and going to refactor redirects as a separate one

packages/js/plugin/src/PluginPackage.ts
nerfZael: That's a good idea I think, I'll probably do that as a follow-up PR since I wanted to revisit PluginWrappers and use them in default client config instead if PluginPackages anyway.

Niraj-Kamdar (approved):
This is awesome!

Niraj-Kamdar (dismissed):

packages/js/client/src/PolywrapClient.ts
nerfZael: It actually handles it, but typings are a bit out of date, I've updated them now

PR #1239, opened 2022-09-15
Client (Python): Msgpack implementation

cbrzn:


packages/py/msgpack/msgpack_wrap/__init__.py
```diff
@@ -0,0 +1,27 @@
+import msgpack
+from dataclasses import dataclass, is_dataclass
+
+
+def sanitize(v):
+    for key in v.keys():
+        if is_dataclass(v[key]):
+            v[key] = v[key].__dict__
```
Niraj-Kamdar: do not access `__dict__` directly.

use `asdict` function from the dataclassses package.

packages/py/msgpack/tests/test_msgpack.py
```diff
@@ -0,0 +1,62 @@
...
+    custom_object = {
+        "firstKey": "firstValue",
+        "secondKey": "secondValue"
+    }
+
+    encoded = msgpack_encode(custom_object)
+    assert encoded == bytes(expected_array_like)
+
+    decoded = msgpack_decode(encoded)
+    assert decoded == custom_object
```
Niraj-Kamdar: encode dictionary as our generic map and encode dataclasses as the object (msgpack map family).

Niraj-Kamdar (changes requested):

packages/py/msgpack/msgpack_wrap/__init__.py
Niraj-Kamdar: You don't need to recurse when using asdict, it will take care of it for you

cbrzn: I think we still want to have recursion here. The reason is that if the value is a dictionary and has nested instances of classes, we need to check those. Does this make sense to you?

Niraj-Kamdar: Gotcha! It makes sense in that case but try to run this code in python3.10 and you will understand why it is bad idea to access `__dict__` :)

```python
from dataclasses import dataclass

@dataclass
class Test1:
    a: str
    
@dataclass(slots=True)
class Test2:
    a: str
    
print(Test1(a="1"))
print(Test1(a="1").__dict__)

print(Test2(a="2"))
print(Test2(a="2").__dict__)

```

cbrzn: Yah I removed the `__dict__` and we are now using `asdict` :smile:

PR #1260, opened 2022-09-20
Enable fetching arrays from `view` functions with Ethereum plugin

### LINKED ISSUE #1241: Automatically handle arrays and structs in solidity contract ABIs for Ethereum plugin calls
Due to the way Solidity encodes an ABI, a solidity function that returns a Struct or a dynamically sized array (or map) is difficult to call. 

See: https://docs.soliditylang.org/en/v0.8.17/abi-spec.html

A struct or array is actually encoded a a tuple. 

```
T[k] for any T and k:

enc(X) = enc((X[0], ..., X[k-1]))

i.e. it is encoded as if it were a tuple with k elements of the same type.
```
```
T[] where X has k elements (k is assumed to be of type uint256):

enc(X) = enc(k) enc([X[0], ..., X[k-1]])

i.e. it is encoded as if it were an array of static size k, prefixed with the number of elements.
```

This is kind of annoying but the resolution is straightforward when the return value is a struct or when we know the size of the returned array.

For example, let's say we want to call this function:
```solidity
function getPendingJobs() external view returns (Job[] memory)
```

And let's say the struct `Job` looks like this:
```
struct Job { 
  uint foo;
  bool: bar;
}
```

Let's say you expect the method to return an array of two jobs. Then the abi you call will look like this:
```typescript
"function getPendingJobs() external view returns (tuple(tuple(uint foo, bool bar), tuple(uint foo, bool bar)))"
```

But what if we don't know the length of the returned array? I don't know! And I'm probably not the only one. Wouldn't it be nice if the Ethereum plugin could handle this automagically?

Note that the same issue exists for all dynamically sized types used as method arguments or return values.

I want to be able to call an Ethereum contract with the Ethereum plugin using an ABI string that looks like:
`"function getPendingJobs(jobs: Job[]) external view returns (Job[] memory)"`

**Additional context**
It seems Ethers has solved this problem and it is open source. If there is no way for us to use ethers differently to handle these situations, then we can look at the ethers repo to see how they do it.

sraver:
Closes #1241

The approach has been to rely on the native way that `ethers` uses. So in order to enable processing returned array values, `method` is allowed to be a string containing the ABI of the desired method. 
When that is the case, the returned value is processed differently to be able to create an object and stringify it.

There's three different cases that have been taken care of:

### Return a primitive value (uint, address, ...)
In this case the result is the same as to when using human-readable ABI

### Return an array of primitives
In this case, the result is a stringified array of the values.

### Return an array of structs
In this case, the result is a stringified array of objects that have as many keys as the struct have, each with its value.


packages/js/plugins/ethereum/src/utils/storage.ts
```diff
@@ -0,0 +1,53 @@
+import { Connection } from "@polywrap/ethereum-plugin-js";
```
dOrgJelli: Having this file here is odd, as it's only used for testing. I'd suggest moving it within the `__tests__` directory.

packages/js/plugins/ethereum/src/index.ts
```diff
@@ -67,10 +67,42 @@ export class EthereumPlugin extends Module<EthereumPluginConfig> {
...
+      isRawAbiMethod = true;
+    } catch (e) {
+      method = [args.method];
+      isRawAbiMethod = false;
+    }
+    const contract = connection.getContract(args.address, method, false);
     const funcs = Object.keys(contract.interface.functions);
     const res = await contract[funcs[0]](...parseArgs(args.args));
-    return res.toString();
+    if (!isRawAbiMethod) {
```
dOrgJelli: Instead of having this logic here, I would suggest moving it within a `parseResult` function that can be reused. If we have this parsing problem for view methods, we most likely have the same problem for transaction methods as well.

So, in addition to moving to reusable function, we should use it in all other places where we're naively doing `.toString` on the result of a contract method call.

packages/js/plugins/ethereum/src/index.ts
```diff
@@ -67,10 +67,42 @@ export class EthereumPlugin extends Module<EthereumPluginConfig> {
     _client: Client
   ): Promise<string> {
     const connection = await this._getConnection(args.connection);
-    const contract = connection.getContract(args.address, [args.method], false);
+    let method, isRawAbiMethod;
+    try {
+      method = JSON.parse(args.method);
```
dOrgJelli: This doesn't account for the case where the caller passes in a method's ABI in JSON, but doesn't wrap in an array.

packages/js/plugins/ethereum/src/__tests__/e2e.spec.ts
```diff
@@ -28,6 +29,9 @@ const contracts = {
   SimpleStorage: {
     abi: require("./contracts/SimpleStorage.ABI.json"),
     bytecode: `0x${require("./contracts/SimpleStorage.Bytecode.json").object}`,
+    abiSinglePrimitiveMethod: '[{"inputs":[],"name":"get","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"}]',
```
dOrgJelli, line 36: would like to see a test where we use the "hand written" form of the method's signature, in addition to the JSON signature.

For example `function foo(...) returns(address[])`

dOrgJelli (changes requested):

packages/js/plugins/ethereum/src/utils/storage.ts
sraver: my bad! I thought it was in ^^

packages/js/plugins/ethereum/src/index.ts
sraver: agreed. The other two methods where we could use these, afaik, would be on `callContractStatic`, `callContractMethod` and `sendTransaction`. The last ones are a proper tx, so we should be getting the results from `tx.data`.

Will add

packages/js/plugins/ethereum/src/index.ts
sraver: you mean if someone would send something like.. 

`{"inputs":[],"name":"get","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"}`

?

packages/js/plugins/ethereum/src/__tests__/e2e.spec.ts
sraver: indeed, that makes lot of sense :ok_hand:

packages/js/plugins/ethereum/src/index.ts
sraver: Probably also on `callContractMethodAndWait`, but `TxReceipt` doesnt have an attribute `data` to set the results :/

sraver: hm, i assumed that i could get the return values from a transaction, but no i cant. Had never done that before. Only static calls return the values when the function is not `view`, therefore... seems like `callContractStatic` and `callContractView` would be the only two methods to parse results i think

packages/js/plugins/ethereum/src/index.ts
```diff
@@ -67,34 +67,39 @@ export class EthereumPlugin extends Module<EthereumPluginConfig> {
...
+            ? ethers.BigNumber.from(args.txOverrides.gasLimit)
+            : undefined,
+          value: args.txOverrides?.value
+            ? ethers.BigNumber.from(args.txOverrides.value)
+            : undefined,
+        }
+      );
       return {
-        result: res.toString(),
+        result: result.length ? this._parseResult(abi, result) : "",
```
sraver: @dOrgJelli im not very proud of this line.. but i wanted to keep compatibility on how it is being formatted now. The alternative was parsing an empty array as an empty string, and that didnt feel like a right result to return from the `_parseResult` function. 

wyt?

sraver:
@dOrgJelli other places where a contract is instantiated are

https://github.com/polywrap/toolchain/blob/origin/packages/js/plugins/ethereum/src/index.ts#L204
https://github.com/polywrap/toolchain/blob/origin/packages/js/plugins/ethereum/src/index.ts#L258
https://github.com/polywrap/toolchain/blob/origin/packages/js/plugins/ethereum/src/index.ts#L393

should raw ABI be allowed on those calls too, even when there is no possible array return value?

packages/js/plugins/ethereum/src/index.ts
```diff
@@ -411,6 +416,58 @@ export class EthereumPlugin extends Module<EthereumPluginConfig> {
   ): Promise<Connection> {
     return this._connections.getConnection(connection || this.env.connection);
   }
+
+  private _constructAbi(method: string) {
```
dOrgJelli: minor NIT: all of these private methods that don't require the use of `this` should be put into the `utils` folder as "pure" functions. I think they'd fit nicely into the "./utils/parse.ts" module.

dOrgJelli (changes requested):
I think `_callContractMethod` needs the same type of return value parsing, no? Feel free to add the parsing everywhere where we're receiving return results from contract methods. Would hate to have this bug persist in some corners of the implementation.

packages/js/plugins/ethereum/src/index.ts
dOrgJelli: What is the type of `result`? Is the `.toString()` in the original code necessary? I feel like you should just be able to do `result: this._parseResult(result)` no?

sraver: Depending on the function we are calling, `result` will either be an array or an object(string, BigInteger, ..). 
Normally parsing it with the function should be the only thing to do with it, but right now we have a test [here](https://github.com/polywrap/toolchain/blob/origin/packages/js/plugins/ethereum/src/__tests__/e2e.spec.ts#L135) that is asserting that the return value of this function is empty string, and in reality the raw `result` of that function is an empty array(idk why cos the function `function register(bytes32 label, address owner)` doesn't specify any return.. but that comes back).
I fear that changing it could break something if there is someone loosely checking truthy/falsey on the result of staticCall.

But yeah, agree that simply parsing the result is what makes sense. Should i change the test?

packages/js/plugins/ethereum/src/index.ts
sraver: 100%, done :ok_hand:

sraver:
> I think `_callContractMethod` needs the same type of return value parsing, no? Feel free to add the parsing everywhere where we're receiving return results from contract methods. Would hate to have this bug persist in some corners of the implementation.

I thought so too at first tbh, but then, this method is used to execute transactions rather than simple calls, so the received data will never be the `return` from the function, but the TxResponse with the details, therefore we have nothing to parse in those cases. Return data from non-constant functions can only be retrieved from Solidity executors, off-chain executors only receive the receipt.

In any case, I'm adding the `constructAbi` to those methods too so at least the `method` input is consistent across all the functions. I don't think no one will actually uses raw format if there's no need, but seems just right i guess.

dOrgJelli:
> Return data from non-constant functions can only be retrieved from Solidity executors, off-chain executors only receive the receipt.

Ah whoops, you're correct, my bad! Thank you for the diligence @sraver 🙏

PR #1261, opened 2022-09-20
feat: getRandom JS Plugin

### LINKED ISSUE #901: getRandomValues Support
Currently wrappers cannot access the system's `getRandomValues` interface. We should build a plugin to support this. More research & design is needed.

Relevant links:
https://docs.rs/getrandom/latest/getrandom/index.html 
https://docs.rs/getrandom/latest/getrandom/macro.register_custom_getrandom.html
https://www.assemblyscript.org/stdlib/math.html#variants
https://www.assemblyscript.org/concepts.html#special-imports
https://www.w3.org/TR/WebCryptoAPI/#Crypto-method-getRandomValues
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues

dOrgJelli:
closes: https://github.com/polywrap/toolchain/issues/901

### DIFF
--- a/packages/js/client-config-builder/package.json
+++ b/packages/js/client-config-builder/package.json
@@ -31,6 +31,7 @@
     "@polywrap/ipfs-plugin-js": "0.7.0",
     "@polywrap/ipfs-resolver-plugin-js": "0.7.0",
     "@polywrap/logger-plugin-js": "0.7.0",
+    "@polywrap/random-plugin-js": "0.7.0",
     "@polywrap/tracing-js": "0.7.0",
     "@polywrap/wrap-manifest-types-js": "0.7.0"
   },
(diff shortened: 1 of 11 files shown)


dOrgJelli:
Closing this PR. I overlooked the fact that there exists a WASI "get random bytes" method: https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#-random_getbuf-pointeru8-buf_len-size---result-errno

Instead of creating this plugin, we can create a Rust wrapper, use [this create](https://docs.rs/getrandom/latest/getrandom/), compile for `wasi`, and boom we're all good!

However this is obviously dependent on the Polywrap client supporting WASI, which is being tracked here: https://github.com/polywrap/toolchain/issues/1262

PR #1272, opened 2022-09-22
Deploy manifest changes to be similar to Test manifest

namesty:


packages/cli/src/commands/deploy.ts
```diff
@@ -26,8 +26,8 @@ type DeployCommandOptions = {
   verbose?: boolean;
 };
 
-type ManifestSequence = DeployManifest["sequences"][number];
-type ManifestStep = ManifestSequence["steps"][number];
+type ManifestJob = DeployManifest["jobs"][number];
```
dOrgJelli, line 29: These types should come from the generated manifest types, no? Checkout how the test manifest uses reference typings in its json-schema.

dOrgJelli (changes requested):

packages/cli/src/lib/deploy/sequence.ts
```diff
@@ -1,9 +1,9 @@
-import { Step, StepName, StepResult, UriOrPrevStepResult } from "./step";
+import { DeployStep, StepName, StepResult, UriOrPrevStepResult } from "./step";
```
dOrgJelli, line 1: The name of this file should be `job.ts` instead of `sequence.ts` no?

packages/cli/src/lib/deploy/sequence.ts
```diff
@@ -13,22 +13,22 @@ export interface SequenceResult {
...
+interface DeployJobArgs {
   name: string;
-  steps: Step[];
+  steps: DeployStep[];
   config: Record<string, unknown>;
   printer: GluegunPrint;
 }
 
-export class Sequence {
+export class DeployJob {
```
dOrgJelli, line 23: This could implement the `DeployJob` interface from the generated `polywrap.deploy.yaml` manifest types.

packages/js/manifests/polywrap/src/formats/polywrap.deploy/0.2.0.ts
```diff
@@ -14,40 +14,42 @@ export interface DeployManifest {
...
-       * Name of the step
-       */
-      name: string;
+    [k: string]: {
       /**
-       * Name of the deployer package.
+       * Deployment steps
        */
-      package: string;
+      steps: {
```
dOrgJelli: This should be a type, let's use reference types in the json-schema

dOrgJelli (changes requested):

packages/cli/src/commands/deploy.ts
namesty: Hmm, the thing is that there is a special condition set for the `Workflow` types in the template scripts for those types (literally an `isWorkflow` condition). Even if we wanted to do that here as well (which I tried), we'd have to add another condition so that it only applies for version `0.2` of the DeployManifest, else it will error because version `0.1` did not use ref types. I considered adding that as well but it seems too hacky to me; maybe we could improve type exporting on the manifest packages in a separate PR

packages/cli/src/lib/deploy/sequence.ts
namesty: Same comment about type exporting applies here

packages/cli/src/commands/deploy.ts
dOrgJelli: okay no problem, sounds good!

dOrgJelli (approved):

PR #1273, opened 2022-09-23
Rihp/python client: msgpack tests

rihp:
right now tuples are not being sanitised into lists

error code 

`FAILED tests/test_msgpack.py::test_sanitize_tuple_returns_list - TypeError: list expected at most 1 argument, got 2`


packages/py/polywrap-core/tests/test_get_implementations_spec.py
```diff
@@ -57,4 +57,5 @@ def test_interface_implementations_not_redirected():
 
     get_implementations_results = get_implementations(Uri(interface1_uri), interfaces, redirects)
 
+    ## comment: is this assertion correct using the `in` operand? the Jest version is a `.toEqual` instead.
```
cbrzn: I think this is a good catch and indeed we should be using `==` instead of in 🤔

rihp: updated to use `==` instead 👍

packages/py/polywrap-msgpack/tests/test_msgpack.py
```diff
@@ -71,3 +77,78 @@ def test_generic_map_decode():
...
+def test_sanitize_complex_list_returns_list():
+    complex_list = [1, 'foo', 'bar', 0.123, True, None]
+    assert sanitize(complex_list) == complex_list
+
+def test_sanitize_nested_list_returns_nested_list():
+    nested_list = [23, [[0.123,'dog'], 'cat'], 'boat', ['moon', True]]
+    assert sanitize(nested_list) == nested_list 
+
+def test_sanitize_set_returns_set():
+    assert sanitize({'bob','alice','megan','john'}) == {'bob','alice','megan','john'}
```
Niraj-Kamdar: set should be sanitized to list

packages/py/polywrap-msgpack/tests/test_msgpack.py
```diff
@@ -71,3 +77,78 @@ def test_generic_map_decode():
...
+
+def test_sanitize_complex_number_returns_complex_number():
+    assert sanitize(3 + 5j) == 3 + 5j
```
Niraj-Kamdar: `"3 + 5j"`

packages/py/polywrap-msgpack/tests/test_msgpack.py
```diff
@@ -71,3 +77,78 @@ def test_generic_map_decode():
...
+
+def test_sanitize_single_tuple_returns_list():
+    assert sanitize((8)) == [8] 
+
+def test_sanitize_long_tuple_returns_list():
+    assert sanitize((2,3,6)) == [2,3,6] 
+
+def test_sanitize_nested_tuples_returns_nested_list():
+    nested_tuple = (23, ((0.123,'dog'), 'cat'), 'boat', ('moon', True))
+    assert sanitize(nested_tuple) == nested_tuple
```
Niraj-Kamdar: This should sanitize as an array

Niraj-Kamdar (changes requested):
Add some examples with dataclasses with and without slots just like you have it for dict

Ex: `list of  dataclass`, `dict of dataclass`, etc

packages/py/polywrap-msgpack/tests/test_msgpack.py
rihp: updated 👍

packages/py/polywrap-msgpack/tests/test_msgpack.py
rihp: added 👍

packages/py/polywrap-msgpack/tests/test_msgpack.py
rihp: done !

rihp:
tuples should be working now after some changes to the sanitize function, 

the only error we on this test suite have is regarding the sanitization of sets into lists.

this comes naturally with sanitization, as when you turn a set into a list, the items of the created list might be shuffled when compared to the way you have saved the set initially. 

we could try decoding the encoded list into a set again and assert that probably

packages/py/polywrap-msgpack/polywrap_msgpack/__init__.py
```diff
@@ -21,8 +21,18 @@ def sanitize(value: Any) -> Any:
...
+        return sanitize(array)
+        ##previous implementation below:
+        #array: List[Any] = list(*value)
+        #return [sanitize(a) for a in array]
+    elif type(value) is set:
+        sanitized_set: List[Any] = [i for i in value]
+        return sanitized_set
+    elif type(value) == complex:
+        print(value)
+        print(str(value))
```
Niraj-Kamdar: remove debug leftover

packages/py/conftest.py
```diff
@@ -0,0 +1,214 @@
+
+from typing import Dict
+from pytest import fixture
+from dataclasses import dataclass
```
Niraj-Kamdar: this shouldn't be here but under msgpack/tests directory

packages/py/polywrap-msgpack/tests/test_msgpack.py
```diff
@@ -71,3 +69,119 @@ def test_generic_map_decode():
...
+
+# DATA CLASSES
+
+def test_sanitize_dataclass_object_returns_dict(dataclass_object1, dataclass_object1_as_dict):
+    assert sanitize(dataclass_object1) == dataclass_object1_as_dict
+
+def test_sanitize_list_of_dataclass_objects_returns_list_of_dicts(dataclass_object1, dataclass_object2):
+    print(dataclass_object1)
+    assert sanitize([dataclass_object1, dataclass_object2]) == [dataclass_object1.__dict__, dataclass_object2.__dict__]
+
```
Niraj-Kamdar, line 182: test dict of dataclass returns dict of dict

Niraj-Kamdar (changes requested):
checkout the comments

packages/py/polywrap-msgpack/tests/test_msgpack.py
rihp: added 2 new tests for this case, one for normal dataclass objects and other one without.

packages/py/polywrap-core/.gitignore
```diff
@@ -157,4 +157,6 @@ cython_debug/
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-#.idea/
\ No newline at end of file
+#.idea/
+
+.packages/py/__pycache__
```
cbrzn: I think that inside of `packages/py`, we should create a `.gitignore` file and add `__pycache__`. Lastly, removing this change

packages/py/polywrap-msgpack/polywrap_msgpack/__init__.py
```diff
@@ -21,8 +21,16 @@ def sanitize(value: Any) -> Any:
         array: List[Any] = value
         return [sanitize(a) for a in array]
     elif type(value) is tuple:
-        array: List[Any] = list(*value)
-        return [sanitize(a) for a in array]
+        array: List[Any] = list(value)
+        return sanitize(array)
+        ##previous implementation below:
+        #array: List[Any] = list(*value)
+        #return [sanitize(a) for a in array]
```
cbrzn: Remove this

packages/py/polywrap-msgpack/tests/test_msgpack.py
```diff
@@ -71,3 +69,126 @@ def test_generic_map_decode():
...
+def test_sanitize_complex_dict_returns_sanitized_values():
+    # This test is not passing because when we sanitize a set, the contents are shuffled around.
+    complex_dict = {'name': ['John', 'Doe'],
+        'position':[-0.34478,12.98453],
+        'color': 'green',
+        'age':33,
+        'origin':(0,0),
+        'is_online': True,
+        'pet': None,
+        'friends': {'bob','alice','megan','john'} }
```
cbrzn, line 157: is the attribute `friends` here expected to be a dict or a list? I think it should be changed to a list, right?

packages/py/polywrap-msgpack/tests/test_msgpack.py
```diff
@@ -71,3 +69,126 @@ def test_generic_map_decode():
     decoded = msgpack_decode(encoded)
 
     assert decoded == {'firstKey': 'firstValue', 'secondKey': 'secondValue'}
+
+# STRINGS
+
+def test_sanitize_str_returns_same_str():
+    assert sanitize("https://docs.polywrap.io/") == "https://docs.polywrap.io/"
+
+#def test_sanitized_polywrap_ens_uri(UriPath):
```
cbrzn: Let's remove all the unncessary comments

packages/py/polywrap-core/.gitignore
rihp: done

packages/py/polywrap-msgpack/polywrap_msgpack/__init__.py
rihp: wanted to make sure the new implementation was correct before getting rid of the comment
what was making this not work before, i think, was that we were placing the sanitized list within another list, therefore nestingthem uncessarily

packages/py/polywrap-msgpack/tests/test_msgpack.py
rihp: It's input as a python set (https://www.w3schools.com/python/python_sets.asp), and sets once sanitized are expected to output a lists ([see line 166](https://github.com/polywrap/toolchain/pull/1273/files/a5ba60c83d4b9e45aa14494e50a81a721da9fd17#diff-ce8abd84375041768c63587bc34514c95ac1797881274ef52f433a62b65e4ebfR166))

this though makes the pytest fail at times because the sets dont have an specific order, so when we sanitize them into a list they are scrambled into a newly-ordered list which doesnt always have its items in the same order as we wrote them on the set on Line 158. thus failing every now and then when we `assert sanitize(complex_dict) == sanitized_complex_dict
`

cbrzn: Gotcha! What about converting it into an ordered list and comparing it that way? Hence, guaranteeing that tests wont fail

cbrzn (approved):

packages/py/polywrap-core/tests/test_get_implementations_spec.py
```diff
@@ -57,4 +57,4 @@ def test_interface_implementations_not_redirected():
 
     get_implementations_results = get_implementations(Uri(interface1_uri), interfaces, redirects)
 
-    assert Uri(implementation1_uri) in get_implementations_results
+    assert Uri(implementation1_uri) == get_implementations_results
```
Niraj-Kamdar, line 60: This should be list I suppose

packages/py/polywrap-msgpack/polywrap_msgpack/__init__.py
```diff
@@ -21,8 +21,13 @@ def sanitize(value: Any) -> Any:
         array: List[Any] = value
         return [sanitize(a) for a in array]
     elif type(value) is tuple:
-        array: List[Any] = list(*value)
-        return [sanitize(a) for a in array]
+        array: List[Any] = list(value)
+        return sanitize(array)
+    elif type(value) is set:
+        sanitized_set: List[Any] = [i for i in value]
```
Niraj-Kamdar, line 27: this can be simply: `list(value)`

PR #1277, opened 2022-09-26
Use Result type and throw fewer errors in Client and Core

### LINKED ISSUE #1018: Client: Throw Less Errors
In the JS client + core libraries, we throw errors. This is a hard error handling method to develop against, because you do not know what functions will throw and which will not. Treating all functions as "throwable" is not very realistic.

Instead we can adopt a better strategy, return wrapped result + error objects (aka `Result<T>`).

This is a well known development practice that we use in all of our wasm languages.

### LINKED ISSUE #1245: Use Result type from @polywrap/result instead of InvokeResult
The InvokeResult type returned from invocations is defined as follows:
```typescript
/**
 * Result of an Wrapper invocation.
 *
 * @template TData Type of the invoke result data.
 */
export interface InvokeResult<TData = unknown> {
    /**
     * Invoke result data. The type of this value is the return type
     * of the method. If undefined, it means something went wrong.
     * Errors should be populated with information as to what happened.
     * Null is used to represent an intentionally null result.
     */
    data?: TData;
    /** Errors encountered during the invocation. */
    error?: Error;
}
```
It should be replaced with the Result type from @polywrap/result to get better type checking from TypeScript
```typescript
export type Result<T, E = undefined> =
  | { ok: true; value: T }
  | { ok: false; error: E | undefined };
```

krisbitney:
This PR replaces most thrown errors in the client and core with the `Result` pattern. Notably, `client.invoke` now returns a `Result<TData, Error>`.

Closes https://github.com/polywrap/toolchain/issues/1018, https://github.com/polywrap/toolchain/issues/1245


dOrgJelli (dismissed):

packages/js/client/src/PolywrapClient.ts
```diff
@@ -202,7 +215,7 @@ export class PolywrapClient implements Client {
   >(options: QueryOptions<TVariables, TUri>): Promise<QueryResult<TData>> {
     let result: QueryResult<TData>;
 
-    try {
+    err: try {
```
cbrzn, line 221: Just curious here: What's the reason for this syntax? I am noticing that you are then using a break but we can also return `{ errors: [parseResult.error as Error] }`? Probably I am wrong but just wanted to understand this :P

packages/js/plugins/uri-resolvers/ens-resolver/src/index.ts
```diff
@@ -125,22 +125,22 @@ export class EnsResolverPlugin extends Module<EnsResolverPluginConfig> {
...
           );
         }
 
-        return data;
+        return result.value;
       }
 
       throw Error(
-        `Ethereum.callContractView returned nothing.\nData: ${data}\nError: ${error}`
+        `Ethereum.callContractView returned nothing.\nData: \nError: `
```
cbrzn, line 143: Should we have the `\nData: \nError:`? Might be miss leading but not sure

packages/js/core/src/algorithms/get-implementations.ts
```diff
@@ -24,9 +25,19 @@ export const getImplementations = Tracer.traceFunc(
...
-          ? applyRedirects(interfaceImplementations.interface, redirects)
-          : interfaceImplementations.interface;
+        let fullyResolvedUri: Uri;
+        if (redirects) {
+          const redirectsResult = applyRedirects(
+            interfaceImplementations.interface,
+            redirects
+          );
+          if (!redirectsResult.ok) {
+            continue;
```
nerfZael, line 35: Why are we ignoring the error here instead of returning it?

cbrzn (dismissed):
This is super cool :) Feel free to ignore my comments

packages/js/plugins/logger/src/__tests__/e2e/e2e.spec.ts
```diff
@@ -13,8 +13,8 @@ describe("log method", () => {
       },
     });
 
-    expect(response.error).toBeUndefined();
-    expect(response.data).toBeDefined();
-    expect(response.data).toBe(true);
+    if (!response.ok) fail(response.error);
+    expect(response.value).toBeDefined();
```
nerfZael, line 17: I think this one is redundant

nerfZael:
@krisbitney  PR looks awesome! I left a couple comments and there's some minor fixes in this PR: https://github.com/polywrap/toolchain/pull/1285

packages/js/client/src/PolywrapClient.ts
krisbitney: It's because of the `clearContext` below the try catch statement. I wanted to be sure it still ran since it was running on error before the changes.

packages/js/plugins/uri-resolvers/ens-resolver/src/index.ts
krisbitney: I went back and forth on this. Ultimately I decided not to change the behavior.

packages/js/core/src/algorithms/get-implementations.ts
krisbitney: I was just trying not to change the behavior.

nerfZael: Isn't the previous implementation to throw an error here, but the current to silence it?

krisbitney: if so, I made a mistake!

krisbitney: You're right. The old implementation looked like this:
```
  const fullyResolvedUri = redirects
    ? applyRedirects(interfaceImplementations.interface, redirects)
    : interfaceImplementations.interface;
```
And applyRedirects threw errors. Good catch!!!

krisbitney: I fixed it in this PR: https://github.com/polywrap/toolchain/pull/1288

nerfZael: Great!

PR #1284, opened 2022-09-27
Wasm wrapper fixes and tests

nerfZael:


pileks (dismissed):
LGTM

packages/js/wasm/src/helpers/createWasmWrapper.ts
```diff
@@ -0,0 +1,62 @@
...
+      wasmModuleOrFileReaderOrManifestOptions as GetManifestOptions
+    );
+
+    if (!result.ok) {
+      throw result.error;
+    }
+    return result.value as WasmWrapper;
+  } else if (
+    !fileReaderOrManifestOptions ||
+    (fileReaderOrManifestOptions as GetManifestOptions).noValidate === true ||
```
krisbitney: Can this test be `(fileReaderOrManifestOptions as GetManifestOptions).noValidate !== undefined`?

packages/js/wasm/src/helpers/createWasmWrapper.ts
```diff
@@ -0,0 +1,62 @@
...
+  wasmModuleOrFileReaderOrManifestOptions?:
+    | Uint8Array
+    | IFileReader
+    | GetManifestOptions,
+  fileReaderOrManifestOptions?: IFileReader | GetManifestOptions,
+  manifestOptions?: GetManifestOptions
+): Promise<WasmWrapper> => {
+  if (
+    !wasmModuleOrFileReaderOrManifestOptions ||
+    (wasmModuleOrFileReaderOrManifestOptions as GetManifestOptions)
```
krisbitney, line 19: Can this test be (wasmModuleOrFileReaderOrManifestOptions as GetManifestOptions).noValidate !== undefined? I'm guessing you already thought of that, but I'm just checking. 😄

packages/js/wasm/src/WasmWrapper.ts
```diff
@@ -60,6 +62,43 @@ export class WasmWrapper implements Wrapper {
...
+  ): Promise<WasmWrapper>;
+  static async from(
+    manifestBufferOrFileReader: Uint8Array | IFileReader,
+    wasmModuleOrFileReaderOrManifestOptions?:
+      | Uint8Array
+      | IFileReader
+      | GetManifestOptions,
+    fileReaderOrManifestOptions?: IFileReader | GetManifestOptions,
+    manifestOptions?: GetManifestOptions
+  ): Promise<WasmWrapper> {
```
krisbitney, line 93: Should this return a Result?

krisbitney (dismissed):
It looks good! I added some comments about things that it looks like you've already considered but that I thought were worth checking.

packages/js/wasm/src/helpers/createWasmWrapper.ts
dOrgJelli: This makes sense to me!

packages/js/wasm/src/helpers/createWasmWrapper.ts
dOrgJelli: makes sense to me

packages/js/wasm/src/WasmWrapper.ts
dOrgJelli: Yah I think we should start propagating errors instead of throwing.

dOrgJelli (changes requested):
Kris left two great points around propagating errors via `Result<T, E>` and optimizing the conditional statements. I think these things should be fixed before merging.

packages/js/wasm/src/helpers/createWasmWrapper.ts
nerfZael: Yep, this was a blunder, nice catch!

packages/js/wasm/src/helpers/createWasmWrapper.ts
nerfZael: yes

packages/js/wasm/src/WasmWrapper.ts
nerfZael: This method is intended to be used (and be very ergonomic) when the developer knows it will not fail (or doesn't care) because he's using compiled wrapper binaries (manifest and wasm module buffers).
E.g.
```typescript
  resolver: RecursiveResolver.from([
     ...,
     { uri: "some-uri", wrapper: await WasmWrapper.from(manifestBuffer, wamModule) }
  ])
});
```

If the developer wants a `safe` version (that returs Result) he can use WasmPackage.from().createWrapper() which is less ergonomic but gets the job done :)

PR #1286, opened 2022-09-27
Manifest migration strategy update

### LINKED ISSUE #1211: Project manifest migration strategy update
We should probably rethink our manifest migration strategy. While we can have all of these options set as optional manifest properties, I'd say that we'll likely want to create a proper migration strategy sooner than later, as it will reduce future work.

## Current migration strategy:
Auto-generated code requires us to implement certain, fixed migrators.
```
available manifests:
0.1.0
0.2.0
0.3.0
0.4.0

migrators/migration paths:
0.1.0 -> 0.4.0
0.2.0 -> 0.4.0
0.3.0 -> 0.4.0
```
After update to 0.5.0:
We need to update/create `n+1` migrators if we want full migration support.
```
available manifests:
0.1.0
0.2.0
0.3.0
0.4.0
0.5.0

migrators/migration paths:
0.1.0 -> 0.5.0 #added/modified
0.2.0 -> 0.5.0 #added/modified
0.3.0 -> 0.5.0 #added/modified
0.4.0 -> 0.5.0 #added
```

## Proposed migration strategy:
Allow developers to declare their own migrators, from whichever version to whichever other version.
A simple search algorithm would then find the appropriate migration path, and could even allow for downgrades if the correct migrator exists.

```
available manifests:
0.1.0
0.2.0
0.3.0
0.4.0

migrators:
0.1.0 -> 0.2.0
0.2.0 -> 0.3.0
0.3.0 -> 0.4.0
0.1.0 -> 0.3.0

migration paths to 0.4.0:
0.1.0 -> 0.2.0 -> 0.3.0 -> 0.4.0
0.1.0 -> 0.3.0 -> 0.4.0
```

After update to 0.5.0:

```
available manifests:
0.1.0
0.2.0
0.3.0
0.4.0
0.5.0

migrators:
0.1.0 -> 0.2.0
0.2.0 -> 0.3.0
0.3.0 -> 0.4.0
0.1.0 -> 0.3.0
0.4.0 -> 0.5.0 #added

migration paths to 0.4.0:
0.1.0 -> 0.2.0 -> 0.3.0 -> 0.4.0 -> 0.5.0
0.1.0 -> 0.3.0 -> 0.4.0 -> 0.5.0
```

pileks:
Closes #1211

This PR introduces a developer-defined collection of Migrators, which are simple objects that describe migrations from one version of a manifest format to another.

Now, devs can define both up and down migrations, and adding a new version of a certain manifest will require writing only one migration.

Small implementation detail:
Under the assumption that all migrations have the same "level of complexity", a simple "migration graph" is traversed using BFS and the shortest migration path is returned (or `undefined` if none is found). See issue #1211 for examples.

Possible future work:
- [x] The `polywrap manifest migrate` CLI command could now have an additional optional parameter `-f, --format` wherein the user could specify to which format they want to migrate, the default being the latest format. Create an issue if people agree.


cbrzn (dismissed):
:fire:

packages/js/manifests/polywrap/src/formats/polywrap/migrate.ts
```diff
@@ -10,30 +10,15 @@ import {
...
+import { migrators } from "./migrators";
 
 export function migratePolywrapManifest(
   manifest: AnyPolywrapManifest,
   to: PolywrapManifestFormats
 ): PolywrapManifest {
   let from = manifest.format as PolywrapManifestFormats;
 
-  // HACK: Patch fix for backwards compatability
-  if(from === "0.1" && ("0.1.0" in migrators)) {
```
krisbitney, line 33: After this PR, what will happen if a manifest uses "0.1" as the version?

packages/js/manifests/polywrap/src/migrations/findShortestMigrationPath.ts
```diff
@@ -0,0 +1,52 @@
+import { Migrator } from "./migration";
+
+type MigrationWithSearchHistory = [Migrator, Migrator[]];
+
+// A simple BFS intended to find the shortest migration path from one version to another
+// Returns an array of migrations in the order they need to be performed to migrate from one version to another
+export function findShortestMigrationPath(
```
krisbitney, line 7: I'm curious - did you consider using a stack-based queue to avoid the cost of `Array.shift`, or of using a priority queue and implementing something like Dijkstra's shortest path algorithm? Did you decide the overhead isn't worth it due to small N? 

I'm asking solely out of curiosity. I'm not asking you to change this. I don't think performance will be an issue anytime soon.

packages/js/manifests/polywrap/src/migrations/migration.ts
```diff
@@ -0,0 +1,5 @@
+export type Migrator = {
+  from: string;
+  to: string;
+  migrate: (manifest: unknown) => unknown;
```
krisbitney: Can the `unknown` here be replaced with something like `AnyManifest`?

```typescript
type AnyManifest = AnyPolywrapManifest | AnyPluginManifest | AnyAppManifest | AnyInfraManifest | ...

export type Migrator = {
  from: string;
  to: string;
  migrate: (manifest: AnyManifest) => AnyManifest;
};

```

krisbitney (changes requested):

krisbitney:
Amazing work 🔥. This will be really helpful. Much better than the whole "exponentially increasing number of migrator implementations" situation!

packages/js/manifests/polywrap/src/formats/polywrap/migrate.ts
pileks: You can take a look at e.g. `packages/js/manifests/polywrap/src/formats/polywrap.app/migrators/index.ts` where you will find a migration to `0.2.0` both from `0.1` and `0.1.0`. It's still a slight hack, but it feels less hacky to me, as we are explicitly defining this migration instead of relying on a patch fix inside the `migrate...` function.

packages/js/manifests/polywrap/src/migrations/findShortestMigrationPath.ts
pileks: To be perfectly honest, I went with the simplest (and first) approach that came to mind, and to me, it was "good enough" considering that I believe our `N` will not even reach 10 anytime soon. I didn't want to introduce a whole new data structure just to support the search algorithm at this point too. So yes, the overhead was simply not worth it IMO.

My reasoning behind not using something like Djikstra pretty much comes down to the fact that we are not concerned with any kind of "distance" or "weight" value, hence BFS essentially does the same thing as Djikstra in our case.

A good thing here is that we can simply swap out the algorithm at any time we wish to do so, so if it ever becomes an issue, we're one simple PR away from having it optimized.

packages/js/manifests/polywrap/src/migrations/migration.ts
pileks: Yup, just tried it out, works like a charm!

N.B.: One of us will need to add support for your `codegen` stuff here as well, depending on which one makes it in first!

nerfZael (approved):

packages/js/manifests/polywrap/src/formats/polywrap/migrate.ts
krisbitney: Sounds good!

packages/js/manifests/polywrap/src/migrations/migration.ts
krisbitney: Yeah, for sure. We'll get yours merged first I think. I'm about to approve it. 😁

packages/js/manifests/polywrap/src/migrations/findShortestMigrationPath.ts
krisbitney: I feel you 💯 Your BFS algorithm works well here, and I like how you set it up so the algorithm could be adjusted if that were ever needed. Super cool stuff!

krisbitney (approved):

krisbitney:
P.S. the `--format` option for the manifest command is a cool idea. I think it will sense to have it as we grow and get more users and publish more versions. People will be working with versions of Polywrap.

PR #1287, opened 2022-09-28
Maybe async refactor

nerfZael:


pileks (approved):
LGTM - We should probably dig a little into the TS spec to see what exactly allows us to await a union of `T` and `Promise<T>`, but if it fits I sits.

cbrzn (approved):
:1st_place_medal: looks good!

packages/js/core/src/types/MaybeAsync.ts
```diff
@@ -1,18 +1 @@
 export type MaybeAsync<T> = Promise<T> | T;
```
dOrgJelli, line 1: oh wow, had no clue you could await a non-async function and it'd return just fine.

dOrgJelli (approved):

PR #1288, opened 2022-09-28
Fix: correctly handle error in getImplementations algorithm

krisbitney:
Reverts an accidentally change to the behavior of getImplementations in https://github.com/polywrap/toolchain/pull/1277, where an error that was previous thrown instead became ignored

### DIFF
--- a/packages/js/core/src/algorithms/get-implementations.ts
+++ b/packages/js/core/src/algorithms/get-implementations.ts
@@ -32,7 +32,7 @@ export const getImplementations = Tracer.traceFunc(
             redirects
           );
           if (!redirectsResult.ok) {
-            throw redirectsResult.error;
+            return redirectsResult;
           }
           fullyResolvedUri = redirectsResult.value;
         } else {
--- a/packages/js/core/src/algorithms/get-implementations.ts
+++ b/packages/js/core/src/algorithms/get-implementations.ts
@@ -23,7 +23,7 @@ export const getImplementations = Tracer.traceFunc(
     const addAllImplementationsFromImplementationsArray = (
       implementationsArray: readonly InterfaceImplementations<Uri>[],
       wrapperInterfaceUri: Uri
-    ) => {
+    ): Result<true, Error> => {
       for (const interfaceImplementations of implementationsArray) {
         let fullyResolvedUri: Uri;
         if (redirects) {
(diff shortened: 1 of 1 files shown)


packages/js/core/src/algorithms/get-implementations.ts
```diff
@@ -23,7 +23,7 @@ export const getImplementations = Tracer.traceFunc(
     const addAllImplementationsFromImplementationsArray = (
       implementationsArray: readonly InterfaceImplementations<Uri>[],
       wrapperInterfaceUri: Uri
-    ) => {
+    ): Result<true, Error> => {
```
nerfZael: Is there a need for this to be a Result if in the `ok` case you're returning the data and in the `error` case you're throwing?

packages/js/core/src/algorithms/get-implementations.ts
```diff
@@ -32,7 +32,7 @@ export const getImplementations = Tracer.traceFunc(
             redirects
           );
           if (!redirectsResult.ok) {
-            continue;
+            throw redirectsResult.error;
```
nerfZael: Might be best to just return this error

nerfZael (changes requested):

packages/js/core/src/algorithms/get-implementations.ts
krisbitney: I went back and forth on how to implement this and I meant to return the error instead of throwing it

packages/js/core/src/algorithms/get-implementations.ts
krisbitney: Yes, 100%. Thanks!

packages/js/core/src/algorithms/get-implementations.ts
nerfZael: Since this is always true when it's `ok`, maybe it should be Result<undefined, Error>
Returning would look like: ResultOk(undefined)
Then users could know they can always ignore result.data, though I guess it's pretty much the same if the type is true.
Just a minor nitpick

krisbitney: That's a good point. You're 100% right. I changed it.

nerfZael (approved):

cbrzn (approved):

Issue #1178, opened 2022-08-23
Make the Serde dependency optional in Rust wrappers


krisbitney:
**Is your feature request related to a problem? Please describe.**
Rust wrappers are nearly twice the size of AssemblyScript wrappers

**Describe the solution you'd like**
Make the Serde package (crate) optional. 

See Jordan's comment below for context:
- [ ] Add codegen manifest https://github.com/polywrap/toolchain/issues/1278
- [ ] Add codegen manifest extensions for binding targets https://github.com/polywrap/toolchain/issues/1279
- [ ] Add option to enable or disable serde in Rust codegen manifest https://github.com/polywrap/toolchain/issues/1280

**Additional context**
- allow user to change generated code by giving option to not run codegen while building https://github.com/polywrap/toolchain/issues/796
- Improved JSON Serialization Bindings For AssemblyScript https://github.com/polywrap/toolchain/issues/633

dOrgJelli:
@krisbitney a few quick thoughts.

1. It feels weird to have language-specific annotations within the schema. In my mind, the schema should be "portable" to multiple implementation languages. I understand that the schema is just used at development-time, and what it gets built to is the ABI, but it still feels a bit... odd.
2. I was thinking there could be an easy way in Rust to add traits to a pre-defined struct, maybe by inheriting & adding the serde attributes, but Rust doesn't support inheritance. Instead the developer would have to compose them like so:
```rust
struct BaseNonSerde {
  prop: u8;
}

#[derive(Serialize, Deserialize)]
struct EnableSerde {
  base: BaseNonSerde;
}
```
This is pretty poor UX though...
3. It seems to me that we want the ability to configure the codegen / schema binding process. Why not use the same developer experience we implemented for different build images (see [here](https://github.com/polywrap/toolchain/blob/origin/packages/cli/src/lib/defaults/build-images/wasm/assemblyscript/polywrap.build.ext.json))?

The solution would go like this:
- We would make a new Polywrap manifest for the codegen process (we've been talking about this for a while). Call it `polywrap.codegen.yaml`.
- Each binding language (`wasm/rust`, `wasm/assemblyscript`, `plugin/typescript`, etc) can now define its own manifest extension schema for `polywrap.codegen.yaml`, just like the `polywrap.build.ext.json` file linked above.
[...]

PR #1289, opened 2022-09-29
Codegen manifest

### LINKED ISSUE #1278: Add codegen manifest
Make a new Polywrap manifest for the codegen process. It can be called `polywrap.codegen.yaml`.

This is a precursor to https://github.com/polywrap/toolchain/issues/1279, which will allow each binding language (wasm/rust, wasm/assemblyscript, plugin/typescript, etc) to define its own manifest extension schema for polywrap.codegen.yaml, just like the polywrap.build.ext.json file.

**Additional context**
Parent issue: https://github.com/polywrap/toolchain/issues/1178

### LINKED ISSUE #1279: Add codegen manifest extensions
Within manifest extensions specific to each binding target (e.g. wasm/rust), we can add custom configuration properties for things like enabling or disabling generated code (e.g. disabling serde in Rust bindings). 

These custom configuration properties would be passed into the schema bindings for their respective language, which would allow them modify what code is generated based on them.

**Additional context**
Parent issues: https://github.com/polywrap/toolchain/issues/1178, https://github.com/polywrap/toolchain/issues/1278

krisbitney:
This PR adds a new manifest type: `polywrap.codegen.yaml`. The purpose of the codegen manfiest is to give developers a nice UX for managing binding generation. 

It will be especially useful for target-specific codegen instructions, such as disabling the `serde` package in Rust Wasm wrapper bindings.

The codegen manifest schema currently looks like this: 

```yaml
format: 0.1.0
codegenDir: ./src/wrap # codegen output will go here
skipOnBuild: false # if true, codegen will not be run when building Wasm wrapper
publishDir: ./build # when codegen is run for a plugin, wrap.info will be output here
config: # target-specific config, determined by ext schema
```

I also updated the build manifest with a `bindings` option that allows users to specify a directory other than `./src/wrap` for the location of their generated bindings.

Closes https://github.com/polywrap/toolchain/issues/1278, https://github.com/polywrap/toolchain/issues/1279

Related: https://github.com/polywrap/toolchain/issues/1178


pileks (changes requested):
I see that there were no changes made to the `manifest` command in the CLI. You can take a look and add `codegen` as one of the `manifestTypes` options there - it should be pretty straightforward to add support for it. Feel free to contact me if you end up stuck or with any awesome ideas for the command itself.

krisbitney:
> I see that there were no changes made to the `manifest` command in the CLI. You can take a look and add `codegen` as one of the `manifestTypes` options there - it should be pretty straightforward to add support for it. Feel free to contact me if you end up stuck or with any awesome ideas for the command itself.

Thanks for pointing this out. The manifest command is new and I haven't used it yet (but it looks amazing!). I totally spaced on updating it. I just made a commit that includes the update.

pileks (approved):

packages/js/client/src/__tests__/e2e/wasm-rs.spec.ts
```diff
@@ -269,6 +269,11 @@ describe("wasm-rs test cases", () => {
       wrapperUri
     );
   });
+
+  it("codegen-manifest", async () => {
```
pileks, line 298: I assume this basically just runs the `buildWrapper` function and the test is that it won't throw?

packages/cli/src/lib/project/AppProject.ts
```diff
@@ -99,19 +105,84 @@ export class AppProject extends Project<AppManifest> {
...
       bindLanguage: appManifestLanguageToBindLanguage(
         await this.getManifestLanguage()
       ),
+      config: codegenManifest?.config,
     });
   }
 
-  private _getGenerationDirectory(generationSubPath = "src/wrap"): string {
-    return path.join(this.getManifestDir(), generationSubPath);
+  /// Polywrap Codegen Manifest (polywrap.build.yaml)
```
pileks: wouldn't this be:
```typescript
/// Polywrap Codegen Manifest (polywrap.codegen.yaml)
```
? 🙂

pileks (dismissed):
Overall, apart from two *minor* comments I left, this looks awesome!

One nitpick that came to mind is that there was some duplicated code in `PluginProject`, `AppProject` and `PolywrapProject`. Namely, all the `codegenManifest` related functions seem like they could be extracted into the parent `Project` abstract class.

This might also apply to other extension manifests, so it could very easily be a separate issue/PR.

packages/js/client/src/__tests__/e2e/wasm-rs.spec.ts
krisbitney: Yes, exactly. It wouldn't build if it wasn't working. Instead of generating code bindings to the "./src/wrap" folder, it's generating code in './src/wrapx' and reading from there.

packages/cli/src/lib/project/AppProject.ts
krisbitney: Yes, 100%. Thanks!

krisbitney:
> Overall, apart from two _minor_ comments I left, this looks awesome!
> 
> One nitpick that came to mind is that there was some duplicated code in `PluginProject`, `AppProject` and `PolywrapProject`. Namely, all the `codegenManifest` related functions seem like they could be extracted into the parent `Project` abstract class.
> 
> This might also apply to other extension manifests, so it could very easily be a separate issue/PR.

Yeah, definitely. I moved the codegen manifest methods and config props to the `Project` abstract class. 

I looked at the other extensions, but I left their methods in the `PolywrapProject`. All the extensions are in all of the project manifests, but I think right now we only use the other ones for Wasm wrappers (`build`, `deploy`, `meta`). Do you think that makes sense, or does it make more sense to move the methods to `Project`?

packages/js/client/src/__tests__/e2e/wasm-rs.spec.ts
pileks: IDK if this would impact anything deeper, but a nice little piece of syntactic sugar could be that we explicitly say "Hey, we expect that this won't throw.
Something like:
```typescript
expect(async () => {
    const wrapperPath = `${GetPathToTestWrappers()}/wasm-rs/codegen-manifest`;
    await buildWrapper(wrapperPath);
}).not.toThrow();
```
Total nitpick, we're just making sure that it's immediately obvious from the code what we're expecting.

pileks:
> Yeah, definitely. I moved the codegen manifest methods and config props to the Project abstract class.
>I looked at the other extensions, but I left their methods in the PolywrapProject. All the extensions are in all of the project manifests, but I think right now we only use the other ones for Wasm wrappers (build, deploy, meta). Do you think that makes sense, or does it make more sense to move the methods to Project?

I believe we're more than good with just this as part of this PR. I agree with your sentiment!

pileks (dismissed):

packages/js/manifests/polywrap/src/formats/polywrap.app/migrators/index.ts
```diff
@@ -1,15 +1,27 @@
...
-    migrate: migrate_0_1_0_to_0_2_0
-  }
-];
\ No newline at end of file
+    migrate: migrate_0_1_0_to_0_2_0,
+  },
+  {
+    from: "0.1.0",
+    to: "0.3.0",
+    migrate: migrate_0_1_0_to_0_3_0,
```
pileks, line 20: Since there already is a migration path from 0.1.0 to 0.3.0 through 0.2.0, this one is essentially superfluous, but can remain here as a "shortcut" option! 😄 
Not a blocker.

packages/js/manifests/polywrap/src/formats/polywrap.plugin/migrators/index.ts
```diff
@@ -1,15 +1,27 @@
...
-    migrate: migrate_0_1_0_to_0_2_0
-  }
-];
\ No newline at end of file
+    migrate: migrate_0_1_0_to_0_2_0,
+  },
+  {
+    from: "0.1.0",
+    to: "0.3.0",
+    migrate: migrate_0_1_0_to_0_3_0,
```
pileks, line 20: Same as previous comment - essentially superfluous, but doesn't hurt!

packages/js/manifests/polywrap/src/formats/polywrap/migrators/index.ts
```diff
@@ -1,15 +1,27 @@
...
-    migrate: migrate_0_1_0_to_0_2_0
-  }
-];
\ No newline at end of file
+    migrate: migrate_0_1_0_to_0_2_0,
+  },
+  {
+    from: "0.1.0",
+    to: "0.3.0",
+    migrate: migrate_0_1_0_to_0_3_0,
```
pileks, line 20: Same as previous comment.

pileks (approved):
LGTM, all seems ready for a merge!

packages/js/manifests/polywrap/src/formats/polywrap.app/migrators/index.ts
krisbitney: I added it before the migrations rework and just kept it because it's there 🤷‍♂️

packages/js/manifests/polywrap/src/formats/polywrap.plugin/migrators/index.ts
krisbitney: Same as other one--I added it before the migrations rework and just kept it because it's there 🤷‍♂️

packages/js/manifests/polywrap/src/formats/polywrap/migrators/index.ts
krisbitney: Same thing again! I added it before the migrations rework and just kept it because it's there 🤷‍♂️

packages/js/manifests/polywrap/src/formats/polywrap.app/migrators/index.ts
pileks: Totally valid and it's a shortcut, so we're good!

packages/js/manifests/polywrap/src/formats/polywrap.codegen/0.1.0.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export interface CodegenManifest {
+  /**
+   * Polywrap codegen manifest format version.
+   */
+  format: "0.1.0" | "0.1";
+  /**
+   * Output directory for the generated code (default: ./src/wrap)
+   */
+  codegenDir?: string;
```
dOrgJelli, line 17: Camel case isn't use in any of our other schemas, AFAIK we've been using snake case (see `import_abis`).

Also I think we should work on the naming of these properties to be a bit more ergonomic and grouped by functionality. For example, you could group `codegenDir` and `publishDir` into a single root property which is `output`.

For example, these could be valid:
```yaml
output:
  types: ...
  manifest: ...
```

The existing naming is pretty bad :P sorry about that

dOrgJelli (changes requested):

krisbitney:
It turns out Serde is already successfully removed by wasm-snip, and this was the original motivation for this feature. Is there any interest in this for its own sake, or should we scrap it for now and revisit it in the future if there is a need?

krisbitney:
It turns out Serde is already successfully removed by wasm-snip, and this was the original motivation for this feature. Closing for now. We can revisit the idea in the future if there is a reason to.

PR #1290, opened 2022-09-29
Client (Python): Fix circular dependency in core package

rihp:


rihp:
having some time figuring our where the error is generating so tried to graph the dependencies and this is what we have right now at '/polywrap-core/polywrap_core/__init__.py'

polywrap_core___init__

cbrzn:
Interesting! @rihp 

Based on this, I would first try to remove all the explicit exports from the `__init__` files and import everything by hand, then add one by one import to see where it breaks. If this doesn't make sense feel free to DM :-D

cbrzn (approved):

PR #1291, opened 2022-09-30
Make config props read only

### LINKED ISSUE #1282: JS Client: Make Config Props Read-Only
**Describe the bug**
Currently the client's configuration is not read-only, meaning that a user that's creating a new client instance could hang onto a reference of the config, and modify it during the client's run-time.

This is problematic because the client could have multiple asynchronous invocations that require details from the config. Modifying the configuration could lead to undefined behavior.

pileks:
This PR resolves #1282 

Both ClientConfig and PolywrapClientConfig are now `readonly` with all their properties also being marked as `readonly`.

This essentially naively prevents users from modifying the client config.

### DIFF
--- a/packages/js/client/src/PolywrapClient.ts
+++ b/packages/js/client/src/PolywrapClient.ts
@@ -43,11 +43,14 @@ import { Tracer, TracerConfig, TracingLevel } from "@polywrap/tracing-js";
 import { ClientConfigBuilder } from "@polywrap/client-config-builder-js";
 import { Result, ResultErr, ResultOk } from "@polywrap/result";
 
-export interface PolywrapClientConfig<TUri extends Uri | string = string>
+interface PolywrapClientConfigInternal<TUri extends Uri | string = string>
   extends ClientConfig<TUri> {
-  tracerConfig: Partial<TracerConfig>;
-  wrapperCache?: IWrapperCache;
+  tracerConfig: Readonly<Partial<TracerConfig>>;
+  wrapperCache?: Readonly<IWrapperCache>;
 }
+
+export interface PolywrapClientConfig<TUri extends Uri | string = string>
+  extends Readonly<PolywrapClientConfigInternal<TUri>> {}
 
 export class PolywrapClient implements Client {
   private _config: PolywrapClientConfig<Uri> = ({
@@ -111,7 +114,10 @@ export class PolywrapClient implements Client {
     } else {
       Tracer.disableTracing();
     }
-    this._config.tracerConfig = tracerConfig ?? {};
+    this._config = {
+      ...this._config,
+      tracerConfig: tracerConfig ?? {},
+    };
   }
 
   @Tracer.traceMethod("PolywrapClient: getRedirects")
--- a/packages/js/client/src/PolywrapClient.ts
+++ b/packages/js/client/src/PolywrapClient.ts
@@ -43,16 +43,13 @@ import { Tracer, TracerConfig, TracingLevel } from "@polywrap/tracing-js";
 import { ClientConfigBuilder } from "@polywrap/client-config-builder-js";
 import { Result, ResultErr, ResultOk } from "@polywrap/result";
 
-type PolywrapClientConfigInternal<
+export type PolywrapClientConfig<
   TUri extends Uri | string = string
-> = ClientConfig<TUri> & {
-  tracerConfig: Readonly<Partial<TracerConfig>>;
-  wrapperCache?: Readonly<IWrapperCache>;
-};
-
-export type PolywrapClientConfig<TUri extends Uri | string = string> = Readonly<
-  PolywrapClientConfigInternal<TUri>
->;
+> = ClientConfig<TUri> &
+  Readonly<{
+    tracerConfig: Readonly<Partial<TracerConfig>>;
+    wrapperCache?: Readonly<IWrapperCache>;
+  }>;
 
 export class PolywrapClient implements Client {
   private _config: PolywrapClientConfig<Uri> = ({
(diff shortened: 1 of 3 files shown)


packages/js/core/src/types/Client.ts
```diff
@@ -15,13 +15,17 @@ import { UriResolverHandler } from "./UriResolver";
...
 import { Result } from "@polywrap/result";
 
-export interface ClientConfig<TUri extends Uri | string = string> {
-  redirects: UriRedirect<TUri>[];
-  plugins: PluginRegistration<TUri>[];
-  interfaces: InterfaceImplementations<TUri>[];
-  envs: Env<TUri>[];
-  resolver: IUriResolver<unknown>;
-}
+type ClientConfigInternal<TUri extends Uri | string = string> = {
```
krisbitney: Instead of having two types, can we combine it into one type like this?
```typescript
export type ClientConfig<TUri extends Uri | string = string> = Readonly<{
  redirects: readonly UriRedirect<TUri>[];
  plugins: readonly PluginRegistration<TUri>[];
  interfaces: readonly InterfaceImplementations<TUri>[];
  envs: readonly Env<TUri>[];
  resolver: Readonly<IUriResolver<unknown>>;
}>;
```
This is just a stylistic preference on my part, so feel free to ignore this comment!

packages/js/client/src/PolywrapClient.ts
```diff
@@ -43,11 +43,16 @@ import { Tracer, TracerConfig, TracingLevel } from "@polywrap/tracing-js";
 import { ClientConfigBuilder } from "@polywrap/client-config-builder-js";
 import { Result, ResultErr, ResultOk } from "@polywrap/result";
 
-export interface PolywrapClientConfig<TUri extends Uri | string = string>
-  extends ClientConfig<TUri> {
-  tracerConfig: Partial<TracerConfig>;
-  wrapperCache?: IWrapperCache;
-}
+type PolywrapClientConfigInternal<
```
krisbitney: Is it possible to declare only one type instead of two here? Maybe something like this:

```typescript
export type PolywrapClientConfig<
  TUri extends Uri | string = string
> = ClientConfig<TUri> &
  Readonly<{
    tracerConfig: Readonly<Partial<TracerConfig>>;
    wrapperCache?: Readonly<IWrapperCache>;
  }>;
```

This is just a stylistic preference on my part, so feel free to ignore this comment!

packages/js/core/src/types/Client.ts
krisbitney: Also, we use `interface` elsewhere in the client and core. I think it makes sense for us to maintain the same code style throughout, even if that means we can't use the `Readonly<>` syntax.

```typescript
export interface ClientConfig<TUri extends Uri | string = string> {
  readonly redirects: readonly UriRedirect<TUri>[];
  readonly plugins: readonly PluginRegistration<TUri>[];
  readonly interfaces: readonly InterfaceImplementations<TUri>[];
  readonly envs: readonly Env<TUri>[];
  readonly resolver: Readonly<IUriResolver<unknown>>;
}
```

packages/js/client/src/PolywrapClient.ts
krisbitney: Also, same comment here about using `interface` instead of `type`

krisbitney (changes requested):

packages/js/core/src/types/Client.ts
pileks: Regarding your stylistic preference, your approach is a strictly better solution, and I'm going with it. The style I have here right now is mostly a remnant of some experimentation from before. TBH I'm amazed that it didn't come to mind when I was working on this...

There are notably two issues with `interface` which led me to change it over to `type`:
- First, as you mentioned, is that we can't make an interface `readonly` by definition. This would allow for something like the following to be valid code:
    ```typescript
    const a = new ClientConfigBuilder().addDefaults().build();

    a.envs = [];
    ```
    Although `envs` is a `ReadOnlyArray<Env>`, the `envs` property itself can be reassigned.    I wanted to prevent this, and using `type` allowed me to neatly solve this.
- I have attempted to export an `interface` which `extends` an internal interface (which is not exported). 
    This syntax worked, but produced a linter error (a valid one - `no-empty-interface`):
    ```typescript
    export interface Interface extends Readonly<InterfaceInternal> {};
    ```
    However, when fixing the lint error with:
    ```typescript
    export interface Interface extends Readonly<InterfaceInternal>;
    ```
    The compiler was having trouble with determining the underlying type of the interface (since `InterfaceInternal` was not exported).
    I've gone over some issues on the TS GitHub, and while there are some TS settings that can be fiddled with (amongst a lithany of hacks that also "make it work"), I've kind of figured that this is either a bug or simply not supported. 
    I've come to the conclusion that  I'd rather avoid using interfaces overall in this case, especially as I believe that, in this case, `interface` and `type` are essentially interchangeable, with the only real difference being that interfaces support declaration merging (something like `partial` in C#), which is an anti-pattern in my eyes.

Sorry for the long reply, I really wanted to give context to this and I'd love to discuss this further if you have any ideas or simply disagree with me! :smile:

packages/js/client/src/PolywrapClient.ts
pileks: See previous comment, good catch!

packages/js/core/src/types/Client.ts
krisbitney: The `Readonly<>` type wrapper is the same as adding `readonly` to each top-level property. The `Readonly<>` type is just syntax sugar. So the interface example I shared is equivalent.

https://www.typescriptlang.org/docs/handbook/utility-types.html#readonlytype

krisbitney (dismissed):
Amazing work!

krisbitney (approved):

cbrzn (approved):
lgtm :green_apple:

PR #1305, opened 2022-10-07
Fix: check for fallback providers in IPFS plugin

krisbitney:
A test was checking if IpfsOptions was undefined, but it is always defined. The intended behavior appears to be to check for the existence of fallback providers. If they are not present, the invocation can be simplified.


dOrgJelli (approved):
Great catch!

PR #1316, opened 2022-10-10
Hotfix CLI hanging behavior

pileks:
After the removal of our spinner, if a CLI command didn't explicitly call `process.exit()`, I have encountered a behavior where the CLI would just hang (simply not exit) after it completed the command.

This PR fixes this behaviour, and should probably be hotfixed into the latest release of Polywrap, as it's a behavior reproducible on `0.9.0`.

To reproduce the faulty behavior,  `polywrap create` the `wasm-as` template with the current `0.9.0` release of Polywrap, and then try either `polywrap codegen` or `polywrap build`.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -11,5 +11,6 @@ export const run = async (argv: string[]): Promise<void> => {
     }
   }
 
-  program.parse(argv);
+  await program.parseAsync(argv);
+  process.exit();
 };
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -12,5 +12,4 @@ export const run = async (argv: string[]): Promise<void> => {
   }
 
   await program.parseAsync(argv);
-  process.exit();
 };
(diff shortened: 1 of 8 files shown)


dOrgJelli:
Hmmmm, while I agree with the problem that needs to be fixed, the CI seems to have been "nuked" 😆 @pileks would you mind looking into why this is?

cbrzn:
huh - interesting error, wonder how this slipped to origin

pileks:
~~@dOrgJelli @cbrzn I see that you guys have re-run the CI and everything went smoothly (apart from the `Test-Cli` part). I ran the test on 2 machines - my PC and a VM which I use from my iPad and both seem to have worked fine. It does seem like it was something temporary in nature?~~

pileks:
Disregard previous comment, I was running `build` instead of `test` locally. I can confirm that the `create` command has some wonky behavior and isn't exiting properly. Discussed this with @cbrzn and he is preparing a fix.

cbrzn (approved):

PR #1317, opened 2022-10-11
Check for null when using UriResolverExtensionFileReader to read wasm module

krisbitney:
When a Wasm wrapper dev tries to invoke an interface in their wrapper, an error isn't thrown in the `wasm-js` package. Instead of returning a UInt8Array containing a Wasm pacakge, the `_getWasmModule` method in WasmWrapper.ts returns `null`. The` null` value is passed around until a null pointer exception is thrown when trying to call `indexOf` on it in the `indexOfArray` method in `utils.ts` in the `asyncify-js` package. 

The null value arrises because the `UriResolverExtensionFileReader` in the `uri-resolver-extensions` package wants to return a UInt8Array or an error. However, the`getFile` method of the uri resolver interface returns `Bytes | null`. All of the uri resolver implementations return null on failure. So we need to check that the retrieved file data is not null.

### DIFF
--- a/packages/js/uri-resolver-extensions/src/UriResolverExtensionFileReader.ts
+++ b/packages/js/uri-resolver-extensions/src/UriResolverExtensionFileReader.ts
@@ -34,9 +34,12 @@ export class UriResolverExtensionFileReader implements IFileReader {
       path
     );
     if (!result.ok) return result;
-    if (!result.value) {
+    if (result.value === undefined) {
       return ResultErr(new Error(`File not found at ${path}`));
     }
-    return result;
+    return {
+      value: result.value,
+      ok: true
+    };
   }
 }
--- a/packages/js/uri-resolver-extensions/src/UriResolverExtensionFileReader.ts
+++ b/packages/js/uri-resolver-extensions/src/UriResolverExtensionFileReader.ts
@@ -18,6 +18,7 @@ export class UriResolverExtensionFileReader implements IFileReader {
   ) {}
 
   async readFile(filePath: string): Promise<Result<Uint8Array, Error>> {
+    const path = combinePaths(this.wrapperUri.path, filePath);
     const result = await UriResolverInterface.module.getFile(
       {
         invoke: <TData = unknown, TUri extends Uri | string = string>(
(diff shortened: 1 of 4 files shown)


dOrgJelli (dismissed):
Great find!

packages/js/uri-resolver-extensions/src/UriResolverExtensionFileReader.ts
```diff
@@ -32,5 +32,10 @@ export class UriResolverExtensionFileReader implements IFileReader {
       this.resolverExtensionUri,
       combinePaths(this.wrapperUri.path, filePath)
     );
+    if (!result.ok) return result;
+    if (!result.value) {
+      return ResultErr(new Error(`File not found at ${filePath}`));
```
nerfZael: Swapping `filePath` for `combinePaths(this.wrapperUri.path, filePath)` would give a more descriptive error message.

packages/js/uri-resolver-extensions/src/UriResolverExtensionFileReader.ts
```diff
@@ -18,7 +18,7 @@ export class UriResolverExtensionFileReader implements IFileReader {
   ) {}
 
   async readFile(filePath: string): Promise<Result<Uint8Array, Error>> {
-    return await UriResolverInterface.module.getFile(
+    const result = await UriResolverInterface.module.getFile(
```
nerfZael, line 22: We should change the signature of the getFile method here to return `Promise<Result<Uint8Array | undefined, Error>>` instead of `Promise<Result<Uint8Array, Error>>`.
This mimicks the URI resolver interface signature

nerfZael (changes requested):

packages/js/uri-resolver-extensions/src/UriResolverExtensionFileReader.ts
dOrgJelli: Should we do this in a separate PR @nerfZael?

nerfZael: I think it can be a part of this PR, since it's a one line change that doesn't impact much and most importantly, it addresses the line below `if (!result.value)` which doesn't make sense without it since according to the signature, `result.value` can never be falsy.

dOrgJelli: Ah okay I see what you meant, the typings for the `UriResolverInterface` are incorrect, fixed!

PR #1320, opened 2022-10-12
fix: proper view method result parsing

dOrgJelli:
closes: https://github.com/gelatodigital/polywrap-resolver/issues/9


packages/js/plugins/ethereum/src/utils/parsing.ts
```diff
@@ -26,41 +24,23 @@ export function constructAbi(method: string): any[] {
...
-  return JSON.stringify(objects);
-}
-
-function stringifySimpleArray(result: any): string {
-  const objects: string[] = [];
-  for (const element of result) {
-    objects.push(element.toString());
+    return prepared;
+  } else if (
+    typeof value === "object" &&
```
nerfZael, line 45: What are some examples where this is useful?

nerfZael (dismissed):

packages/js/plugins/ethereum/src/utils/parsing.ts
dOrgJelli: This is necessary because ethers.js utilizes the `BigNumber` object to represent numbers larger than JS' `number` range (ex: `uint256`). When you simply `JSON.stringify(...)` a `BigNumber` the result is not desirable and just shows the internals of the object. This is because `BigNumber` does not implement a `toJSON()` method. So instead, we leverage object's `toString()` method if it's available.

PR #1328, opened 2022-10-13
feat: add wrapper subinvoke test case

Niraj-Kamdar:


cbrzn (approved):

packages/test-cases/cases/wrappers/wasm-as/simple-subinvoke/invoke/src/index.ts
```diff
@@ -1,5 +1,10 @@
...
 
-export function add(args: Args_add): string {
-  Add_Module.add
+export function add(args: Args_add): i32 {
+  let importedArgs: ImportedArgs_add = {
+    a: args.a,
+    b: args.b
+  }
+  return Add_Module.add(importedArgs).unwrap()
 }
```
Niraj-Kamdar, line 10: Why this change? I'd prefer to have a different output for invoke and subinvoke. It'll be easy to debug and spot some horrible bugs that would be passed safely due to a buggy client implementation.

cbrzn: I changed it just to make it work and have a super simple example

>  I'd prefer to have a different output for invoke and subinvoke

Gotcha, if we just use the `add` method in the `invoke` and add `1` to that result, would that be enough for you? Meaning that the `invoke` method would return the sum of two numbers + 1, and the `subinvoke` would just return the sum of two numbers

Niraj-Kamdar: Yeah, that'd work. You can call it `add_and_increament` then

dOrgJelli (approved):

Niraj-Kamdar:
@dOrgJelli I had requested changes that weren't addressed yet. @cbrzn Can you create a patch PR for this?

cbrzn:
@Niraj-Kamdar yah no worries is on my todo list

PR #1336, opened 2022-10-17
CLI Reference section

### LINKED ISSUE #229: Polywrap CLI Reference Section
As discussed, this will be in the form of a README in the CLI's package.

pileks:
Remaining work:

- [x] expand `run` command / get feedback from owner ( @Niraj-Kamdar )
- [x] expand `infra` command / get feedback from owner
- [x] expand `deploy` command / get feedback from owner
- [x] ~~expand on `codegen`'s `-s, --script` option~~ Seems relatively unused and not documented elsewhere - can do as a separate PR if need be.
- [x] ~~generally expand on the `-c, --client-config` option~~ Wherever it’s needed, has links to the explanation

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -115,7 +115,6 @@ This command generates types and bindings for your project based on your project
 - `-s, --script <path>`
   Path to a custom generation script (JavaScript | TypeScript).
   This script is run in place of the standard codegen script if provided.
-  **TODO: Add details about the custom codegen script**
 
 - `-c, --client-config <config-path>`
   Use a custom Polywrap Client configuration.
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -78,7 +78,6 @@ This outputs the project's ABI schema (Wasm and Interface) and binary package (W
 
 - `-c, --client-config <config-path>`
   Use a custom Polywrap Client configuration.
-  See [The `-c, --client-config` option](#the--c,---client-config-option) for details
 
 - `-n, --no-codegen`
   Don't perform codegen before building.
@@ -119,7 +118,6 @@ This command generates types and bindings for your project based on your project
 
 - `-c, --client-config <config-path>`
   Use a custom Polywrap Client configuration.
-  See [The `-c, --client-config` option](#the--c,---client-config-option) for details
 
 #### Special note
 
@@ -270,7 +268,6 @@ polywrap test [options]
 
 - `-c, --client-config <config-path>`
   Use a custom Polywrap Client configuration.
-  See [The `-c, --client-config` option](#the--c,---client-config-option) for details
 
 - `-o, --output-file <output-file-path>`
   Specify the output file path for the workflow result
@@ -428,7 +425,6 @@ polywrap docgen <action>
 
 - `-c, --client-config <config-path>`
   Use a custom Polywrap Client configuration.
-  See [The `-c, --client-config` option](#the--c,---client-config-option) for details
 
 - `-i, --imports`
   Generate docs for your project's dependencies as well.
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -64,9 +64,34 @@ polywrap create --help
 
 ### `build | b`
 
-Build Polywrap projects.
+Build Wasm and Interface Polywrap projects.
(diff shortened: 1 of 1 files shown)


packages/cli/README.md
```diff
@@ -1,13 +1,522 @@
...
+
+Deploy Polywrap projects.
+
+```bash
+polywrap deploy
+```
+
+`deploy` reads the Deploy manifest (`polywrap.deploy.yaml` by default) and executes the jobs and steps listed inside.
+
+For more information on the Deploy command and the Deploy manifest, see [Configure Polywrap deployment pipeline](https://docs.polywrap.io/quick-start/build-and-deploy-wasm-wrappers/deploy-pipeline).
```
cbrzn, line 189: This is probably not related to this PR, but I think the link this is pointing to is outdated (for manifest `0.1`) - The new deploy manifest version doesn't support `stages` but rather `jobs`. Check it here: https://github.com/polywrap/toolchain/blob/origin-0.10-dev/packages/test-cases/cases/cli/wasm/deploy/001-sanity/polywrap.deploy.yaml

packages/cli/README.md
```diff
@@ -1,13 +1,522 @@
...
+polywrap infra up --modules=eth-ens-ipfs
+```
+
+The default infrastructure module defines a docker container with:
+
+- A test server at http://localhost:4040
+- A Ganache Ethereum test network at http://localhost:8545
+- An IPFS node at http://localhost:5001
+
+It also sets up ENS smart contracts at initialization, so you can build wrappers and deploy them to an ENS registry on your locally hosted testnet. The Ethereum address of the ENS registry is 0xe78A0F7E598Cc8b0Bb87894B0F60dD2a88d6a8Ab.
```
cbrzn: Should we point the user to [all addresses available](https://github.com/polywrap/toolchain/blob/origin-0.10-dev/packages/js/test-env/src/index.ts#L15-L20) rather than just mentioning the registry?

packages/cli/README.md
```diff
@@ -1,13 +1,522 @@
...
+
+### `test | t`
+
+Run Test manifests. 
+
+The `test` command executes a series of Wrapper invocations called **steps** organized into **jobs**.
+All steps within a job are run in series, while jobs are run in parallel.
+
+```bash
+polywrap run [options]
```
cbrzn:
```suggestion
polywrap test [options]
```

packages/cli/README.md
```diff
@@ -1,13 +1,522 @@
...
+  Use a custom Polywrap Client configuration.
+
+- `-o, --output-file <output-file-path>`
+  Specify the output file path for the workflow result
+
+- `-j, --jobs <jobs...>`
+  Specify ids of jobs that you want to run
+
+#### The Test manifest (`polywrap.test.yaml`)
+Basic structure:
```
cbrzn, line 279: I have mixed feelings because this is super cool, but in other commands when it comes to manifest we redirect users to documentation - Should that happen here too?

cbrzn:
Should we add an example of how a custom client config file looks like? Usually, I would go to `test-cases` but would be cool if the dev, when reading this for the first time, have an idea of how to do things? Just thinking out loud feel free to ignore this hehe

pileks:
That was my thought initially too, but then I linked to a few places with the custom client config.

Now that you commented the same thing, I think I could add it where it belongs.

Maybe a separate "Custom client config" mini-section, with the various options linking to it.

packages/cli/README.md
pileks: You're absolutely correct, I'll write an updated version within the docs.

pileks: https://github.com/polywrap/documentation/pull/233

I've decided to update it directly within the docs - figured it would be doubly-useful.

packages/cli/README.md
pileks: I agree that this would be much better suited to live within the `docs` repo.

Now that I look at it, https://docs.polywrap.io/quick-start/workflows/running-workflows seems outdated. Maybe rename it to `Runnning tests` and update with information found here.

What do you think?
CC @dOrgJelli @krisbitney

krisbitney: That's a good idea. I think it should also be added to the manifest package readme so we can prepare for the future when our reference documentation is created by copying the readmes into the docs

dOrgJelli (approved):

PR #1340, opened 2022-10-17
Feat: Client validate method

### LINKED ISSUE #1319: `client.validate(uri)` - Validate Wrapper->Client Compatibility
Currently, integrators of wrappers do not have an easy way to determine if a wrapper will run successfully within the client's current configuration. We need a simple method to help answer the question "can this client run this wrapper?"

Ahead-of-time dependency availability detection for wrappers, helping ensure wrappers will execute successfully without any dependency related errors (not found, mismatched abi, etc).

Available through a simple `client.validate(...)` function, that internally does things like:
- resolves wrapper
- resolves all dependencies
- ensures dependencies are the correct ABI that the wrapper is expecting

NOTE: we can provide feature toggles to toggle on/off various validation checks.

cbrzn:
This PR aims to add the `validate` function to the client, being able to guarantee that the client can communicate with the given wrapper URI ahead-of-time

The interface is the following:
```typescript
interface ValidateOptions {
  recursive: boolean
  abi: boolean
}
validate(uri: Uri | string, options: ValidateOptions) -> boolean
```
- If `abi` is true, make sure the ABI from 1st level dependencies are available (the ones from `importedModuleType`)
- If `recursive` is true, it will check all uris and make sure it can resolve them

Also, a compare function has been implemented in `js/manifests/wrap` to compare two array of method definitions, this function allow us to make sure that dependencies are equal and wont fail in runtime


packages/js/client/src/PolywrapClient.ts
```diff
@@ -542,6 +545,77 @@ export class PolywrapClient implements Client {
...
+  public async validate<TUri extends Uri | string>(
+    uri: TUri,
+    options: ValidateOptions
+  ): Promise<Result<boolean, Error>> {
+    const wrapper = await this.loadWrapper(Uri.from(uri));
+    if (!wrapper.ok) {
+      return ResultErr(new Error(wrapper.error?.message));
+    }
+
+    const manifest = await this.getManifest(uri);
```
krisbitney: You could call `wrapper.getManifest` here

packages/js/client/src/PolywrapClient.ts
```diff
@@ -542,6 +545,77 @@ export class PolywrapClient implements Client {
...
+      const importUri = (importedModuleType: ImportedModuleDefinition) => {
+        return this.tryResolveUri({ uri: importedModuleType.uri });
+      };
+      const resolvedModules = await Promise.all(importedModules.map(importUri));
+      const moduleNotFound = resolvedModules.find(
+        (m: { ok: boolean }) => !m.ok
+      );
+
+      if (moduleNotFound) {
+        return ResultErr((moduleNotFound as { error: Error }).error);
```
krisbitney: It would be cool to return errors for all the modules that weren't found, rather than just one

packages/js/client/src/PolywrapClient.ts
```diff
@@ -542,6 +545,77 @@ export class PolywrapClient implements Client {
...
+      const abi = manifest.value.abi;
+      const importedModules: ImportedModuleDefinition[] =
+        abi.importedModuleTypes || [];
+
+      const importUri = (importedModuleType: ImportedModuleDefinition) => {
+        return this.tryResolveUri({ uri: importedModuleType.uri });
+      };
+      const resolvedModules = await Promise.all(importedModules.map(importUri));
+      const moduleNotFound = resolvedModules.find(
+        (m: { ok: boolean }) => !m.ok
```
krisbitney: Depending on your preference, you could use the same syntax as below to get this on one line `resolvedModules.find(({ ok }) => !ok);`

packages/js/client/src/PolywrapClient.ts
```diff
@@ -542,6 +545,77 @@ export class PolywrapClient implements Client {
...
+      }
+
+      if (options.recursive) {
+        const validateImportedModules = importedModules.map(({ uri }) =>
+          this.validate(uri, options)
+        );
+        const invalidUris = await Promise.all(validateImportedModules);
+        const invalidUri = invalidUris.find(({ ok }) => !ok);
+        if (invalidUri) {
+          return ResultErr((invalidUri as { error: Error }).error);
```
krisbitney: Same as above, it would be cool to return all invalid URIs

packages/js/client/src/__tests__/core/sanity.spec.ts
```diff
@@ -1,9 +1,29 @@
...
+  providers,
+  stopTestEnvironment,
+} from "@polywrap/test-env-js";
+import { ensResolverPlugin } from "@polywrap/ens-resolver-plugin-js";
+import { Connection, Connections, ethereumPlugin } from "@polywrap/ethereum-plugin-js";
+import { ipfsPlugin } from "@polywrap/ipfs-plugin-js";
+import { ipfsResolverPlugin } from "@polywrap/ipfs-resolver-plugin-js";
 
-jest.setTimeout(200000);
+jest.setTimeout(20000000);
```
krisbitney: It looks like you forgot to re-adjust the timeout before pushing

packages/js/client/src/__tests__/core/sanity.spec.ts
```diff
@@ -94,30 +114,190 @@ describe("sanity", () => {
...
+
+    process.env = {
+      ...process.env,
+      IPFS_GATEWAY_URI: providers.ipfs,
+      ENS_REG_ADDR: ensAddresses.ensAddress,
+      ENS_REGISTRAR_ADDR: ensAddresses.registrarAddress,
+      ENS_RESOLVER_ADDR: ensAddresses.resolverAddress,
+    };
+
+    await initTestEnvironment();
```
krisbitney: To speed up the test, I think you could avoid using the test environment by deleting the build folders at the start of the test and then building the wrappers intermittently instead of adding resolvers. So you would only need to use the fs resolver and it would be added at the start. Instead of adding resolvers you would build the wrappers so that the build folders would exist. What do you think?

packages/js/client/src/__tests__/utils/validate/wrapper-a/package.json
```diff
@@ -0,0 +1,8 @@
+{
```
krisbitney, line 1: Could the new test wrappers be moved to `test-cases`? Or is it better if they are here?

packages/js/core/src/types/Client.ts
```diff
@@ -73,4 +78,9 @@ export interface Client
     uri: TUri,
     options: GetImplementationsOptions
   ): Result<TUri[], Error>;
+
+  validate<TUri extends Uri | string>(
+    uri: TUri,
+    options?: ValidateOptions
+  ): Promise<Result<boolean, Error>>;
```
krisbitney: Since validate can't return false, should this be `Result<true, Error>`?

packages/js/manifests/wrap/scripts/templates/compare-ts.mustache
```diff
@@ -0,0 +1,28 @@
+import {
```
krisbitney: The header for generated files is missing here

packages/js/manifests/wrap/src/formats/wrap.info/compare.ts
```diff
@@ -0,0 +1,28 @@
+import {
+  MethodDefinition, PropertyDefinition,
+} from ".";
+
+export function compareAbis(
```
krisbitney: Since compareAbis doesn't explicitly depend on the manifest version, could it be a non-generated (normal) method?

krisbitney (changes requested):
This is amazing! I think this will be an extremely useful addition to the client. Very awesome work.

I made some suggestions for possible changes, and I highlighted a couple of minor mistakes.

A feature that could be nice (likely for a future PR) would be the ability to ignore validation of some URIs. To me, it's really convenient that it is possible to use a wrapper without all of its dependencies. A dependency might only be necessary for some methods that aren't being used. So it would be nice to have a way to incorporate that sort of information into the validation check.

I want to reiterate once again how amazing this is. The options you included here are really fantastic. 🐐🐐🐐

packages/js/manifests/wrap/src/formats/wrap.info/compare.ts
cbrzn: I think it should be dependant on `Abi` version - Wdyt?

krisbitney: Sure, I just meant to say that right now it's not explicitly dependent on an `Abi` version.

It looks like the mustache template doesn't have a way to adapt to changes in the schema. The mustache template is static--it doesn't depend on any outside configuration, such as the `Abi`. So if the `Abi` structure changes, someone will have to write a new mustache template with a new method. Whether it's generated or not, it's going to break just the same.

I might be missing something, but it looks that way to me at the moment.

What are the advantages of generating this method versus not? Can we make it robust to changes in `Abi` structure, so that it updates to accommodate changes in the `Abi`?

packages/js/client/src/__tests__/core/sanity.spec.ts
cbrzn: this makes total sense, thanks :pray:

packages/js/client/src/__tests__/utils/validate/wrapper-a/package.json
cbrzn: I will leave them here. The reason is that there's no folder `client` specific there, since we want all the clients to have their own repo, I would suggest having this here

packages/js/manifests/wrap/src/formats/wrap.info/compare.ts
cbrzn: I will move this to `js/wrap/src`, I also changed the name of it to `compareSignatures`, I found it more explicit - lmk your thoughts!

krisbitney: LGTM!

krisbitney (dismissed):
This is such a cool feature!

Niraj-Kamdar (dismissed):
This is awesome!

packages/js/manifests/wrap/src/compare.ts
```diff
@@ -0,0 +1,28 @@
+import {
+  MethodDefinition, PropertyDefinition,
+} from ".";
+
+export function compareSignatures(
```
pileks: I'd refactor this to compare signatures of two `MetodDefinition`s instead of two `MethodDefinition[]`s. Just a sylistic preference for easier future reusability.

pileks (changes requested):
`compare.ts` is exporting a single function. Would it make sense to rename it to `compareSignatures.ts`, additionally `export default` the func?
Maybe also move it into a `utils` folder?

I'm thinking about this mainly from a ease-of-use, ease-of-finding point of view.

packages/js/manifests/wrap/src/compare.ts
cbrzn: yah this make sense - updated! please feel free to review and let me know any feedback :pray:

dOrgJelli (approved):

PR #1347, opened 2022-10-19
refactor: move wrap manifest schema out of the toolchain

Niraj-Kamdar:
This PR moves Wrap manifest out of the toolchain to its own repo `polywrap/wrap`. 
`manifest-js` package now directly references the manifests hosted on github with HTTP URL. 
New version committed on wrap repo would automatically get generated in manifest-js package when we run generate command


dOrgJelli (dismissed):
Looks great!

PR #1359, opened 2022-10-21
feat: JS Wrapper Around The Polywrap CLI

dOrgJelli:
This PR creates a JS SDK that wraps the CLI, allowing developers to more easily script against the `polywrap` CLI. This has been previously discussed here: https://github.com/polywrap/toolchain/pull/1343

The `@polywrap/cli-js` package spawns the `polywrap` CLI as a child-process. This is done for the following reasons:
- Makes it easy to run multiple tasks in parallel.
- Forms a clear dividing line between parent & child process, so you can better control things like system-level restrictions.
- The CLI already assumes it is running in a separate process, 1 command = 1 process life-time, so changing this assumption could cause some code to not run well (maintain bad state between runs, etc).
- Allows us to recreate the CLI in another language that isn't JS.

Moreover, this JS SDK aims to be completely type-safe, and tied to the typings used within the CLI command's themselves. We've written it in a way that the types flow easily from the command's source files, into the JS SDK, see here:
https://github.com/polywrap/toolchain/blob/ece1d33f7af4b9f22b7405fc15ac4d3435667a2c/packages/cli/src/commands/index.ts#L29-L66

And here:
https://github.com/polywrap/toolchain/blob/ece1d33f7af4b9f22b7405fc15ac4d3435667a2c/packages/js/cli/src/commands/index.ts#L30-L54

This gives us an easy to use (typed) interface like so:
```typescript
await Commands.build({
  manifestFile: "...",
  outputDir: "...",
  codegen: true
}, {
  cwd: "/path/to/project",
  env: {
    CUSTOM_ENV_VAR: "..."
  }
});
```


packages/cli/src/commands/build.ts
```diff
@@ -66,25 +64,27 @@ export const build: Command = {
         `-c, --client-config <${intlMsg.commands_common_options_configPath()}>`,
         `${intlMsg.commands_common_options_config()}`
       )
-      .option(`-n, --no-codegen`, `${intlMsg.commands_build_options_n()}`)
+      .option(`-n, --skip-codegen`, `${intlMsg.commands_build_options_n()}`)
```
pileks: `-n` seems a bit unnatural here now…

Usually, when the 1st character is already taken, the next available character is taken.

For options that are not commonly used, I have often seen the practice of even completely omitting the one-character option, and leaving only the long version.

Maybe you might want to consider one of these?

So either `-k` or no single-char option at all here?

This is just my opinion though, feel free to ignore.

packages/cli/src/commands/index.ts
```diff
@@ -2,7 +2,71 @@ export * from "./build";
...
+  InfraActions
+} from "./infra";
+import {
+  ManifestSchemaCommandOptions,
+  ManifestMigrateCommandOptions,
+  ManifestType
+} from "./manifest";
+import { RunCommandOptions } from "./run";
+
+export interface CommandTypings {
```
pileks, line 29: I’d place this into its own `CommandTypings.ts` file and use `index.ts` as a barrel, instead of defining it in `index.ts`.

pileks (dismissed):
All in all, looks amazing! I’d want to see my 2nd comment be addressed, but it’s good either way.

cbrzn:
Hey @dOrgJelli would it be possible to fix LGTM alerts before merging this?

packages/cli/src/commands/index.ts
dOrgJelli: So the idea with having it here is that it will "force" developers to keep it updated as new commands are added. If we move this into a separate file I'm worried it will be "out of sight and out of mind" and it won't be updated.

If it doesn't get updated, then the typing-magic that ensures the cli-js package is correct will not work, and the cli-js package will fall out of sync w/ the CLI's actual interface.

packages/cli/src/commands/build.ts
dOrgJelli: Resolved, reverted back to `--no-codegen`

krisbitney (dismissed):
LGTM

pileks (dismissed):
Overall a LGTM from me!

packages/cli/src/__tests__/e2e/build.spec.ts
```diff
@@ -108,7 +108,7 @@ describe("e2e tests for build command", () => {
     }
   };
 
-  it("Should show help text", async () => {
+  it.only("Should show help text", async () => {
```
cbrzn: I guess this can be removed?

packages/cli/src/__tests__/e2e/build.spec.ts
```diff
@@ -229,7 +229,7 @@ describe("e2e tests for build command", () => {
     });
   })
 
-  describe("Local strategy", () => {
+  describe.only("Local strategy", () => {
```
cbrzn: Also this

packages/cli/src/__tests__/e2e/build.spec.ts
krisbitney: Done, thanks!

packages/cli/src/__tests__/e2e/build.spec.ts
krisbitney: Removed, thanks!

Niraj-Kamdar (commented):
I agree about exposing the functionality of CLI with JS API, but from what I believe most of the tool that provide the API lib in native language exposes the underlying API methods that CLI would run, they doesn't generally wrap the CLI with exec because user can easily do this if they want to by themselves.

dOrgJelli:
Hey @Niraj-Kamdar, the reasons for having the JS SDK spawning the CLI as a child process are as follows:
- Makes it easy to run multiple tasks in parallel.
- Forms a clear dividing line between parent & child process, so you can better control things like system-level restrictions.
- The CLI already assumes it is running in a separate process, 1 command = 1 process life-time, so changing this assumption could cause some code to not run well (maintain bad state between runs, etc).
- Allows us to recreate the CLI in another language that isn't JS.

Also the point of this package is to allow users to easily script against the CLI in JS, while maintaining type safety with the CLI's commands & their arguments + options. We've written it in a way that the types flow easily from the command's source files, into the JS SDK, see here:
https://github.com/polywrap/toolchain/blob/ece1d33f7af4b9f22b7405fc15ac4d3435667a2c/packages/cli/src/commands/index.ts#L29-L66

And here:
https://github.com/polywrap/toolchain/blob/ece1d33f7af4b9f22b7405fc15ac4d3435667a2c/packages/js/cli/src/commands/index.ts#L30-L54

This gives us an easy to use (typed) interface like so:
```typescript
await Commands.build({
  manifestFile: "...",
  outputDir: "...",
  codegen: true
}, {
  cwd: "/path/to/project",
  env: {
    CUSTOM_ENV_VAR: "..."
  }
});
```

Given all of this, would you agree this package is useful and implemented well?

packages/cli/src/__tests__/e2e/infra.spec.ts
```diff
@@ -430,18 +430,17 @@ describe("e2e tests for infra command", () => {
...
-          "--modules=ganache,dev-server"
+          "--modules=ganache dev-server"
         ],
         getTestCaseDir(1),
       );
 
       await waitForPorts([
-        { port: 8546, expected: true },
-        { port: 8545, expected: true }
+        { port: 8546, expected: true }
```
krisbitney: Is this correct? The test currently in `origin-dev` awaits both ports:

```ts
      await waitForPorts([
        { port: 8546, expected: true },
        { port: 8545, expected: true }
      ]);
```

krisbitney (changes requested):

pileks (dismissed):
Other than the comments already listed out, looks good!

The explanation you gave to Niraj helps a lot too.

I'd advise that you reuse that comment to edit the PR description so that we have an immediate reference for this PR (without scrolling through tons of comments).

packages/cli/src/__tests__/e2e/infra.spec.ts
dOrgJelli: Thanks for picking up on this, it was actually an error in how we were parsing arrays, I've fixed it and reverted this test-case modification.

pileks (dismissed):

dOrgJelli:
@pileks description updated! Thanks for the recommendation.

pileks (approved):

Niraj-Kamdar (approved):
LGTM!

PR #1356, opened 2022-10-20
CLI - building interfaces uses a hidden, "empty" build strategy

pileks:
Closes #1355 

Targetting `0.9` as it seems like a non-breaking change.

The idea behind this fix is that building Interface projects doesn't require a `BuildStrategy` at all.
The `-s` argument is essentially ignored for Interface projects, and a hidden `EmptyBuildStrategy` is used, which does nothing.
Additionally, an `info` message is rendered to the user.

Users can't explicitly use this strategy.

PR #1362, opened 2022-10-24
[0.10] - CLI - building interfaces uses a hidden, "empty" build strategy

pileks:
`0.10` rebase of #1356


dOrgJelli (changes requested):
Based on my comments here, I think we should close this and implement a more future-proof strategy which addresses some of our architectural issues: https://github.com/polywrap/toolchain/pull/1356#pullrequestreview-1153745673

PR #1367, opened 2022-10-26
Config builder and CLI config refactor

### LINKED ISSUE #1378: Config builder and CLI config refactor

nerfZael:
Closes #1378


packages/js/client-config-builder/src/BaseClientConfigBuilder.ts
```diff
@@ -0,0 +1,348 @@
...
+  }
+
+  removeEnv(uri: Uri | string): ClientConfigBuilder {
+    const envUri = Uri.from(uri);
+
+    const idx = this.config.envs.findIndex((x) => Uri.equals(x.uri, envUri));
+
+    if (idx > -1) {
+      this.config.envs.splice(idx, 1);
+    }
```
cbrzn, line 178: nit: wouldn't it be easier to just use `filter` here?

cbrzn: i think this comment can be applied to all the `remove*` that uses this logic

packages/cli/src/lib/option-parsers/client-config.ts
```diff
@@ -1,19 +1,15 @@
...
+import { ClientConfigBuilder } from "@polywrap/client-config-builder-js";
 import path from "path";
+import { IClientConfigBuilder } from "@polywrap/client-config-builder-js";
 
 export async function parseClientConfigOption(
   clientConfig: string | undefined
-): Promise<Partial<ClientConfig<Uri>>> {
-  const builder = new ClientConfigBuilder().addDefaults();
+): Promise<IClientConfigBuilder> {
+  let builder = new ClientConfigBuilder().addDefaults();
```
cbrzn: I am probably missing something, but we are adding defaults to the builder, but in every command, we are explicitly adding `noDefaults` to `true`. 

So my question is: Is there any reason for adding defaults?

packages/test-cases/cases/cli/wasm/build-cmd/assemblyscript/001-sanity/yarn.lock
```diff
@@ -0,0 +1,70 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
cbrzn: Remove this

packages/js/client-config-builder/src/BaseClientConfigBuilder.ts
nerfZael: Haven't written this code, but I suspect it's a bit faster than using filter since the arrays are always unique.
I'm fine with either solution
cc @pileks

packages/cli/src/lib/option-parsers/client-config.ts
nerfZael: The idea is that the user (in the configure method of the custom config file) can either use the defaults, extend them or override them completely by creating a new instance of IClientConfigBuilder. If we used `noDefaults: false` then the user couldn't override the default resolver configuration and if we didn't add defaults before the configure method then the user couldn't inspect them and potentially change the flow of configuration based on them.
Also, by returning IClientConfigBuilder instead of something like a union of ClientConfig and CoreClientConfig it leaves the ability for the configuration to be extended further like it's done in the `JobRunner` with its custom config option

packages/js/client-config-builder/src/BaseClientConfigBuilder.ts
Niraj-Kamdar: In that case, having a `Set` would be the fastest. I belive the loading will also be easier

packages/js/client-config-builder/src/BaseClientConfigBuilder.ts
```diff
@@ -0,0 +1,348 @@
...
+          this.config.interfaces.indexOf(existingInterface),
+          1
+        );
+      }
+    }
+
+    return this;
+  }
+
+  addRedirect(from: Uri | string, to: Uri | string): ClientConfigBuilder {
```
Niraj-Kamdar, line 289: What do you think about having redirects as a `map` instead of a `list` of `from-to`?

packages/js/client-config-builder/src/BaseClientConfigBuilder.ts
```diff
@@ -0,0 +1,348 @@
...
+
+    return this;
+  }
+
+  addInterfaceImplementations(
+    interfaceUri: Uri | string,
+    implementationUris: Array<Uri | string>
+  ): ClientConfigBuilder {
+    const interfaceUriSanitized = Uri.from(interfaceUri);
+    const implementationUrisSanitized = implementationUris.map(Uri.from);
```
Niraj-Kamdar, line 234: This can also represented as map to simplify things

packages/js/client-config-builder/src/BaseClientConfigBuilder.ts
nerfZael: We talked about making envs a map, but yea, redirects could also make sense. I think the same applies to wrappers and packages.
I'll look into the best way to do that in a follow-up PR

dOrgJelli (approved):

cbrzn (approved):

pileks (approved):

Issue #1371, opened 2022-10-29
"Unsupported type" error on `codegen` when custom type in Map


sraver:
**Describe the bug**
"Unsupported type" error raises on the type generation when using custom types on a Map on the `schema.graphql`.
It also happens on the generation of types when importing an interface with custom types on a Map.

**Please provide the steps to reproduce and if possible a minimal demo of the problem.**
1. Define an interface like:
```
type Module {
    getMetadata: Metadata!
}

type Metadata {
    tokens: Map! @annotate(type: "Map<String!, [Token!]!>!")
}

type Token {
    address: String!
}
```
2. Build and deploy the interface
3. Import the interface into a module as usual, and implement `type Module implements Interface_Module {}`
4. Generate types `yarn codegen`

**Expected behavior**
Process should complete as usual and the types should be created inside `src/wrap`

**Screenshots / logs**
```
...
✅ Manifest loaded from ./src/__tests__/types/polywrap.app.yaml
🔄 Generate types
❌ Failed to generate types: Error: Unsupported type Token
Error: Unsupported type Token
```

**Relevant environment details:**
- Debian 11
- Polywrap 0.9.3

**Additional context**
The error doesn't happen when instead of a custom type, a primitive one is used, like String.

Niraj-Kamdar:
Some updates on the issue. In our schema-compose package, we aren't adding the namespace to the map types which causes the codegen to fail.

PR #1372, opened 2022-10-30
feat(schema-compose): importing type with map from external imports

Niraj-Kamdar:
- Closes: #1371

### DIFF
--- a/packages/schema/compose/src/resolve.ts
+++ b/packages/schema/compose/src/resolve.ts
@@ -29,6 +29,10 @@ import {
   EnumRef,
   EnvDefinition,
   ImportedEnvDefinition,
+  MapDefinition,
+  ArrayDefinition,
+  PropertyDefinition,
+  MethodDefinition,
 } from "@polywrap/wrap-manifest-types-js";
 import {
   parseSchema,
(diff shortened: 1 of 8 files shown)


packages/schema/parse/src/transform/finalizePropertyDef.ts
```diff
@@ -154,6 +154,10 @@ function resolveObjectOrEnumKind(
 
   const unresolved = property.unresolvedObjectOrEnum;
 
+  if (unresolved.type === "ExternalType") {
```
dOrgJelli: This debugging code should be removed.

packages/schema/compose/src/resolve.ts
```diff
@@ -787,6 +882,7 @@ async function resolveExternalImports(
             properties: type.properties,
           };
         } else if (impObjIdx > -1) {
+          console.log("IMPORTED OBJECT");
```
dOrgJelli: Remove debug logging.

dOrgJelli (changes requested):
Just some minor cleanup needed, looks great though!

dOrgJelli (approved):

PR #1379, opened 2022-11-01
[0.10] - Add logging to migration actions

pileks:
This PR introduces the `@polywrap/logging-js` package which currently contains the types/interfaces required for packages other than the CLI to interact with logging.

This introduces one caveat - `intlMsg` currently only works on the CLI level.

I'd like to open some discussion as to how we'd want to handle intl in the case where we really do go forward with logging being used in multiple packages.

### DIFF
--- a/packages/js/manifests/polywrap/src/formats/polywrap.test/deserialize.ts
+++ b/packages/js/manifests/polywrap/src/formats/polywrap.test/deserialize.ts
@@ -46,8 +46,11 @@ export function deserializePolywrapWorkflow(
   );
 
   if (versionCompare === -1) {
+    // Warn user to migrate their manifest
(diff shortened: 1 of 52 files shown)


packages/js/manifests/polywrap/src/formats/polywrap.test/deserialize.ts
```diff
@@ -46,8 +46,11 @@ export function deserializePolywrapWorkflow(
   );
 
   if (versionCompare === -1) {
+    // Warn user to migrate their manifest
+    options?.logger?.warn(`PolywrapWorkflow is using an older version of the manifest format (${anyPolywrapWorkflow.format}). Please update your manifest to the latest version (${latestPolywrapWorkflowFormat}) by using the "polywrap manifest migrate <type>" command.`);
```
dOrgJelli, line 50: This looks fine to me, but just want to call out that we're making an explicit reference here to the Polywrap CLI, which is an external dependency to this package.

dOrgJelli (approved):

PR #1382, opened 2022-11-02
CLI - `asc` should not be assumed globally available when running `local` build strategy

pileks:
While running tests, I noticed that the `local` build strategy was failing to build, giving me an output of `asc not found`.

Looking into `local.sh` I found out that the default build script assumes that `asc` is present as a global command (e.g. `npm i -g typescript`), whereas my setup did not have this.

I have changed the script to assume that TS is present locally (`yarn` should take care of that) and switched to running it using `npx asc...`

Let me know if this is something we don't want to have.


dOrgJelli (approved):

dOrgJelli:
Using `npx` makes sense to me, since it will resolve local & global installs (in that order).

PR #1384, opened 2022-11-02
CLI Build - clean up `yarn.lock` files in test cases after testing local build strategy

pileks:
As the title says, I've noticed that running CLI tests (specifically the ones for the `local` build strategy) leaves behind `yarn.lock` files inside the test case folders it operates within (currently only 001-sanity).

This should ensure proper cleanup of these stray files.

Additionally, it seems that commit `28bd34a25d50824a070aaf33696a01ec64eb7cee` changed the `ethereum` plugin's schema, which resulted in a change in the codegen output for some of the `test-cases` wrappers.
I have copy-pasted the updated expected files.


packages/cli/src/__tests__/e2e/build.spec.ts
```diff
@@ -229,6 +229,25 @@ describe("e2e tests for build command", () => {
   })
 
   describe("Local strategy", () => {
+
+    // Local strategy runs `yarn` by default, so we need to ensure that we clean up lockfiles
```
nerfZael, line 233: I'm not sure local strategy should run yarn by default

nerfZael (approved):
We really shouldn't be using the Ethereum plugin for tests but that's a topic for another issue/PR

packages/cli/src/__tests__/e2e/build.spec.ts
```diff
@@ -229,6 +229,25 @@ describe("e2e tests for build command", () => {
...
+      for (let i = 0; i < testCases.length; i++) {
+        const yarnLockfile = path.join(getTestCaseDir(0), "yarn.lock");
+        if(fs.existsSync(yarnLockfile)){
+          await fs.promises.unlink(yarnLockfile);
+        }
+      }
+    };
+
+    beforeAll(async () => {
+      await cleanupYarnLockfiles();
```
nerfZael, line 244: I don't think we need this

pileks: While I'd agree in general, it doesn't hurt to have it for situations where execution might be user-interrupted. While not a regular use case, I'd vote to keep it just to ensure that we have the same state before and after the test.

Any particular reason as to why you'd remove it?

packages/cli/src/__tests__/e2e/build.spec.ts
pileks: It's pretty normal for build processes to run the package install command once before building a project (angular does this by default, so do .net projects).

If you feel differently I'd certainly want to open an issue in which we could discuss this.

Niraj-Kamdar: I don't see any cons TBH even if the user has already installed all the deps because it won't install things if it already exists. However, I don't have a specific preference because installing deps yourself is common before building.

Niraj-Kamdar (approved):

packages/cli/src/__tests__/e2e/build.spec.ts
nerfZael: Nope, you're right

packages/cli/src/__tests__/e2e/build.spec.ts
nerfZael: Depends on the build process. Yarn doesn't do it
I think we could discuss this, but as part of another issue

dOrgJelli (approved):

Issue #1363, opened 2022-10-25
Split `PolywrapProject` into `WasmProject` and `InterfaceProject`


pileks:
From @dOrgJelli’s comments on PR #1356:

* There are 2 major architectural changes I think are needed to better support interfaces:
  1. We should split `PolywrapProject` into `WasmProject` & `InterfaceProject`.
  2. We should decouple the concept of generating the `wrap.info` from the `Compiler`, and use it for the `build` and `codegen` commands (see the comment in `codegen.ts` that says `// HACK: ...`.

The logic for build should be something like this:
```typescript
const project = createProject(manifest);
const supportedProject =
  project.isWasm() ||
  project.isPlugin() ||
  project.isInterface();

if (!supportedProject) {
  throw Error("Unsupported project type...");
}

// 1. Parse schema & build wrap.info
const schemaComposer = new SchemaComposer(...);
const infoCompiler = new WrapInfoCompiler(...);
infoCompiler.compile();

// 2, Compile wrap.wasm
if (project.isWasm()) {
  const wasmCompiler = new WrapWasmCompiler(...);
  wasmCompiler.compile();
}
```

And we can refine the semantics around the `polywrap build` and `polywrap codegen` commands. Basically, we shouldn't run codegen automatically when you run `polywrap build`, and we shouldn't generated the `wrap.info` file when you run `polywrap codegen`.

- Plugin devs will need to run `polywrap codegen` for types & `polywrap build` for `wrap.info` the file.
- Wasm devs will need to run `polywrap codegen` for types & `polywrap build` for `wrap.info` & `wrap.wasm` files.
- Interface devs will need to run `polywrap build` for `wrap.info` file.
- App devs will need to run `polywrap codegen` for types.

^^^ I guess this is the 3rd major architectural change :P

_Originally posted by @dOrgJelli in https://github.com/polywrap/toolchain/issues/1356#issuecomment-1289566074_

Related sub-tasks:

- [x] #1365 For the `polywrap build` and `polywrap codegen` semantics
- [ ] more to come…

pileks:
Currently blocked and waiting on resolution of #1365

pileks:
The remaining work here is to create an actual split between WasmProject and InterfaceProject inside the codebase. We can safely proceed with `0.10` without this change, as it's an internal refactor and provides no QOL to the end-users as it is right now.

PR #1385, opened 2022-11-03
CLI - separate build codegen actions

### LINKED ISSUE #1365: CLI - Separate `build` and `codegen` command actions
Related to #1363 

When running the `polywrap codegen` command within a Plugin project, an additional build step is executed.

This logic/step should be moved into the `build` command to further improve semantics between `codegen` and `build`.

pileks:
Closes #1365 and allows for further work on #1363 and others (namely Plugin Build Strategies, TBD).

~~**CLI test cases will be fixed once #1384 is merged!**~~ Merged.

This PR entirely separates the `build` and `codegen` command functionalities.

Running `polywrap codegen` on a `plugin` project will no longer emit `wrap.info` into the `./build` folder.

**This PR also introduces a possible circular dependency issue, as some of our plugin projects will stop working once the newest CLI version is released.**
A possible workaround for this is to have the `build` script of each project run `polywrap codegen` and continue with the build regardless of the result of the codegen, then removing it once we have the next release up.
Thoughts?

Remaining work/merge prerequisites:

- [x] Merge #1384 
- [x] See what to do in regards to the circular dependency issue described above.
  - I have decided to solve this by adding a `build:patch` script to all plugin packages which attempts to run `polywrap build` in order to output the WRAP ABI. If it fails, a message is `echo`ed and the script exits with code 0, making the build work for the current version `polywrap` in `dependencies` and for the new, unreleased version. This way we won't have an interruption in the build process, and can later on write out the proper codegen/build logic (which is first `codegen` then `build` AND make sure that both exit succesfully).


Niraj-Kamdar (dismissed):
LGTM!

packages/cli/src/commands/build.ts
```diff
@@ -164,51 +186,123 @@ async function run(options: BuildCommandOptions) {
...
     const result = await execute();
 
     if (!result) {
       process.exit(1);
     }
+
+    process.exit(0);
   } else {
     // Execute
     await execute();
```
krisbitney, line 262: Is `execute` called twice in this branch? On line 308 and in the watcher?

packages/cli/src/commands/build.ts
```diff
@@ -164,51 +186,123 @@ async function run(options: BuildCommandOptions) {
...
-    polywrapManifestPath: manifestFile,
-    logger,
-  });
-  await project.validate();
+  const project = await getProjectFromManifest(manifestFile, logger);
 
-  const polywrapManifest = await project.getManifest();
-  await validateManifestModules(polywrapManifest);
+  if (!project) {
+    return;
```
krisbitney, line 188: Should we return an exit code here?

krisbitney (changes requested):
LGTM! 

I left a couple of comments with questions. Let me know what you think. Maybe no changes are needed.

packages/cli/src/commands/build.ts
pileks: ...and print an error, yes!

packages/cli/src/commands/build.ts
pileks: Yes it is!

We execute once first, then set up the watcher, as the watcher only triggers on changes in watched files.

krisbitney: Okay, right, understood!

krisbitney (dismissed):

krisbitney:
Can this be merged @pileks @dOrgJelli @Niraj-Kamdar ?

pileks:
@krisbitney From a standpoint of implementation - Yes, it's just a matter of being reviewed! 😄

packages/cli/lang/en.json
```diff
@@ -17,11 +18,13 @@
...
   "commands_build_options_options": "options",
   "commands_build_options_t": "Use the development server's ENS instance",
-  "commands_build_options_n": "Skip code generation",
+  "commands_build_options_codegen": "Perform code generation before build",
+  "commands_build_options_codegen_dir": "Codegen output directory (default: {default})",
   "commands_build_options_w": "Automatically rebuild when changes are made (default: false)",
   "commands_build_options_s": "Strategy to use for building the wrapper",
   "commands_build_options_s_strategy": "strategy",
   "commands_build_options_l": "Log file to save console output to",
+  "commands_build_error_codegen_failed": "Code generation failed!",
```
dOrgJelli: NIT: I don't think we need to add exclamations (`!`) to output messages, I don't believe we do this elsewhere.

krisbitney (dismissed):

packages/cli/src/commands/build.ts
```diff
@@ -164,51 +186,123 @@ async function run(options: BuildCommandOptions) {
...
-  };
+      const result = await compiler.compile();
+
+      if (!result) {
+        return result;
+      }
+
+      return true;
+    };
+  } else if (isPluginManifestLanguage(language)) {
```
dOrgJelli: I think that we should be handling this inside of the "Compiler", like we already do for interface projects. I'll make a push now to add this.

dOrgJelli (changes requested):

packages/cli/lang/en.json
dOrgJelli: Made these changes here: https://github.com/polywrap/toolchain/pull/1467

packages/cli/src/commands/build.ts
dOrgJelli: Made these changes here: https://github.com/polywrap/toolchain/pull/1467

dOrgJelli (changes requested):
Also @pileks we should update the CLI's README.

packages/cli/lang/en.json
pileks: Thanks for the PR, merged.

pileks:
@dOrgJelli I've performed the minimal required changes to the readme just now.

We can refactor the readme to use `doc-snippets` in a later PR - for now this should be it.

LMK if this works.

dOrgJelli (approved):

PR #1427, opened 2022-11-17
Split client-js into core-client-js and client-js

nerfZael:


packages/js/core-client/jest.rs.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
cbrzn: I think this file can be removed, since the tests are in `client`

cbrzn (commented):
cool! do you think it would make sense to move some tests to `core-client` package too? to test at least the main functionalities (like get env or get implementations)?

packages/js/core-client/jest.rs.config.js
nerfZael: I left this so we can add tests later on

nerfZael:
> cool! do you think it would make sense to move some tests to `core-client` package too? to test at least the main functionalities (like get env or get implementations)?

Yea, I think a lot of tests could probably be moved
Lets do that in another PR though

PR #1430, opened 2022-11-20
feat: add polywrap.yaml project resources

dOrgJelli:


packages/test-cases/cases/cli/wasm/build-cmd/assemblyscript/015-resource-files/polywrap.yaml
```diff
@@ -0,0 +1,11 @@
+format: 0.3.0
+project:
+  name: test-project
+  type: wasm/assemblyscript
+source:
+  module: ./src/index.ts
+  schema: ./src/schema.graphql
+resources:
```
nerfZael: I'd propose having a single directory for resources:
```
resources: ./resources
```
And resources dir would look like e.g.:
- README.md
- some-dir
  - some-file.txt
And then build artifacts would be:
- wrap.info
- wrap.wasm
- README.md
- some-dir
  - some-file.txt
Instead of having list of files and directories.
This allows the developer to add files to the build artifacts despite them having the same name as the files in the project (e.g. having separate project and build README's that are both named README.md)

dOrgJelli: Okay I agree with this approach, nice 👍

packages/cli/src/lib/Compiler.ts
```diff
@@ -48,12 +48,15 @@ export class Compiler {
...
 
-        // Compile the Wrapper
+        // Compile & Output: wrap.wasm
         await this._buildModules();
       }
 
-      // Output Polywrap Metadata
-      await this._outputPolywrapMetadata();
+      // Copy: Resource files
+      await this._copyResourceFiles();
```
pileks: Is it necessarily the Compiler's responsibility to copy resources and metadata?
IMO, strictly from a definition (what is a compiler?) standpoint, it would make more sense to move this into the build process itself (CLI, build strategy, something third).

Food for thought, and a possible refactor for later.

pileks (dismissed):
Looks great in general, my only comment is more of a "big picture" issue.

packages/test-cases/cases/cli/wasm/build-cmd/assemblyscript/015-resource-files/polywrap.yaml
dOrgJelli: Okay this has been implemented, super nice simplification!

packages/cli/src/lib/Compiler.ts
dOrgJelli: Yep was thinking about this as well. Currently we have the metadata copy logic within the compiler, so that's why I added this there as well.

I'm planning on making a follow up PR that removes the `polywrap.meta.yaml` manifest entirely, so that will be removed from the compiler.

Here's how I see it:
- Build process has multiple steps.
- The "primary" step is compiling the wrap.wasm module.
- Compiling the wrap.wasm module can be done using different "strategies" (local, vm)
- I would propose we organize our semantics like so:
```
build-steps/
    codegen/
    build-wrap-info/
    build-wrap-wasm/
        strategies/
            vm/
            local/
    validate-wrap-wasm/
    copy-resources/
```

These build steps can be functional units, that are configured by input parameters. This way you could execute an array of pre-configured build-steps in sequence, and they could become wrappers more easily (and user-configurable if desired).

Just a quick brainstorm on how we can flatten the architecture, make it more functional in nature, and allow for future extension.

PR #1431, opened 2022-11-21
Wrap error structure

### LINKED ISSUE #701: Polywrap Specific Error Structure

krisbitney:
This PR adds a custom `WrapError` error structure that is capable of parsing errors from strings, allowing us to improve error formatting and provide additional information for debugging.

Example error output for URI Resolution Exception:
```bash
    WrapError: SubInvocation exception encountered
    code: 50 WRAPPER INVOKE ABORTED
    uri: wrap://fs//Users/kris/WebstormProjects/monorepo/packages/test-cases/cases/wrappers/wasm-as/subinvoke-error/invoke/build
    method: subWrapperNotFound
    args: {
      "a": 1,
      "b": 1
    } 
    source: { file: "~lib/@polywrap/wasm-as/containers/Result.ts", row: 171, col: 13 }
    
    Another exception was encountered during execution:
    WrapError: Unable to find URI wrap://ens/not-found.eth.
    code: 27 URI NOT FOUND
    uri: wrap://ens/not-found.eth
    uriResolutionStack: [
      "wrap://ens/not-found.eth => PackageToWrapperCacheResolver",
      [
        "wrap://ens/not-found.eth => UriResolverAggregator",
        [
          "wrap://ens/not-found.eth => StaticResolver - Miss",
          "wrap://ens/not-found.eth => ExtendableUriResolver",
          [
            "wrap://ens/ipfs-resolver.polywrap.eth => PackageToWrapperCacheResolver (Cache) => wrapper (wrap://ens/ipfs-resolver.polywrap.eth)",
            "wrap://ens/not-found.eth => ResolverExtension (wrap://ens/ipfs-resolver.polywrap.eth)",
            "wrap://ens/ens-resolver.polywrap.eth => PackageToWrapperCacheResolver (Cache) => wrapper (wrap://ens/ens-resolver.polywrap.eth)",
            "wrap://ens/not-found.eth => ResolverExtension (wrap://ens/ens-resolver.polywrap.eth)",
            "wrap://ens/fs-resolver.polywrap.eth => PackageToWrapperCacheResolver (Cache) => wrapper (wrap://ens/fs-resolver.polywrap.eth)",
            "wrap://ens/not-found.eth => ResolverExtension (wrap://ens/fs-resolver.polywrap.eth)",
            "wrap://ens/http-resolver.polywrap.eth => PackageToWrapperCacheResolver (Cache) => wrapper (wrap://ens/http-resolver.polywrap.eth)",
            "wrap://ens/not-found.eth => ResolverExtension (wrap://ens/http-resolver.polywrap.eth)",
            "wrap://ipfs/QmfRCVA1MSAjUbrXXjya4xA9QHkbWeiKRsT7Um1cvrR7FY => PackageToWrapperCacheResolver (Cache) => wrapper (wrap://ipfs/QmfRCVA1MSAjUbrXXjya4xA9QHkbWeiKRsT7Um1cvrR7FY)",
            "wrap://ens/not-found.eth => ResolverExtension (wrap://ipfs/QmfRCVA1MSAjUbrXXjya4xA9QHkbWeiKRsT7Um1cvrR7FY)"
          ]
        ]
      ]
    ]

```

Example error output for exception thrown during Wasm invocation:
```bash
    WrapError: SubInvocation exception encountered
    code: 50 WRAPPER INVOKE ABORTED
    uri: wrap://fs//Users/kris/WebstormProjects/monorepo/packages/test-cases/cases/wrappers/wasm-as/subinvoke-error/invoke/build
    method: throwsInTwoSubinvokeLayers
    args: {
      "a": 1,
      "b": 1
    } 
    source: { file: "~lib/@polywrap/wasm-as/containers/Result.ts", row: 171, col: 13 }
    
    Another exception was encountered during execution:
    WrapError: SubInvocation exception encountered
    code: 50 WRAPPER INVOKE ABORTED
    uri: wrap://ens/bad-math.eth
    method: subInvokeWillThrow
    args: {
      "0": 130,
      "1": 161,
      "2": 97,
      "3": 1,
      "4": 161,
      "5": 98,
      "6": 1
    } 
    source: { file: "~lib/@polywrap/wasm-as/containers/Result.ts", row: 171, col: 13 }
    
    Another exception was encountered during execution:
    WrapError: __wrap_abort: I threw an error!
    code: 50 WRAPPER INVOKE ABORTED
    uri: wrap://ens/bad-util.eth
    method: iThrow
    args: {
      "0": 129,
      "1": 161,
      "2": 97,
      "3": 0
    } 
    source: { file: "src/index.ts", row: 5, col: 5 }
```

Example error output for exception thrown during plugin wrapper invocation:
```bash
    WrapError: I'm throwing!
    code: 50 WRAPPER INVOKE ABORTED
    uri: wrap://plugin/mock
    method: methodThatThrows
    args: {} 
    source: { file: "/Users/kris/WebstormProjects/monorepo/packages/js/client/src/__tests__/helpers/mockPluginRegistration.ts", row: 13, col: 17 }
```

Closes https://github.com/polywrap/toolchain/issues/701


packages/js/client/package.json
```diff
@@ -17,7 +17,8 @@
     "test": "jest --passWithNoTests --runInBand --verbose=true --detectOpenHandles --forceExit",
     "test:ci": "jest --passWithNoTests --runInBand --verbose --detectOpenHandles --forceExit",
     "test:rust": "jest --passWithNoTests --runInBand --verbose --detectOpenHandles --forceExit --config ./jest.rs.config.js",
-    "test:watch": "jest --watch --passWithNoTests --verbose --detectOpenHandles"
+    "test:watch": "jest --watch --passWithNoTests --verbose --detectOpenHandles",
+    "test:errors": "yarn test src/__tests__/core/error-structure.spec.ts"
```
dOrgJelli: Can this be removed? I think it was meant for testing.

packages/js/core/src/types/WrapError.ts
```diff
@@ -0,0 +1,343 @@
...
+ */
+export enum WrapErrorCode {
+  UNKNOWN,
+  CLIENT_LOAD_WRAPPER,
+  CLIENT_GET_FILE,
+  CLIENT_GET_IMPLEMENTATIONS,
+  CLIENT_VALIDATE_RESOLUTION,
+  CLIENT_VALIDATE_ABI,
+  CLIENT_VALIDATE_RECURSIVE,
+  CLIENT_QUERY_MALFORMED,
```
dOrgJelli: Can we remove these `CLIENT_QUERY_...` error codes, since I think we should remove the GraphQL dependency (and the `.query(...)` method) from the client entirely in the near future?

packages/js/core/src/types/WrapError.ts
```diff
@@ -0,0 +1,343 @@
...
+      | "reason"
+      | "uri"
+      | "method"
+      | "args"
+      | "file"
+      | "row"
+      | "col"
+      | "resolutionStack"
+      | "cause"
+    > = WrapError.re.exec(error);
```
dOrgJelli, line 157: This is really great :D

packages/js/wasm/src/imports.ts
```diff
@@ -215,9 +246,21 @@ export const createImports = (config: {
...
-          `__wrap_abort: ${msg}\nFile: ${file}\nLocation: [${line},${column}]`
-        );
+        let code = WrapErrorCode.WASM_INVOKE_ABORTED;
+
+        // check if error was thrown in bindings
+        if (
+          file.startsWith("~lib/@polywrap/wasm-as/msgpack/") ||
+          file.startsWith("src/wrap/")
+        ) {
+          code = WrapErrorCode.WASM_SERIALIZATION_ERROR;
```
dOrgJelli: Hmmmm, is there maybe a more consistent way of detecting this? This feels wrong...

dOrgJelli (changes requested):

packages/js/client/package.json
krisbitney: yes, exactly, it was left in by mistake

packages/js/core/src/types/WrapError.ts
krisbitney: yeah, for sure

packages/js/wasm/src/imports.ts
krisbitney: It's definitely hacky. It would be nice to hear other ideas.

The other main ways I thought to handle this were:
1. Add a prefix to errors thrown in msgpack or bindings, such as "$wrapCode_42 [message]", that could be stripped and parsed. We could also let wrapper devs use this to set error codes when they throw.
2. I thought to create a special `wrap_abort` method to be called, but I couldn't find a way to get source info manually.
3. We could instead not use a serialization error code at all. In this case, I think it would be a good idea to add a little more information to the errors we throw in msgpack and bindings so that it's more clear to the user that it's a serialization error.

dOrgJelli: I think a better approach to this would be:
- Have the wasm bindings be "WrapError aware".
- Throw the `WASM_SERIALIZATION_ERROR` code from within the wasm bindings.
- Maybe parse & extract the code within the `__wrap_abort` function from the `msg` argument? Not sure this is needed though.

dOrgJelli: Actually, I retract my last comment. I think the `WrapError` should be a client-side only thing (for now). The errors should (IMO) only consist of things that the client can "know" have happened. Based on this, I think the `WASM_SERIALIZATION_ERROR` is inherently "flawed" since it steps over the wasm boundary, requiring the client to know what happened within the wasm module's runtime.

When you look at the full set, you can see that all of them (except the last) are client-side things, but the last one requires wasm-side context:
```
WASM_INVOKE_ABORTED
WASM_SUBINVOKE_ABORTED
WASM_INVOKE_FAIL
WASM_MODULE_NOT_FOUND
WASM_STATE_ERROR
WASM_SERIALIZATION_ERROR
```

I think we can just remove the `WASM_SERIALIZATION_ERROR` entirely, because the error message will have enough information for a user to debug further (using the serialization context stack we have already implemented).

Thoughts?

krisbitney: This sounds fantastic. Apologies for taking so long to respond!

packages/js/core/src/types/WrapError.ts
```diff
@@ -0,0 +1,334 @@
...
+Error code naming convention (approximate):
+    type of handler
+    type of functionality
+    piece of functionality
+    ==> handler_typeFn_pieceFn
+
+Error code map:
+  0-24 -> Client
+  25-49 -> URI resolution
+  50-74 -> Wasm wrapper invocation & sub-invocation
```
nerfZael: We've been moving away from having the client and especially the core differentiating between wasm wrappers and plugin wrappers.
It'd be great if we could fill the need for this in some other way than having wrapper type specific codes.

packages/js/core/src/types/WrapError.ts
```diff
@@ -0,0 +1,334 @@
...
+Wrap error codes provide additional context to WrapErrors.
+
+Error code naming convention (approximate):
+    type of handler
+    type of functionality
+    piece of functionality
+    ==> handler_typeFn_pieceFn
+
+Error code map:
+  0-24 -> Client
```
nerfZael: Was there special reasoning behind 24? Also applies to other numbers

packages/js/wasm/src/WasmWrapper.ts
```diff
@@ -151,7 +154,13 @@ export class WasmWrapper implements Wrapper {
       const args = options.args || {};
       const wasmResult = await this._getWasmModule();
       if (!wasmResult.ok) {
-        return wasmResult;
+        const error = new WrapError(wasmResult.error, {
+          code: WrapErrorCode.WASM_MODULE_NOT_FOUND,
```
nerfZael: I don't think it's always because the module was not found. It could be some other error retrieving the module.

packages/js/core/src/types/WrapError.ts
```diff
@@ -0,0 +1,334 @@
...
+    this.args = options.args;
+    this.source = options.source;
+    this.resolutionStack = options.resolutionStack;
+    this.cause = options.cause;
+    this.stack = options.stack;
+    this.prev = options.prev;
+    Object.setPrototypeOf(this, WrapError.prototype);
+  }
+
+  static parse(error: string): WrapError | undefined {
```
nerfZael, line 109: I think this belongs in wasm-js. It's only relevant for wasm wrappers

packages/js/core/src/types/WrapError.ts
```diff
@@ -0,0 +1,334 @@
...
+    } else if (code < 75) {
+      return "InvokeError";
+    } else if (code < 100) {
+      return "PluginError";
+    } else {
+      return "WrapError";
+    }
+  }
+
+  private static metaMessage(code: WrapErrorCode): string {
```
nerfZael: It'd be better to move the metaMessage to some other place other than core-js.
I don't think the core should care about verbose message strings.

packages/js/core/src/types/WrapError.ts
```diff
@@ -0,0 +1,334 @@
...
+
+Error code map:
+  0-24 -> Client
+  25-49 -> URI resolution
+  50-74 -> Wasm wrapper invocation & sub-invocation
+  75-99 -> Plugin invocation & sub-invocation
+  100-255 -> Unallocated
+ */
+export enum WrapErrorCode {
+  UNKNOWN,
```
nerfZael: I think UNKNOWN kind of defeats the purpose of error codes.

packages/js/core/src/types/WrapError.ts
```diff
@@ -0,0 +1,334 @@
...
+  WASM_SUBINVOKE_ABORTED,
+  WASM_INVOKE_FAIL,
+  WASM_MODULE_NOT_FOUND,
+  WASM_STATE_ERROR,
+  PLUGIN_INVOKE_FAIL = 75,
+  PLUGIN_METHOD_NOT_FOUND,
+  PLUGIN_ARGS_MALFORMED,
+}
+
+export type WrapErrorSource = Readonly<{
```
nerfZael: Is this a wasm implementation detail? If so, it should be in the wasm-js library

nerfZael (changes requested):

packages/js/core/src/types/WrapError.ts
krisbitney: I just pushed a commit combining them into one category

packages/js/core/src/types/WrapError.ts
krisbitney: I just allocated 25 code numbers to each section

packages/js/wasm/src/WasmWrapper.ts
krisbitney: I changed the error to `WRAPPER_READ_FAIL`

packages/js/core/src/types/WrapError.ts
krisbitney: This is a very good point. I moved it to `src/packages/js/wasm/serc/helpers/wrapErrorUtils`.

packages/js/core/src/types/WrapError.ts
krisbitney: I'm not sure where else it could live. I think it brings some value to the errors, but maybe not. If we think the messages add value, then I think it makes sense for WrapError to know how to stringify itself without external assistance. If the messages aren't very useful, we can just dump them.

What do you think?

packages/js/core/src/types/WrapError.ts
krisbitney: Good point. I removed it.

packages/js/core/src/types/WrapError.ts
krisbitney: I moved the type declaration to wasm-js and am instead using a type literal in the WrapError class.

It's specific to Wasm wrappers, but it's also necessarily a property of WrapError since WrapError needs to know how to handle both Wasm and plugin errors. Maybe it would make more sense to have something like a core WrapError and then either subclass it or use a composition/dependency injection sort of pattern where the core WrapError holds a more specific type of error that it delegates some behavior to?

packages/js/wasm/src/imports.ts
dOrgJelli: Hey no problem, also I didn't see your comment above originally when I typed my 2nd response (didn't reload my page). I really like your idea about "Add a prefix to errors thrown in msgpack or bindings, such as "$wrapCode_42 [message]", that could be stripped and parsed. We could also let wrapper devs use this to set error codes when they throw."

I think this would be a great thing to add in the future so we can easily have new error codes "cross the boundary" and make it more part of the WRAP interoperability standard (similar to HTTP error codes). For now just starting with this as a client-side only thing is an easy first step.

packages/js/core/src/types/WrapError.ts
dOrgJelli: I'm actually pro `0 = UNKNOWN`. I do not think we should ascribe meaning to a `0` error code, and we should treat it as `undefined`. This is because often times when transmitting data, you may be reading from a buffer, and if something is not initialized it will be zero. This would lead to a false positive of whatever error we ascribe to `0` (I've seen this in the past when working on telemetry systems). So, I think we should have `UNKNOWN` as the 0th error code to ensure errors are interpreted non-zero values.

packages/js/core/src/types/WrapError.ts
dOrgJelli: I think sub-classing a core `WrapError` type makes sense. This is how it works with custom TypeScript / JavaScript errors.

packages/js/core/src/types/WrapError.ts
```diff
@@ -0,0 +1,212 @@
...
+  CLIENT_VALIDATE_ABI,
+  CLIENT_VALIDATE_RECURSIVE,
+  URI_RESOLUTION = 25,
+  URI_RESOLVER,
+  URI_NOT_FOUND,
+  WRAPPER_INVOKE_ABORTED = 50,
+  WRAPPER_SUBINVOKE_ABORTED,
+  WRAPPER_INVOKE_FAIL,
+  WRAPPER_READ_FAIL,
+  WRAPPER_STATE_ERROR,
```
dOrgJelli: I would rather call this something more universal / catch-all, that could be used in all types of wrappers. Basically what it is being used for now is to catch all internal errors of the wasm wrapper that should never happen. Effectively it's for bad internal states that shouldn't exist, but could possibly if the internal mechanisms within the wrapper fail in some way.

So, I think we should rename this to `WRAPPER_INTERNAL_ERROR`, so that it can be used to represent any unforeseen errors within the wrapper's internals. These errors shouldn't happen if the wrapper was functioning as intended, but could given extreme circumstances. This could be used for `PluginWrapper`, `WasmWrapper`, and even an `HttpWrapper` if we build that in the future. For example `HttpWrapper` could throw an internal error if there are no networking capabilities available on the device. These internal errors can be made more "colorful" by doing the `WrapError` sub-classing like you mentioned above.

Idk, just a thought, curious what you think.

packages/js/core-client/src/PolywrapCoreClient.ts
```diff
@@ -316,10 +330,16 @@ export class PolywrapCoreClient implements CoreClient {
         errors: errors.length === 0 ? undefined : errors,
       };
     } catch (error: unknown) {
+      const unknownQueryErrorToWrapError = (e: Error): WrapError =>
+        new WrapError((e as Error)?.message, {
```
dOrgJelli: Should we add a default string here if a message property is not found, instead of `undefined`? Maybe "client.query: no error message found" or something like that?

dOrgJelli: 2 points to convey:
- from within query method
- no error message found / unknown error obj caught

dOrgJelli (changes requested):

packages/js/core/src/types/WrapError.ts
dOrgJelli: > It's only relevant for wasm wrappers

Is this true though? Can't this be used to parse any stringified `WrapError`? If not, I think that's a problem. To me, we should have a `WrapError` structure, which can be easily stringified in a human readable way, and parsed back into the structure itself.

Another way of thinking about this: why are we parsing a human-readable output of the error, and not an easier format like a json string or a msgpack byte array? I think that, if we're transmitting the whole `WrapError` structure over a boundary, we should encoded it in a way that doesn't require custom regex parsing to get the `WrapError` struct back. We might need to change some wasm interfaces to do this properly (i.e. change `string` to `bytes`). This could be something we do in the future, but good to keep in mind.

This would mean that we have the following `WrapError` interface:
1. `.stringify()` or `.print()` - create human readable string.
2. `.serialize()` or `.toBuffer()` - encode `WrapError` for transmission over some boundary.
3. `.deserialize(x)` or `.fromBuffer(x)` - decode the input `x` (utf-8 string or bytes) into a `WrapError` structure.

packages/js/core/src/types/WrapError.ts
```diff
@@ -0,0 +1,212 @@
...
+    piece of functionality
+    ==> handler_typeFn_pieceFn
+
+Error code map:
+  0-24 -> Client
+  25-49 -> URI resolution
+  50-74 -> Wrapper invocation & sub-invocation
+  75-255 -> Unallocated
+ */
+export enum WrapErrorCode {
```
dOrgJelli, line 25: Some of these error names don't imply a negative situation, and could be interpreted positively. For example `CLIENT_GET_IMPLEMENTATIONS` doesn't inherently sound negative in nature. I'd recommend looking at some reference error codes (ex: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses) and seeing if there are better names we could use for these that imply the error scenario in question a bit more specifically. For example `CLIENT_VALIDATE_ABI` => `CLIENT_VALIDATE_ABI_NOT_FOUND`

dOrgJelli (changes requested):

packages/js/core/src/types/WrapError.ts
krisbitney: Right, it can parse any stringified `WrapError`. 

Right now it's used in wasm-js because Wasm wrapper error messages are written to the memory buffer as strings. The error messages may or may not be WrapErrors, depending on the situation.

Currently, errors are essentially strings that can be nested within each other. The message part of the error can be an entire other error string. This occurs when an error is thrown during a subinvocation and is either not handled or is re-thrown. When a message is only thrown in the top-level invocation, the message is not a stringified error.

```typescript
// subinvocation in assemblyscript
const result = My_Module.method(...);
if (!result.ok) {
  // the entire error is now the message part of a new error.
  // this also happens when unwrapping a bad result.
  throw new Error(result.error);
}
```

Parsing the strings makes it easy to maintain and return a clean error structure that prints a clean error message.

krisbitney: It would be cool to do this with msgpack serialization in the future, 100%!

krisbitney: I know I explained stuff you already know, but I was just clarifying that we are currently only parsing the errors in `wasm-js`. Maybe end-users will want to be able to parse errors, though, and it makes sense to me to let `WrapError` know how to parse itself. Idk

krisbitney: I moved the parsing logic back to `WrapError`

packages/js/core/src/types/WrapError.ts
krisbitney: I added `UNKNOWN` back

packages/js/core/src/types/WrapError.ts
krisbitney: Very good thoughts. I renamed `WRAPPER_STATE_ERROR` to `WRAPPER_INTERNAL_ERROR`.

packages/js/core-client/src/PolywrapCoreClient.ts
krisbitney: Nice catch. I changed it to `e?.message ?? "client.query: no error message found"`

packages/js/core/src/types/WrapError.ts
krisbitney: Good point. I changed them to the following:
```
  CLIENT_LOAD_WRAPPER_ERROR
  CLIENT_GET_FILE_ERROR,
  CLIENT_GET_IMPLEMENTATIONS_ERROR,
  CLIENT_VALIDATE_RESOLUTION_FAIL,
  CLIENT_VALIDATE_ABI_FAIL,
  CLIENT_VALIDATE_RECURSIVE_FAIL
```

packages/js/core/src/types/WrapError.ts
krisbitney: Okay, I'm looking at this now. 

One thing worth considering is that the parsing logic needs to be able to parse any `WrapError` since subinvocations can return any type of `WrapError` string. I think that means the `WrapError` will need to know about all its children, or `wasm-js` will need to know about an error subclass that's defined somewhere like `plugin-js`. Is that okay?

packages/js/core/src/types/WrapError.ts
krisbitney: I removed metaMessage altogether

packages/js/core/src/types/WrapError.ts
krisbitney: I instead added error source info for exceptions thrown in plugin wrappers, making the property more universal.

Example output:
```ts
WrapError: I'm throwing!
code: 50 WRAPPER INVOKE ABORTED
uri: wrap://plugin/mock
method: methodThatThrows
args: {} 
source: { file: "/Users/kris/WebstormProjects/monorepo/packages/js/client/src/__tests__/helpers/mockPluginRegistration.ts", row: 13, col: 17 }
```

krisbitney:
If I marked a conversation "resolved" and you don't feel it was resolved, please feel free to change it back to "unresolved".

dOrgJelli (dismissed):

dOrgJelli:
:fire: :fire: :fire: :fire: :fire: :fire:

packages/js/wasm/src/WasmWrapper.ts
```diff
@@ -172,12 +181,21 @@ export class WasmWrapper implements Wrapper {
...
 
-      const abort = (message: string) => {
-        throw Error(
-          `WasmWrapper: Wasm module aborted execution.\nURI: ${options.uri.uri}\n` +
-            `Method: ${method}\n` +
-            `Args: ${JSON.stringify(args, null, 2)}\nMessage: ${message}.\n`
-        );
+      const abort = (
+        message: string,
+        code: WrapErrorCode = WrapErrorCode.WRAPPER_INVOKE_ABORTED,
```
nerfZael: I don't think this default is ever used

packages/js/wasm/src/WasmWrapper.ts
```diff
@@ -172,12 +181,21 @@ export class WasmWrapper implements Wrapper {
...
+        source?: ErrorSource
+      ) => {
+        const prev = WrapError.parse(message);
```
nerfZael, line 188: Is there a reason to parse the message for code other than WRAPPER_INVOKE_ABORTED ?

packages/js/core/src/types/WrapError.ts
```diff
@@ -0,0 +1,274 @@
...
+Wrap error codes provide additional context to WrapErrors.
+
+Error code naming convention (approximate):
+    type of handler
+    type of functionality
+    piece of functionality
+    ==> handler_typeFn_pieceFn
+
+Error code map:
+  0-24 -> Client
```
nerfZael: Since 0 is Unknown what about:
0 = Unknown
1 - 25 - Client
26 - 50 - ...
...
76 - 255- Unallocated?

packages/js/core/src/types/WrapError.ts
nerfZael: In cases where we're reading from a buffer, and we get a zero there, I think the buffer should be treated as corrupted and not a valid WrapError. I do not think we should be allowing undefined states for an enum we control. The error codes, IMO, should start at 1.

Still, I am fine with keeping this for the time being and being more strict about the codes in the future.

packages/js/core/src/types/WrapError.ts
nerfZael: There's 2 things one can parse: serialized WrapErrors and errors from wasm modules (which may or may not be WrapErrors).
- The serialized WrapError parsing logic should be in the WrapError, or somewhere close. It is easy to parse since we know the structure and does not need regex.
- The wasm module error parsing logic should be in wasm-js. And since it can be anything regex is fine there. 

The parsing logic here is both of those things.

packages/js/core/src/types/WrapError.ts
```diff
@@ -0,0 +1,274 @@
...
+
+export interface WrapErrorOptions {
+  code: WrapErrorCode;
+  uri: string;
+  method?: string;
+  args?: string;
+  source?: ErrorSource;
+  resolutionStack?: CleanResolutionStep;
+  cause?: unknown;
+  prev?: Error;
```
nerfZael: Isn't `prev` always a WrapError?
Also, can we get a more descriptive name for it, maybe `innerError` or something else?

packages/js/core/src/types/WrapError.ts
```diff
@@ -0,0 +1,274 @@
...
+export class WrapError extends Error {
+  readonly name: string = "WrapError";
+  readonly code: WrapErrorCode;
+  readonly reason: string;
+  readonly uri: string;
+  readonly method?: string;
+  readonly args?: string;
+  readonly source?: ErrorSource;
+  readonly resolutionStack?: CleanResolutionStep;
+  readonly cause?: unknown;
```
nerfZael, line 70: Since this is essentially a native Error, and WrapError has a need to be serialized, I think `cause` should be of type `string` (Serialized Error basically).
I'm also open to any other solution, but I don't think we should have an `unknown` type in WrapError

nerfZael (changes requested):

packages/js/wasm/src/WasmWrapper.ts
krisbitney: Parsing message is central to this PR and it has little to do with error codes.

Let's review the status quo:

```ts
    const message = `Error: WasmWeb3Api: Wasm module aborted execution.
    URI: w3://ipfs/QmRZrQkmjK9TnC8fHwJCf7YRqMFiyPvcN5vcYc9ZSXQ9tB
    Module: query
    Method: queryEnv
    Input: {
      "arg": "string"
    }
    Message: __w3_abort: 129
    File: ~lib/@web3api/wasm-as/msgpack/ReadDecoder.ts
    Location: [166,5].`
    
    const abort = (message: string) => {
        throw Error(
          `WasmWrapper: Wasm module aborted execution.\nURI: ${options.uri.uri}\n` +
            `Method: ${method}\n` +
            `Args: ${JSON.stringify(args, null, 2)}\nMessage: ${message}.\n`
        );
      };
      
    abort(message);
```

Any time an exception is thrown during a sub-invocation, `message` is a stringified error. That means errors get nested within errors, leading to output that looks something like this:

```ts
    Error: WasmWeb3Api: Wasm module aborted execution.
    URI: w3://ipfs/QmRZrQkmjK9TnC8fHwJCf7YRqMFiyPvcN5vcYc9ZSXQ9tB
    Module: query
    Method: queryEnv
    Input: {
      "arg": "string"
    }
    Message: Error: WasmWeb3Api: Wasm module aborted execution.
    URI: w3://ipfs/QmRZrQkmjK9TnC8fHwJCf7YRqMFiyPvcN5vcYc9ZSXQ9tB
    Module: query
    Method: queryEnv
    Input: {
      "arg": "string"
    }
    Message: __w3_abort: 129
    File: ~lib/@web3api/wasm-as/msgpack/ReadDecoder.ts
    Location: [166,5].
    File: ~lib/@web3api/wasm-as/msgpack/ReadDecoder.ts
    Location: [166,5].
```

Now imagine that happening multiple layers deep (e.g., a sub-sub-sub-invocation)!

And recall that message can be any message or any stringified polywrap error (either a string as shown above in the status quote, or in this PR a `WrapError`).

This is why parsing the message is necessary.

At first glance, you might wonder why we don't just capture the error after the sub-invocation within `__wrap_subinvoke`. The answer is simply that by this time the error messages are already nested. `__wrap_abort` is called during an invocation, not after it.

The best way around parsing is to allow errors to cross the Wasm boundary. In that case we could be working with msgpack encoded bytes instead of strings. I think that's definitely something we should look at in the future, but I think parsing is a nice incremental improvement over the status quo.

packages/js/wasm/src/WasmWrapper.ts
krisbitney: Nice catch!

packages/js/core/src/types/WrapError.ts
krisbitney: Sure thing!

krisbitney:
@nerfZael 

Somehow I can't reply to some of your messages so I'm responding here.

> In cases where we're reading from a buffer, and we get a zero there, I think the buffer should be treated as corrupted and not a valid WrapError. I do not think we should be allowing undefined states for an enum we control. The error codes, IMO, should start at 1.
> 
> Still, I am fine with keeping this for the time being and being more strict about the codes in the future.

I removed `UNKNOWN` and am starting the enum at 1.

> There's 2 things one can parse: serialized WrapErrors and errors from wasm modules (which may or may not be WrapErrors).
> 
> The serialized WrapError parsing logic should be in the WrapError, or somewhere close. It is easy to parse since we know the structure and does not need regex.
> The wasm module error parsing logic should be in wasm-js. And since it can be anything regex is fine there.
> The parsing logic here is both of those things.

All errors from Wasm modules are made to be of type `WrapError`. 

It is possible that a message is not a stringified `WrapError`, but still we need to check whether it is or not (I explain why in your other question about `parse`). The `parse` method returns `undefined` when a message is not a `WrapError`, in which case `message` is used as the error message in a new `WrapError`.

packages/js/core/src/types/WrapError.ts
krisbitney: It is necessary to have cause be of `unknown` type to accommodate uri resolver errors, which are untyped. WrapError provides an incremental improvement over the status quo in this respect.

Consider the status quo: a uri resolver error (of type `unknown`) that is thrown during a sub-invocation is completely lost. It is ignored when the error is stringified. The user loses all of the information.

In `WrapError`, the uri resolver error is itself stringified and included in the error message. While we lose the uri resolver error itself, we at least keep a stringified version.

In both the status quo and the current PR, the full uri resolver error of `unknown` type is still attached to the error object when it occurs during a top-level invocation. If we change `cause` to be of type `string`, we lose this.

I know you thought before that uri resolver errors should be of type `unknown` to provide flexibility to custom uri resolvers. If `cause` becomes a `string`, there is little point to uri resolver errors being of any type but `string` since that's all the user will ever get.

When serializing `WrapError`, I 100% agree the value of `cause` should be a `string`.

What do you think?

packages/js/core/src/types/WrapError.ts
krisbitney: Yes, you're right!

packages/js/wasm/src/WasmWrapper.ts
krisbitney: If you were instead asking why we parse for error codes at all in the `parse` regex, it is because the error code could be any wrap error code.

An error code will be present if and only if the error was thrown during a subinvocation and then not handled by the higher level invocation (e.g. re-thrown). In that case, `message` could be a stringified version of any `WrapError` that can be returned from `client.invoke`.

nerfZael: No, I'm saying that the `abort` function is only called in `imports.ts`.
It is called in many places there where the message is just a simple string, and the only time it is not is when it is called from within `__wrap_abort` where the code is WRAPPER_INVOKE_ABORTED.

So you could rewrite it like this:
```typescript
const prev = WrapError.parse(message);
```
Into this:
```typescript
const prev = code === WrapErrorCode.WRAPPER_INVOKE_ABORTED
  ? WrapError.parse(message)
  : new WrapError(message)
```
Or something to that effect.

Since the only other code that gets passed here is WRAPPER_INTERNAL_ERROR, you could also write it like this.
```typescript
const prev = code === WrapErrorCode.WRAPPER_INTERNAL_ERROR
  ? new WrapError(message)
  : WrapError.parse(message)
```

Ofc, it is a bit of a bad practice to rely on enums like this, so a better solution could be to replace `abort` with `abortWithInvokeAborted` and `abortWithInternalError`, e.g.
```
  const abortWithInvokeAborted = (
        serializedMessage: string,
        source: ErrorSource
      ) => {
        const code = WrapErrorCode.WRAPPER_INVOKE_ABORTED;
        const prev = WrapError.parse(message);
        ....
    };
    
   const abortWithInternalError = (
        message: string
      ) => {
        const code = WrapErrorCode.WRAPPER_INTERNAL_ERROR;
        const prev = new WrapError(message);
        ....
    }
```

Though I don't think you even need prev for the internal error case.

packages/js/core/src/types/WrapError.ts
nerfZael: Sure, I'm down to keep it `unknown` for now.

nerfZael (changes requested):

nerfZael:
> @nerfZael
> 
> Somehow I can't reply to some of your messages so I'm responding here.
> 
> > In cases where we're reading from a buffer, and we get a zero there, I think the buffer should be treated as corrupted and not a valid WrapError. I do not think we should be allowing undefined states for an enum we control. The error codes, IMO, should start at 1.
> > Still, I am fine with keeping this for the time being and being more strict about the codes in the future.
> 
> I removed `UNKNOWN` and am starting the enum at 1.
> 
> > There's 2 things one can parse: serialized WrapErrors and errors from wasm modules (which may or may not be WrapErrors).
> > The serialized WrapError parsing logic should be in the WrapError, or somewhere close. It is easy to parse since we know the structure and does not need regex.
> > The wasm module error parsing logic should be in wasm-js. And since it can be anything regex is fine there.
> > The parsing logic here is both of those things.
> 
> All errors from Wasm modules are made to be of type `WrapError`.
> 
> It is possible that a message is not a stringified `WrapError`, but still we need to check whether it is or not (I explain why in your other question about `parse`). The `parse` method returns `undefined` when a message is not a `WrapError`, in which case `message` is used as the error message in a new `WrapError`.

We can address this in a followup PR when we use msgpack for error serialization

packages/js/wasm/src/WasmWrapper.ts
krisbitney: Ohhhh okay. That makes complete sense. I totally misunderstood your initial question. I replaced `abort` with two methods: `abortWithInvokeAborted` and `abortWithInternalError`.

krisbitney:
> > @nerfZael
> > Somehow I can't reply to some of your messages so I'm responding here.
> > > In cases where we're reading from a buffer, and we get a zero there, I think the buffer should be treated as corrupted and not a valid WrapError. I do not think we should be allowing undefined states for an enum we control. The error codes, IMO, should start at 1.
> > > Still, I am fine with keeping this for the time being and being more strict about the codes in the future.
> > 
> > 
> > I removed `UNKNOWN` and am starting the enum at 1.
> > > There's 2 things one can parse: serialized WrapErrors and errors from wasm modules (which may or may not be WrapErrors).
> > > The serialized WrapError parsing logic should be in the WrapError, or somewhere close. It is easy to parse since we know the structure and does not need regex.
> > > The wasm module error parsing logic should be in wasm-js. And since it can be anything regex is fine there.
> > > The parsing logic here is both of those things.
> > 
> > 
> > All errors from Wasm modules are made to be of type `WrapError`.
> > It is possible that a message is not a stringified `WrapError`, but still we need to check whether it is or not (I explain why in your other question about `parse`). The `parse` method returns `undefined` when a message is not a `WrapError`, in which case `message` is used as the error message in a new `WrapError`.
> 
> We can address this in a followup PR when we use msgpack for error serialization

Okay, sounds good.

nerfZael (dismissed):
LGTM!

dOrgJelli (approved):

PR #1436, opened 2022-11-24
break: move to using new logging wrappers

dOrgJelli:
Wrappers have been moved here: https://github.com/polywrap/logging


pileks (approved):
LGTM!

PR #1441, opened 2022-12-05
Add timeout option to http request in plugin

krisbitney:
This PR adds `timeout` as an option in http requests for the http plugin


cbrzn (commented):
would it be possible to add tests before merging this?

dOrgJelli (approved):

dOrgJelli (approved):
I don't think tests are necessary since this is a feature within axios, so no need to necessarily test their package ourselves, we can safely assume it works.

PR #1445, opened 2022-12-14
Tests for embedded wrappers

### LINKED ISSUE #1077: Wrappers can be embedded into applications
moved here from https://github.com/polywrap/documentation/issues/111

nerfZael:


packages/js/client/src/__tests__/core/embedded-package.spec.ts
```diff
@@ -0,0 +1,173 @@
+import fs from "fs";
```
cbrzn, line 1: are these on purpose? i guess they should only live in `core-client` and they are repeated?

nerfZael: They are, they test embedded wrappers when added through the `PolywrapClient` which uses our default config with the config builder

dOrgJelli (approved):
Nice!

PR #1446, opened 2022-12-15
replace ethereum plugin with ethereum wrapper

krisbitney:
This PR removes the ethereum plugin and replaces it with the ethereum wrapper.

## Notes on deployment

The ethereum wrapper is in a new repo: https://github.com/polywrap/ethereum. If you want to be added as a codeowner, just create a PR adding your name or let me know and I'll add you.

The plugin code has been moved to https://github.com/polywrap/ethereum/tree/main/plugin-js. Once this PR is merged, we can do a final NPM release indicating the plugin is deprecated.

The ethereum provider plugin that supports the wrapper is currently published to npm as `ethereum-provider-js`.
The ethereum wrapper is currently published to ENS at `ens/wraps.eth:ethereum@1.0.0`.

## Differences between the ethereum wrapper and ethereum plugin

The ethereum wrapper has nearly the same interface and behavior as the ethereum plugin, with a few differences.

First, the ethereum wrapper adds support for EIP 1559 transactions using `maxFeePerGas` and `maxPriorityFeePerGas`. It retains support for legacy transactions with `gasPrice`.

Second, the ethereum wrapper does not yet support timeouts. The method `waitForEvent` is not implemented. The method `awaitTransaction` is implemented, but it only waits for one confirmation and does so without a timeout.

Third, tuple formatting is more restrictive for human-readable ABIs. Whereas the etherum plugin supported tuples written like  `"tuple(string foo, uint256 bar, uint8 baz)"`, the ethereum wrapper requires the same tuple to be written `"(string,uint256,uint8)"`. The word "tuple", the parameter names, and the spaces between the commons are all absent in the latter expression.


packages/js/client-config-builder/src/bundles/getDefaultConfig.ts
```diff
@@ -46,6 +47,7 @@ export const defaultInterfaces = {
   uriResolver: "wrap://ens/uri-resolver.core.polywrap.eth",
   concurrent: "wrap://ens/goerli/interface.concurrent.wrappers.eth",
   logger: "wrap://ens/wrappers.polywrap.eth:logger@1.0.0",
+  ethereumProvider: "wrap://ens/iprovider.polywrap.eth",
```
Niraj-Kamdar: Can we have a better name here?
Ex: `provider.ethereum.polywrap.eth` ?

packages/cli/src/lib/defaults/deploy-modules/ens-recursive-name-register/getClient.ts
```diff
@@ -0,0 +1,96 @@
...
+            uri: defaultPackages.ethereumProvider,
+            package: ethereumProviderPlugin({ connections }),
+          },
+        ]),
+        new ExtendableUriResolver(),
+      ],
+      new WrapperCache()
+    )
+  );
+  const coreClientConfig = new ClientConfigBuilder(undefined, uriResolver)
```
Niraj-Kamdar: Why aren't we using addDefaults?

packages/js/plugins/uri-resolvers/ipfs-resolver/src/index.ts
```diff
@@ -46,62 +47,38 @@ export class IpfsResolverPlugin extends Module<NoConfig> {
       );
 
       if (manifestResult.ok) {
-        manifest = Buffer.from(manifestResult.value);
+        try {
+          manifest = Buffer.from(manifestResult.value);
+          return { uri: null, manifest: manifest };
+        } catch (e) {
+          break;
+        }
```
Niraj-Kamdar: Why do we have try...catch... here?

Niraj-Kamdar: Would conditionals work for this?

Niraj-Kamdar (changes requested):

packages/js/client-config-builder/src/bundles/getDefaultConfig.ts
krisbitney: The purpose is to clarify that it refers to the provider interface, not to an implementation. I actually don't like the ambiguous naming used elsewhere.

Niraj-Kamdar: It's ambiguous since it doesn't mention it's ethereum provider interface. Maybe
`iprovider.ethereum.polywrap.eth` or `interface.provider.ethereum.polywrap.eth`?

packages/js/plugins/uri-resolvers/ipfs-resolver/src/index.ts
krisbitney: The try catch was always there. I just moved it to be only around the only line that can throw. It is necessary because uri resolver extensions never throw; they return `null` instead. All of them work this way. 

There is an issue to address this: https://github.com/polywrap/toolchain/issues/663

And the specific try catch block here is temporary anyway, because there is another PR up that will replace the ipfs resolver plugin with a Wasm wrapper. I would actually love to get that one merged first since there will be merge conflicts. (Also, I fixed the reliability issue so we won't have problems with ipfs URIs not resolving).

packages/js/client-config-builder/src/bundles/getDefaultConfig.ts
krisbitney: Okay, I'll change it.

A name prefixed with `i` usually means interface (e.g. `IUriWrapper`, `IUriPackage`, `IUriResolver`). So I thought it made sense here. But I don't want it to be unclear to other people.

Niraj-Kamdar: i prefix isn't the problem. not mentioning the `ethereum` in uri is ambiguous.  There can be many kinds of providers EX: IPFS Provider, Tezos Provider, Near Provider, etc

krisbitney: ohhhh I see what you're saying

Niraj-Kamdar: We never prefixed the interface with `i` in the URI before (maybe we should?) 
Ex: we assumed ethereum interface to be deployed on `ethereum.polywrap.eth`, ipfs on `ipfs.polywrap.eth`. it was alright since we were redirecting these interface uris to plugins but we should make distinction more clear.

We are still doing the same thing with the logging/logger :https://github.com/polywrap/logging/blob/master/aggregator/resources/README.md

Where logger interface is deployed on the `ens/wrappers.polywrap.eth:logger@1.0.0`. This uri doesn't convey that it's an interface. 

While I always have been adding interface subdomain to actual wrapper domain to show that uri is an interface (Ex: `interface.price.resolver.defiwrapper.eth`)  and an implementation at: `coingecko.price.resolver.defiwrapper.eth`

The question is whether we should explicitly add prefixes to show the uri is interface or a wrapper?

packages/js/plugins/uri-resolvers/ipfs-resolver/src/index.ts
Niraj-Kamdar: It's just `Buffer.from` and I feel like it maybe handled without try...catch by checking the value but it's alright if we are going to replace if with wrapper anyway

packages/js/client-config-builder/src/bundles/getDefaultConfig.ts
krisbitney: Yeah, I see. You're 100% right. 

Right now we are publishing everything to `ens/wrappers.polywrap.eth`. Maybe we should have `ens/interfaces.polywrap.eth` as well?

packages/cli/src/lib/defaults/deploy-modules/ens-recursive-name-register/getClient.ts
krisbitney: Fixed! The complicated config without defaults was leftover from when I was trying to fix something, and I forgot to change it back.

packages/js/client-config-builder/src/bundles/getDefaultConfig.ts
krisbitney: I updated with the uri `ens/wrappers.polywrap.eth:ethereum-provider@1.0.0`

packages/js/plugins/uri-resolvers/ipfs-resolver/src/index.ts
krisbitney: Yeah, you're right. I removed the try-catch.

packages/cli/package.json
```diff
@@ -39,7 +39,7 @@
     "colors": "1.4.0"
   },
   "dependencies": {
-    "wraplib": "0.1.0",
+    "wraplib": "0.1.1",
```
Niraj-Kamdar: Why are we using this library? Can't we just move utility defined in the library to the appropriate package in toolchain Ex: UriResolver to uri-resolver-js

Niraj-Kamdar: If we really want to create a new utility package then we should have it limited to a scope and have better naming than a generic wraplib name.

krisbitney: When I put the `invokeTimeout` function in the CLI lib, I think it gets cut off by build optimizations. The default deployment modules can't find it. It works to have it duplicated in each of the deployment modules, but I thought it made sense to put it somewhere else where it can be easily tested and maintained. It doesn't naturally fit in any of the existing packages.

I created the wraplib library for experimentation, to play around with new ideas. I'm not planning to change the name or scope of it soon. I'll refactor it as it grows.

krisbitney: I added copies of `invokeWithTimeout.ts` to each of the default ens deployment modules and removed `wraplib` as a dependency.

packages/js/client-config-builder/src/bundles/getDefaultConfig.ts
krisbitney: Can I mark this resolved?

packages/cli/src/lib/defaults/deploy-modules/ens/invokeWithTimeout.ts
```diff
@@ -0,0 +1,56 @@
+import {
```
Niraj-Kamdar, line 1: Instead of replicating this to multiple deploy modules, Why don't we create utils and put it there
You can use this: `packages/cli/src/lib/helpers` or create utils inside `packages/cli/src/lib/defaults/deploy-modules`

Niraj-Kamdar (changes requested):

packages/cli/package.json
Niraj-Kamdar: Oh, I see now what you meant. It's weird!

Niraj-Kamdar: I honestly don't know what would be the best solution in this case. I think instead of patching around the actual problem `When I put the invokeTimeout function in the CLI lib, I think it gets cut off by build optimizations. The default deployment modules can't find it.` We should investigate it. maybe @namesty can help?

dOrgJelli (approved):
One small step for the toolchain repo 🌔, one giant leap for wrapper kind 🌟

PR #1447, opened 2022-12-15
feat: Configurable ExtendableUriResolver Interface URI

dOrgJelli:


krisbitney (approved):

packages/js/client/src/__tests__/core/uri-resolution/uri-resolution.spec.ts
```diff
@@ -197,7 +196,7 @@ describe("URI resolution", () => {
     const client = new PolywrapClient({
       interfaces: [
         {
-          interface: coreInterfaceUris.uriResolver.uri,
+          interface: ExtendableUriResolver.interfaceUri.uri,
```
nerfZael, line 199: Is `.uri` neccessary?

packages/js/uri-resolver-extensions/src/ExtendableUriResolver.ts
```diff
@@ -18,11 +17,20 @@ export class ExtendableUriResolver extends UriResolverAggregatorBase<
   Error,
   Error
 > {
-  private readonly resolverName: string;
+  public static interfaceUri: Uri = new Uri(
```
nerfZael, line 20: `extensionInterfaceUri` maybe?
`interfaceUri` kind of implies it's the interface of the ExtendableUriResolver instead of the resolve extension

packages/js/client/src/__tests__/core/uri-resolution/uri-resolution.spec.ts
dOrgJelli: No it's not necessary, good catch.

packages/js/uri-resolver-extensions/src/ExtendableUriResolver.ts
dOrgJelli: Agreed, this does look much nicer.

PR #1457, opened 2022-12-20
Update readme for client and other packages

krisbitney:
This is a draft PR for the purpose of preparing readme documents for the 0.10.0 release. The readme docs will be used for reference documentation on our docs site.

Steps to update a typical readme with reference documentation:
1. If a readme does not exist, create one with an introduction, some usage examples, etc.
2. Add doc comments to exported methods and types.
3. Add `doc-snippets` tokens around the comments and types or method signatures.
4. Use the `doc-snippets` CLI tool to inject the snippets into the readme doc.

You can copy the `doc-snippets` setup and configuration from repos where this has been done.

The aforementioned steps won't make sense for every package. Feel free to adapt as necessary!

TODO:
- [x] @polywrap/client-js
- [x] @polywrap/core-client-js
- [x] @polywrap/client-config-builder-js
- [x] @polywrap/core-js
- [x] polywrap (CLI)
- [x] @polywrap/test-env-js
- [ ] @polywrap/uri-resolvers-js
- [ ] @polywrap/uri-resolver-extensions-js
- [ ] @polywrap/package-validation
- [ ] @polywrap/wasm-js
- [ ] @polywrap/plugin-js
- [ ] @polywrap/tracing-js
- [ ] @polywrap/result
- [ ] @polywrap/msgpack-js
- [ ] @polywrap/wrap-manifest-types-js
- [ ] @polywrap/polywrap-manifest-types-js
- [ ] @polywrap/logging-js
- [ ] @polywrap/asyncify-js
- [ ] @polywrap/os-js


packages/js/client/README.md
```diff
@@ -70,15 +82,14 @@ const config = {
...
   // customize URI resolution
-  resolver: new RecursiveResolver(
-    new PackageToWrapperCacheResolver(wrapperCache, [
-      new ExtendableUriResolver(),
-    ])
-  ),
+  resolvers: [
+    new RecursiveResolver(
+      new PackageToWrapperCacheResolver(wrapperCache, [
+        new ExtendableUriResolver(),
```
pileks: Do we have any docs that explain the various resolvers, at least the ones that come with Polywrap by default?

This might help developers set up their own resolution logic.

I'm thinking we could add this to `packages/js/uri-resolver-extensions/README.md` and reference it here? @nerfZael thoughts?

pileks (dismissed):
Overall LGTM! Comment is non-blocking.

packages/js/client/README.md
krisbitney: Imo that's a good idea. I can work on it soon. I was also thinking to write more detailed reference documentation (in the readme) for the client config builder since it's becoming so important.

It would be cool to create good reference docs (as readmes) for wasm-js and plugin-js, and really every package.

pileks: Ideally we'd have the package creators also write up documentation, or at least a basic outline of it (with functional details), and have someone else (e.g. me, you, any volunteer) fix it up to spec.

krisbitney: yes, 100%

packages/js/core-client/README.md
```diff
@@ -6,3 +6,382 @@
...
+
+const config = { ... };
+const client = new PolywrapCoreClient(config);
+```
+
+### Invoke a wrapper
+
+```ts
+await client.invoke({
+  uri: "ens/rinkeby/helloworld.dev.polywrap.eth",
```
cbrzn:
```suggestion
  uri: "ens/helloworld.dev.polywrap.eth",
```

packages/js/client/README.md
```diff
@@ -7,6 +7,8 @@
 <br/>
 The Polywrap JavaScript client invokes functions of wrappers and plugins. It's designed to run in any environment that can execute JavaScript (think websites, node scripts, etc.). It has TypeScript support.
```
cbrzn: should this paragraph be moved to `core-client`? the reasoning is that this package doesn't take the responsibility of the invocation (or any), it just exists to have an ez way to config the client, right?

packages/js/core-client/README.md
cbrzn: also: i would also suggest this in the `client/README.md` but gh doesn't allow me

packages/js/client/README.md
```diff
@@ -45,11 +47,21 @@ const config = {
       to: "wrap://ens/to.eth",
     }
   ],
-  // declare and configure plugin wrappers
-  plugins: [
+  // add embedded Wasm wrappers
```
cbrzn: in line `43` of this file, there's `// redirect queries from one uri to another` - I would recommend to change it to `// redirect invocation from one uri to another`

it doesn't allow me to add the comment in that line because it wasn't modified lol

pileks:
Should we consider merging this and opening documentation issues for the remaining packages?

krisbitney:
> Should we consider merging this and opening documentation issues for the remaining packages?

Yeah, I think that's a good idea. It will prevent later work fixing merge conflicts.

pileks (dismissed):

pileks (dismissed):

packages/js/client-config-builder/README.md
```diff
@@ -42,15 +42,470 @@ const config = builder.build();
...
+   *   */
+  readonly resolvers: UriResolverLike[];
+}
+```
+
+## ClientConfigBuilder
+
+### Constructor
+```ts
+   * Instantiate a ClientConfigBuilder
```
dOrgJelli, line 224: This section doesn't appear to be rendering correctly, it's missing the starting `/*`

packages/js/client-config-builder/README.md
```diff
@@ -42,15 +42,470 @@ const config = builder.build();
 let client = new PolywrapClient(config);
 ```
 
-## Methods
+## Types
```
dOrgJelli, line 180: Above this line, in the introduction section, I think it could use some work. A few thoughts:
- I don't think "DSL" is properly used here.
- Ideally all code snippets should be from-source so that they remain up-to-date, so this one that's not a snippet worries me as it can become out-of-date.
- I think should start with introducing the main concepts a user must understand to get started. We can introduce the intended usage pattern of `init -> configure -> build`, and we can point to sub-sections for each. "Initialization" shows the constructor signature. "Configure" shows the different "add..." methods. etc
- Lastly, in this intro section, having an e2e example snippet (taken from a buildable source file) would be great to have.

So in summary:
- Brief Description
- Intro Concepts (linking to reference sections further down)
- E2E Example Snippet

Apologies if this is overly verbose :P

packages/js/client-config-builder/README.md
```diff
@@ -42,15 +42,470 @@ const config = builder.build();
...
+   *   | IUriRedirect<Uri | string>
+   *   | IUriPackage<Uri | string>
+   *   | IUriWrapper<Uri | string>
+   *   | UriResolverLike[]
+   *   */
+  readonly resolvers: UriResolverLike[];
+}
+```
+
+## ClientConfigBuilder
```
dOrgJelli, line 219: Can we group the `ClientConfigBuilder` and `ClientConfig` sections underneath a `Reference` section?

packages/js/client-config-builder/package.json
```diff
@@ -16,7 +16,8 @@
     "lint": "eslint --color -c ../../../.eslintrc.js src/",
     "test": "jest --passWithNoTests --runInBand --verbose",
     "test:ci": "jest --passWithNoTests --runInBand --verbose",
-    "test:watch": "jest --watch --passWithNoTests --verbose"
+    "test:watch": "jest --watch --passWithNoTests --verbose",
+    "build:readme": "yarn doc-snippets combine"
```
dOrgJelli, line 22: I worry that this "build:readme" can be easily forgotten about.

I think for inspiration we can  look at how we build the CLI package. We have a `build` command which does a lot of different build steps, and then there is `build:fast` which only builds the source files.

Maybe a similar approach here would make sense?

packages/js/client-config-builder/package.json
```diff
@@ -45,5 +47,41 @@
...
+      "dir": "./readme",
+      "include": "./README.md"
+    },
+    "startTokens": [
+      {
+        "pattern": "$start: ",
+        "inline": false
+      },
+      {
+        "pattern": "/* $: {SNIPPET_NAME} */",
```
dOrgJelli, line 77: Why do we have `{SNIPPET_NAME}` here, but not in the above start token?

dOrgJelli: Oh nvm, I think I figured it out, it's because this one is "inline" so it can't know where to extract the snippet name from like it can in the pattern above. Is that correct?

packages/js/client/readme/README.md
```diff
@@ -0,0 +1,118 @@
...
+```ts
+import { PolywrapClient } from "@polywrap/client-js";
+
+const client = new PolywrapClient();
+```
+
+### Invoke a wrapper
+
+```ts
+await client.invoke({
```
dOrgJelli: We should convert this to a code snippet so we make sure it runs. Maybe extract it from the tests folder or something.

dOrgJelli: Also it seems like we have this problem with the other READMEs I'm seeing here, all the example code snippets aren't being extracted from source. Might I suggest us adding an `examples/` folder next to the `src/` folder where we keep these examples? They would be built so we can make sure they are correct, but their artifacts will not make it into the `build/` folder so they do not get published with the built `src/` files.

dOrgJelli: Another option is to create a `src/__snippets__` folder, that lives alongside the `src/__tests__` folder.

dOrgJelli (changes requested):
This all looks great! Just a few improvements I think are needed before merging.

packages/js/client-config-builder/package.json
krisbitney: right, that's correct. The `{SNIPPET_NAME}` helps define a regex pattern.

packages/js/client/readme/README.md
krisbitney: I added `examples/` folders and set up snippets in `client`, `core-client`, `client-config-builder`, and `test-env`.

packages/js/client-config-builder/package.json
krisbitney: done!

packages/js/client-config-builder/README.md
krisbitney: done!

packages/js/client-config-builder/README.md
krisbitney: i reworked this intro and used snippets as suggested in your other comment

packages/js/client-config-builder/README.md
krisbitney: fixed!

krisbitney:
@dOrgJelli Your suggestions were really good. Thanks for looking at this! 

I think I addressed all of your comments: I updated the client config builder readme. All code examples now use snippets that build with `yarn build`. We can use `yarn build:fast` to just build the source code. 

I think it's in much better shape and ready for a new review!

dOrgJelli (dismissed):
:clap: :fire: :heavy_check_mark:

PR #1460, opened 2022-12-20
feat(wasm-as): codegen should create IModule interface

### LINKED ISSUE #1458: Wasm-as wrapper developers extend & implement IModule interface from wrap codegen
Currently, we have the module type in our GraphQL schema which contains functions that wrapper devs would implement but unlike plugin where plugin devs can easily extend the generated `IModule` interface, In the case of the wrapper, wrap dev needs to define the global functions with appropriate args, and return type. It would be more developer-friendly if we also provide an interface that developers can extend and provide an implementation for.

Let's take an example of a simple wrapper with the following schema:
```graphql
type Module {
  simpleMethod(arg: String!): String!
}
```
**Now let's see how the codegen and wrap devx. would look like for Assemblyscript dev.**
Here's how generated IModule interface would look for the simple wrapper with above schema
```typescript
export abstract class IModule {
  abstract simpleMethod(
    args: Types.Args_simpleMethod
  ): string;
}
```

Here's how a wrap dev can implement the `Module` class by simply extending it
```typescript
export class Module extends IModule {
  simpleMethod(args: Args_simpleMethod): string {
    return args.arg;
  }
}
```
For method that needs to use env, `env` can be accessed with `this.env` readonly property.

Niraj-Kamdar:
- Closes: #1458 
### Features
- Codegen for wasm wrappers now generates ModuleBase class that can be extended by wrap dev

## Future work
This PR only changes devx for wrapper development for the Assemblyscript wrapper. Rust wrapper will be covered in the following PR #1464


packages/test-cases/cases/bind/sanity/output/wasm-as/Module/module.ts
```diff
@@ -0,0 +1,168 @@
...
+  private ifWrapped(argsBuf: ArrayBuffer, env_size: u32): ArrayBuffer {
+    const args = deserializeifArgs(argsBuf);
+
+    const result = this._if(
+      {
+        _if: args._if
+      }
+    );
+    return serializeifResult(result);
+  }
```
dOrgJelli: To me, these functions should remain "external" and "static", just as they used to be, because they do not require access to the internals of the module instance. Instead they should call into the module instance from the outside. This will help keep the module class small.

packages/test-cases/cases/bind/sanity/output/wasm-as/Module/module.ts
```diff
@@ -0,0 +1,168 @@
...
+      throw new Error("Environment is not set, and it is required by this module");
+    }
+    return this._env;
+  }
+
+  public setEnv(env: Types.Env): void {
+    this._env = env;
+  }
+
+  public _wrap_invoke(method_size: u32, args_size: u32, env_size: u32): bool {
```
dOrgJelli: NOTE: the `_wrap_name` function naming convention is reserved for wasm exports, just as `__wrap_name` is reserved for wasm imports.

packages/templates/app/typescript-react/package.json
```diff
@@ -14,8 +14,8 @@
     "@polywrap/react": "0.10.0-pre.6",
     "http-proxy": "1.18.1",
     "http-proxy-middleware": "1.0.6",
-    "react": "16.13.1",
-    "react-dom": "16.13.1",
+    "react": ">=18.0.0",
```
dOrgJelli: I'd prefer we revert this. If you merge the latest `origin-dev` branch, there is a patch fix for the `yarn` install error that these packages were causing: https://github.com/polywrap/toolchain/blob/6d2355f25c78597e4a125bf9818dd31daf0b2de9/package.json#L74-L77

packages/schema/bind/src/bindings/utils.ts
```diff
@@ -25,6 +26,7 @@ export function renderTemplates(
 
           // If the file isn't empty, add it to the output
           if (data) {
+            // fs.writeFileSync(path.join("/Users/niraj/Documents/projects/polywrap/monorepo/packages/schema/bind/src/bindings/assemblyscript/wasm/templates/amodule-type", name.replace("-", ".")), data);
```
dOrgJelli: remove?

packages/schema/bind/src/bindings/assemblyscript/wasm/index.ts
```diff
@@ -136,12 +136,15 @@ export const generateBinding: GenerateBindingFn = (
 
   // Generate module type folders
   if (abi.moduleType) {
+    const moduleType = abi.envType
+      ? { ...abi.moduleType, hasEnv: true }
```
dOrgJelli: The proper way to do this would be to add another transform, see here: https://github.com/polywrap/toolchain/blob/65d7416b5cb78aa51a9267903afb66c5c6e7a975/packages/schema/bind/src/bindings/assemblyscript/wasm/index.ts#L187-L198

dOrgJelli (changes requested):
This looks great! Added a few improvements I think could help.

packages/test-cases/cases/bind/sanity/output/wasm-as/Module/module.ts
Niraj-Kamdar: Yeah but the problem with that was that since the module is now class `module.ifWrapped` method isn't simply a function that InvokeFunction can represent but also require the Module instance as its argument so the function signature changes there and invoke function not only need function but also need object instance for it to work. 

I tried to pass `this.xyzWrapped` directly but it didn't work, I also tried to use closure but they aren't supported. So, I bundled it as we did in PluginModule.

There may be a better way to do this for sure but this is the one that works for me currently.

packages/test-cases/cases/bind/sanity/output/wasm-as/Module/module.ts
Niraj-Kamdar: Sure we can rename it to anything, I just keep it the way we did for the `PluginModule`.

packages/templates/app/typescript-react/package.json
Niraj-Kamdar: Oh great! Thanks for letting me know! I just patched it like that because `yarn install` was failing, and at that time the fix wasn't there. Since this worked for me, I didn't want to dig deep given that this is a draft anyway

packages/schema/bind/src/bindings/utils.ts
Niraj-Kamdar: Yeah, that's just for debugging, I will remove it once it's ready to merge. This is helping me debug it easily.

packages/schema/bind/src/bindings/assemblyscript/wasm/index.ts
Niraj-Kamdar: Yeah for sure!

packages/test-cases/cases/bind/sanity/output/wasm-as/Module/module.ts
Niraj-Kamdar: Nevermind I got the solution!

cbrzn:
> Chore
> Moved validate wrapper test-cases from utils to test-cases directory

I am not sure about moving this outside of the client package - My main thought is that we want to eventually move the client to its own repository (sooner than later) and this might introduce more changes/thoughts to the new structure when it comes the time?

I agree that the `utils` folder is not the best way to do this, but I also think that moving it to the general `test-cases` folder is not a good long-term solution

cbrzn:
> Chore
> Moved validate wrapper test-cases from utils to test-cases directory

I am not sure about moving this outside of the client package - My main thought is that we want to eventually move the client to its own repository (sooner than later) and this might introduce more changes/thoughts to the new structure when it comes the time?

I agree that the `utils` folder is not the best way to do this, but I also think that moving it to the general `test-cases` folder is not a good long-term solution

Niraj-Kamdar:
@cbrzn Until we move the client to its own repo, I think we should keep it the way we have been doing with the test-cases.

packages/schema/parse/src/transform/moduleHasEnv.ts
```diff
@@ -0,0 +1,19 @@
+import { AbiTransforms } from ".";
+
+import { GenericDefinition, WrapAbi } from "@polywrap/wrap-manifest-types-js";
+
+let hasEnv = false;
```
krisbitney: Instead of using a global variable, can you do a function like this?

```typescript
export const moduleHasEnv = (): AbiTransforms => {
  let hasEnv = false;

  return {
    enter: {
      Abi: (abi: WrapAbi): WrapAbi => {
        hasEnv = !!abi.envType;
        return abi;
      },
    },
    leave: {
      ModuleDefinition: (moduleDefinition: GenericDefinition) => ({
        ...moduleDefinition,
        hasEnv,
      }),
    },
  };
};
```

packages/schema/bind/src/bindings/assemblyscript/wasm/templates/entry-ts.mustache
```diff
@@ -15,21 +16,28 @@ import {
...
     method_size,
     args_size
   );
-
   {{#moduleType}}
   {{#methods}}
   {{^first}}else {{/first}}if (args.method == "{{name}}") {
-    return wrap_invoke(args, env_size, {{name}}Wrapped);
+    const result = {{name}}Wrapped(module, args.args, env_size);
+    wrap_invoke_result(result);
```
krisbitney: the last two lines in each successfully-matched branch could be pulled out so it's not repeated. The test case at `test-cases/cases/bind/sanity/output/wasm-as/entry.ts` would look something like:

```typescript
export function _wrap_invoke(method_size: u32, args_size: u32, env_size: u32): bool {
  const module = new Module();
  const args: InvokeArgs = wrap_invoke_args(
    method_size,
    args_size
  );
  let result: ArrayBuffer;
  if (args.method == "moduleMethod") {
    result = moduleMethodWrapped(module, args.args, env_size);
  }
  else if (args.method == "objectMethod") {
    result = objectMethodWrapped(module, args.args, env_size);
  }
  else if (args.method == "optionalEnvMethod") {
    result = optionalEnvMethodWrapped(module, args.args, env_size);
  }
  else if (args.method == "if") {
    result = ifWrapped(module, args.args, env_size);
  }
  else {
    wrap_invoke_error(
      `Could not find invoke function "${args.method}"`
    );
    return false;
  }
  wrap_invoke_result(result);
  return true;
}
```

krisbitney (changes requested):
This is a fantastic PR. This will be really nice for the dev exp!! 

I requested a couple of super small changes that will take only a few minutes to fix.

krisbitney (dismissed):

packages/templates/app/typescript-react/package.json
```diff
@@ -14,8 +14,8 @@
     "@polywrap/react": "0.10.0-pre.6",
     "http-proxy": "1.18.1",
     "http-proxy-middleware": "1.0.6",
-    "react": "16.13.1",
```
nerfZael: Why the downgrade?

packages/schema/bind/src/bindings/assemblyscript/wasm/templates/module-type/module-ts.mustache
```diff
@@ -0,0 +1,43 @@
...
+
+import {
+  wrap_invoke_args,
+  wrap_invoke_result,
+  wrap_invoke_error,
+  wrap_load_env,
+  InvokeArgs,
+} from "@polywrap/wasm-as";
+
+export abstract class IModule {
```
nerfZael: I don't think abstract classes should be prefixed with `I`

nerfZael (commented):
If we're still supporting old way of defining modules we should have tests that cover that.

packages/schema/bind/src/bindings/assemblyscript/wasm/templates/module-type/module-ts.mustache
dOrgJelli: While I agree, I'm not sure what we should call this. A few options I could think of:
```typescript
export class Module extends BaseModule { }
```
```typescript
export class Module extends ModuleMethods { }
```
```typescript
export class Module extends WrapModule { }
```
```typescript
export class Module extends IModule { }
```
```typescript
export class Module extends Module.Interface { }
```
?

packages/schema/bind/src/bindings/assemblyscript/wasm/templates/module-type/module-ts.mustache
```diff
@@ -0,0 +1,43 @@
...
+  {{#hasEnv}}private __env__: Types.Env | null = null;
+
+  public get env(): Types.Env {
```
dOrgJelli: Hey @Niraj-Kamdar @namesty, I think we have a conflict in how we treat `env` here.

First, this getter method is added to the module bindings if the ABI has an `envType` defined (see `moduleHasEnv.ts` and `{{#hasEnv}}` above). It will throw an error if the `env` is not set, stating "it is required by this module".

Now for the inconsistency. We previously added the `@env(required: boolean)` directive so that we can selectively require the environment for specific module methods. For example, see `test-cases/cases/bind/sanity/input/schema.graphql`. This means that by default the `env` is optional for all methods, and is selectively required on a per-method basis.

I think we should take one of the 2 possible paths:
1. Respect this pattern of having `env` optional, and having it required on a per-method basis. This means that we will probably go back to passing the `env` as an argument to the method.
2. Revert the `@env` directive, and go back to requiring the `env` on a per-module basis.

Thoughts?

dOrgJelli: Actually after further inspection @Niraj-Kamdar, the per-method requirement of the environment is already handled well within the `wrap/Module/wrapped.ts` file. See the `objectMethodWrapped` example:
```typescript
export function objectMethodWrapped(module: IModule, argsBuf: ArrayBuffer, env_size: u32): ArrayBuffer {
  if (env_size == 0) {
    throw new Error("Environment is not set, and it is required by method 'objectMethod'")
  }
  
  const envBuf = wrap_load_env(env_size);
  module.__setEnv__(Types.Env.fromBuffer(envBuf));
  ...
}
```

This means that we do not have to handle this scenario within the `IModule` bindings, and we can instead do something like this:
```typescript
export abstract class IModule {
  private __env__: Types.Env | null = null;

  public get env(): Types.Env {
    if (this.__env__ === null) {
      throw new Error("Environment is not set, and is not required by this method.");
    }
    return this.__env__ as Types.Env;
  }

  public __setEnv__(env: Types.Env): void {
    this.__env__ = env;
  }
}
```

It's not the cleanest pattern... I would honestly prefer going back to having the `env` be an argument of the module's method.

packages/test-cases/cases/bind/sanity/output/wasm-as/Module/module.ts
```diff
@@ -0,0 +1,64 @@
...
+  serializeifResult
+} from "./serialization";
+
+import {
+  wrap_invoke_args,
+  wrap_invoke_result,
+  wrap_invoke_error,
+  wrap_load_env,
+  InvokeArgs,
+} from "@polywrap/wasm-as";
```
dOrgJelli, line 8: These imports can be minimized. All imports that have to do with serialization & internal `wrap_...` calls can be removed from what I can tell.

dOrgJelli (changes requested):

packages/templates/app/typescript-react/package.json
Niraj-Kamdar: It was an issue before but I guess it got solved later so can revert that

packages/schema/bind/src/bindings/assemblyscript/wasm/templates/module-type/module-ts.mustache
Niraj-Kamdar: This is to give the same ergonomics of the plugins to the wrapper. As long as the pattern is consistent I don't mind the way in which we implement it. 

If we prefer method args env then we should do the same in the plugin, if we prefer `this.env` then we should do the same in the wrapper.

packages/test-cases/cases/bind/sanity/output/wasm-as/Module/module.ts
Niraj-Kamdar: Yeah definitely, nice catch! forget to remove it after refactor

Niraj-Kamdar:
@cbrzn sorry for overwriting your commits. I got some git issues and wasn't able to fix it without force pushing.

packages/schema/bind/src/bindings/assemblyscript/wasm/templates/module-type/module-ts.mustache
Niraj-Kamdar: using env in function params

dOrgJelli (changes requested):
I'm very eager to get these changes merged, it looks amazing.

I want to make sure we give users a seamless experience and options to go back to the legacy codegen if desired: https://github.com/polywrap/toolchain/blob/dorgjelli/binding-import/vision.md

dOrgJelli:
I suppose 

> I'm very eager to get these changes merged, it looks amazing.
> 
> I want to make sure we give users a seamless experience and options to go back to the legacy codegen if desired: https://github.com/polywrap/toolchain/blob/dorgjelli/binding-import/vision.md

I suppose we could do this "legacy support" and the decoupling of the binding projects from the toolchain repo in another PR right after this one, thoughts @Niraj-Kamdar?

Niraj-Kamdar:
> https://github.com/polywrap/toolchain/blob/dorgjelli/binding-import/vision.md

It makes sense to me! And yes separate PR would be better. I don't want to make this one more monstrous than it already is!

dOrgJelli (dismissed):

PR #1461, opened 2022-12-21
break: remove client.query() & client.subscribe()

dOrgJelli:


Niraj-Kamdar (dismissed):
LGTM! What about subscription?

package.json
```diff
@@ -70,5 +70,9 @@
     "lerna": "4.0.0",
     "prettier": "2.2.1",
     "rimraf": "3.0.2"
+  },
+  "resolutions": {
```
nerfZael, line 74: Security patch or?

nerfZael (dismissed):

cbrzn:
> LGTM! What about subscription?

I would agree to remove subscription here too!

krisbitney:
> > LGTM! What about subscription?
> 
> I would agree to remove subscription here too!

Same, I don't think it's ever been used. It makes sense to remove it!

package.json
dOrgJelli: This was removed here: https://github.com/polywrap/toolchain/pull/1452

And it turns out it's needed within the top-level package.json because of a problem w/ package resolution within the monorepo. React is still used within the template project, and when you install the react template's deps directly (`cd packages/templates/app/typescript-react && yarn`) it works, but at the root monorepo level it fails due to mismatched dependencies.

PR #1464, opened 2022-12-24
feat(wasm-rs): codegen should create IModule interface 

### LINKED ISSUE #1463: Wasm-rs wrapper developers extend & implement IModule trait from wrap codegen
Currently, we have the module type in our GraphQL schema which contains functions that wrapper devs would implement but unlike plugin where plugin devs can easily extend the generated `IModule` interface, In the case of the wrapper, wrap dev needs to define the global functions with appropriate args, and return type. It would be more developer-friendly if we also provide an interface that developers can extend and provide an implementation for.

Let's take an example of a simple wrapper with the following schema:
```graphql
type Module {
  simpleMethod(arg: String!): String!
}
```
**Now let's see how the codegen and wrap devx. would look like for Rust dev.**
Here's how generated IModule interface would look for the simple wrapper with above schema
```rust
pub struct Module {
  env: Option<Env>
}

pub trait IModule {
  fn simple_method(
    args: ArgsSimpleMethod
  ): String;
}
```

Here's how a wrap dev can implement the `IModule` trait for Module struct
```rust
impl IModule for Module {
  fn simple_method(args: ArgsSimpleMethod): String {
    return args.arg;
  }
}
```
Methods that need to access `env` can access it with `self.env`.

Niraj-Kamdar:
- Closes: https://github.com/polywrap/toolchain/issues/1463
### Features
- Codegen for wasm wrappers now generates ModuleTrait that can be implemented by wrap dev


packages/schema/bind/src/bindings/rust/wasm/templates/entry-rs.mustache
```diff
@@ -12,20 +12,29 @@ use polywrap_wasm_rs::{
...
 
+    let mut module = Module::__new__();
     let args: InvokeArgs = invoke::wrap_invoke_args(method_size, args_size);
 
     match args.method.as_str() {
         {{#moduleType}}
         {{#methods}}
-        "{{name}}" => invoke::wrap_invoke(args, env_size, Some({{#toLower}}{{name}}{{/toLower}}_wrapped)),
+        "{{name}}" => {
+            let result = {{#toLower}}{{name}}{{/toLower}}_wrapped(&mut module, args.args.as_slice(), env_size);
```
krisbitney: I think this part can be pulled out from the conditions so it's not repeated in each one: 

```rust
invoke::wrap_invoke_result(result);
true
```

packages/schema/bind/src/bindings/rust/wasm/templates/module-type/module-rs.mustache
```diff
@@ -0,0 +1,60 @@
...
+    {{#detectKeyword}}{{#toUpper}}{{type}}{{/toUpper}}{{/detectKeyword}},
+};
+{{/isEnum}}
+{{/propertyDeps}}
+{{/propertyDeps.length}}
+
+pub struct Module {
+  {{#hasEnv}}pub env: Option<Env>,{{/hasEnv}}
+}
+{{#hasEnv}}
```
krisbitney: Maybe add a return between the `Module` struct and the env trait?

krisbitney (changes requested):
LGTM! I had some super minor comments. 

These PRs are a huge improvement. Thanks for doing this!!

krisbitney (dismissed):
LGTM! Please fix test before merging

packages/schema/bind/src/bindings/rust/wasm/templates/module-type/module-rs.mustache
```diff
@@ -0,0 +1,61 @@
...
+
+pub trait EnvTrait {
+  fn __set_env__(&mut self, env: Env) -> ();
+}
+{{/hasEnv}}
+
+pub trait IModule {
```
nerfZael: I think we should find a better name for the IModule trait, prefixing it with `I` is not very Rust idiomatic

packages/test-cases/cases/wrappers/wasm-rs/asyncify/src/lib.rs
```diff
@@ -1,118 +1,113 @@
...
+use wrap::module::{IModule, Module};
+pub use wrap::*;
 
-pub fn get_data(args: module::ArgsGetData) -> u32 {
-    match StorageModule::get_data(
-        &storage_module::ArgsGetData {},
-    ) {
-        Ok(v) => v.try_into().unwrap(),
-        Err(e) => panic!("{}", e),
+impl IModule for Module {
```
nerfZael: Are we supporting both ways now? With IModule and without?
If so, we should have tests for the latter too.
Also, we should have the result return type for the latter as well

packages/schema/bind/src/bindings/rust/wasm/templates/module-type/module-rs.mustache
Niraj-Kamdar: ModuleTrait?

packages/test-cases/cases/wrappers/wasm-rs/asyncify/src/lib.rs
Niraj-Kamdar: Nope without IModule, it won't work. you need to implement all the methods from the IModule trait in Module.

Niraj-Kamdar: the older functions way is deprecated

PR #1465, opened 2022-12-26
refactor: client and client-config-builder

Niraj-Kamdar:
TODO: env redirects are failing
DEBUG: why custom config tests works locally but fails on CI
DEBUG: app codegen fails locally but works on CI


pileks:
Closing this PR in favor of #1480 as it contains all of @Niraj-Kamdar's changes.

PR #1467, opened 2023-01-03
chore: move plugin build logic into compiler

dOrgJelli:
NOTE: there is a bit of "hacky" logic inside of the `Compiler` class. I think the best way to do this in the future is to have a base `Compiler` class, and derive it into project-specific compilers (wasm, plugin, interface). But this can be an improvement in the future IMO.


pileks (approved):
Great thinking with the move to Compiler, I like it!

packages/cli/src/lib/helpers/wrap.ts
```diff
@@ -31,10 +37,16 @@ const run = async (
 export const generateWrapFile = async (
   abi: WrapAbi,
   name: string,
-  type: "interface" | "wasm" | "plugin",
+  type: string,
   path: string,
   logger: Logger
 ): Promise<void> => {
+  if (supportedWrapTypes.indexOf(type as WrapManifest["type"]) === -1) {
```
cbrzn: minor nit: I find more cleaner to use `find`/`includes` (in this case, `includes`) rather than `indexOf`

pileks: Agree, gonna fix it up.

cbrzn (approved):

Issue #1426, opened 2022-11-16
Able To Run `polywrap deploy` w/o `polywrap.yaml` Manifest


dOrgJelli:
**Describe the bug**
Currently developers need a valid `polywrap.yaml` project manifest defined in order to run `polywrap deploy`.

**Please provide the steps to reproduce and if possible a minimal demo of the problem.**
- Have a wrapper URI that's pre-built (ex: `/ipfs/QmfRCVA1MSAjUbrXXjya4xA9QHkbWeiKRsT7Um1cvrR7FY`)
- Create a `polywrap.deploy.yaml` manifest that deploys the wrapper, for example:
```yaml
format: 0.1.0
stages:
  ipfs_deploy:
    package: ipfs
    uri: fs/./build
    config:
      gatewayUri: "https://ipfs.wrappers.io"
```
- Try to run `polywrap deploy` in the director of `polywrap.deploy.yaml`, it will fail because no `polywrap.yaml` file is found.

**Expected behavior**
Users should be able to run `polywrap deploy` and only need a `polywrap.deploy.yaml` manifest defined.

PR #1476, opened 2023-01-10
feat: wrappers test cases removed from toolchain

### LINKED ISSUE #1203: Multi-Wasm Test Harness
Currently we do not have a clean way of testing multiple wasm languages, and forming a feature matrix. We now support AssemblyScript and Rust, and with TinyGo support on the way very soon it would be nice to have a system created for this.

See @cbrzn comment below for implementation details.

**Additional context**
Currently we have test wrappers that are duplicated in each language here: https://github.com/polywrap/toolchain/tree/origin/packages/test-cases/cases/wrappers

Once these issues has been closed, this issue can be considered done:
- [x] https://github.com/polywrap/wasm-test-harness/issues/1
- [x] https://github.com/polywrap/wasm-test-harness/issues/2
- [x] https://github.com/polywrap/wasm-test-harness/issues/3
- [ ] Remove test-cases/wasm from current codebase

cbrzn:
closes #1203 

depends on: #1482 (& not required: #1480)

this pr has removed the `packages/test-cases/cases/wrappers` folder, and the tests have been migrated to the [wasm test harness](https://github.com/polywrap/wasm-test-harness/tree/master/cases) repository.

client tests have also been updated with a new structure to differentiate between different client features (open to any feedback)

development guidelines have also been updated. would appreciate strong feedback there to make sure that the development flow is clear enough when it comes to updating test cases


cbrzn:
blocked by: https://github.com/polywrap/toolchain/issues/1426

.github/workflows/generate-modified-test-wrappers.yaml
```diff
@@ -0,0 +1,91 @@
...
+      wasm-path:
+        required: false
+        type: string
+
+jobs:
+  generate_wrappers:
+    name: Generate wrappers
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
```
krisbitney, line 18: I don't understand this fully. There is a lot I don't know about github workflows. It looks like you're checking out one repository right after another. Can you check out two at once?

package.json
```diff
@@ -24,8 +24,9 @@
     "reset": "yarn clean && yarn && yarn build",
     "clean": "npx rimraf ./**/node_modules ./**/yarn.lock ./**/build ./**/coverage ./**/.polywrap",
     "dependencies:install": "cd dependencies && yarn",
+    "test-wrappers:install": "lerna run generate:wrappers --scope @polywrap/test-cases",
```
krisbitney, line 27: It looks like there are two scripts added: `test-wrappers:install` and `build:test-cases`. I think we can remove one since they are the same.

packages/js/client/src/__tests__/core/error-structure.spec.ts
```diff
@@ -186,53 +168,56 @@ describe("Error structure", () => {
...
-      expect(result.ok).toBeFalsy();
-      if (result.ok) throw Error("should never happen");
-
-      expect(result.error?.name).toEqual("WrapError");
-      expect(result.error?.code).toEqual(WrapErrorCode.URI_RESOLVER_ERROR);
-      expect(result.error?.uri.endsWith("packages/test-cases/cases/wrappers/wasm-as/simple-deprecated")).toBeTruthy();
-      expect(result.error?.resolutionStack).toBeDefined();
-      expect(`${result.error?.cause}`).toContain(`Unrecognized WrapManifest schema version "0.0.1"`);
-    });
+    // test("Invoke a wrapper of incompatible version", async () => {
```
krisbitney: For this, maybe we could embed the "incompatible version" wrap manifest in the client tests folder and try to invoke it? I'm not sure there needs to be a wasm module if the wrap manifest version is incompatible.

packages/test-cases/index.ts
```diff
@@ -64,3 +67,43 @@ function getFilePath(
     return path.join(directory, file);
   }
 }
+
+export function fetchWrappers(): void {
```
krisbitney: I think this should be an async function and return `Promise<void>` just to make it super clear that it's doing an async task

DEV_GUIDELINES.md
```diff
@@ -43,6 +43,14 @@ Some tests rely on validating `stdout` for which Cue is used. If you need to run
 
 You can install Cue by following the instructions found [here](https://cuelang.org/docs/install/).
 
+In the test-cases directory, you can find the `wrappers` folder, which is auto generated from the releases of the
+[WASM Test Harness](https://github.com/polywrap/wasm-test-harness), check the `fetchWrappers` function from the [test-env package](./packages/test-cases/index.ts). These tests are used mostly for client tests, if you would like to
```
krisbitney: Typo: this should read "test-cases package" instead of "test-env package"

krisbitney (changes requested):
This is really awesome. It's going to be a massive boost to testing speed and dev productivity!

.github/workflows/generate-modified-test-wrappers.yaml
cbrzn: no worries, let me explain: the purpose of this job is to be able to interact with both repositories (toolchain & test harness) and be able to build the wrappers (that live in the harness) with the latest changes in the toolchain. (note that this only happens [when one of the packages](https://github.com/polywrap/toolchain/blob/026067f92407fe25044af1077763ba2dfb028981/.github/workflows/js-ci.yaml#L58-L61) necessary to build the wrappers has changed)

answering your question in concrete: yes, we're downloading both repositories in the workflow, being the toolchain [defined in the "toolchain" directory](https://github.com/polywrap/toolchain/blob/026067f92407fe25044af1077763ba2dfb028981/.github/workflows/generate-modified-test-wrappers.yaml#L29) (represented by the path attribute) - you can check [this](https://github.com/actions/checkout/#checkout-multiple-repos-side-by-side) for more info

so, based on this, what we're doing here is:
- download both repositories
- build the toolchain with the latest changes (from the base branch where pr has been opened)
- pass the path of cli & wasm package (in the ci) to wrapper generation script in the harness
- when wrappers are built successfully, we update them in the `packages/test-cases/cases/wrappers` folder
- tests will run as they have always do 

the purpose of this is to make sure that when we introduce changes, we make sure our test cases don't break and are updated with the latest changes.

it's also worth mentioning that you will need to create a branch **with the same name** in the test harness, with the modified wrappers (or not), to be able to run the workflow; otherwise, it will fail.

open to any feedback or suggestions, and please feel free to ask anything that needs to be clarified

package.json
cbrzn: nice catch ser

.github/workflows/generate-modified-test-wrappers.yaml
krisbitney: This is fantastic!

packages/js/client/src/__tests__/core/error-structure.spec.ts
cbrzn: i will try that! or maybe just modifying an existing manifest with another version just to test that it indeed throws

krisbitney: sounds like a nice solution

cbrzn: i just did an update but it is not working, when I modify the manifest and try to create a wasm wrapper (`WasmWrapper.from(...)`) it fails, because the version is not right (it does a validation on the creation of a wasm wrapper, which I think make sense) - Also tried with a plugin but the invocation passes, which probably means that we have a problem there. Will debug more on Monday but if you want to take a look feel free to :smile:

krisbitney: We should assume the wrapper was created with a previous version of the toolchain, and that's why it exists. I think we just need a wrap manifest with an old version, and to try to invoke it with a filesystem uri

krisbitney: you can use the wrapper that already exists in `origin-dev` that i was using for the test before

krisbitney: Regarding the plugin, we might want to try to import an old plugin made with prealpha polywrap and see what happens. There might be a problem before an invocation can even be made. Maybe we should create an issue to explore this more?

cbrzn: > We should assume the wrapper was created with a previous version of the toolchain, and that's why it exists. I think we just need a wrap manifest with an old version, and to try to invoke it with a filesystem uri

it worked as expected, you're a genius :1st_place_medal:, please check the test and let me know if this make sense :smile: 

> Regarding the plugin, we might want to try to import an old plugin made with prealpha polywrap and see what happens. There might be a problem before an invocation can even be made. Maybe we should create an issue to explore this more?

i went ahead and create a test case of a plugin invocation with a non-accepted version (which is currently skipped, because it's not failing), so I'd say it make sense to create the issue (feel free to check my test to make sure I am thinking it in the right way)

krisbitney (dismissed):
LGTM! Fantastic work

dOrgJelli:
@cbrzn added a fix here: https://github.com/polywrap/wrap-test-harness/pull/39

dOrgJelli (approved):

Issue #1283, opened 2022-09-26
JS Client: Client Config Should Use Unique Maps Instead Of Unique Arrays


dOrgJelli:
**Is your feature request related to a problem? Please describe.**
Currently the client's configuration structure uses arrays to store things like `redirects`, `envs`, `plugins`, etc. These arrays are "unique" based on each item's `uri` property.

This is not very clear from a user's perspective, and instead we should just use a unique map, where the `uri` is the map's key.

dOrgJelli:
TODO: revisit this taking into account the new `ClientConfig` changes. #1480

dOrgJelli:
Update: @pileks said this is ~60% done, is actively working on it, PR to come.

dOrgJelli:
PR: https://github.com/polywrap/toolchain/pull/1557

PR #1480, opened 2023-01-11
Refactor: ClientConfigBuilder-specific `BuilderConfig` object for configuration handling

pileks:
This is a continuation of #1465. Huge props to @Niraj-Kamdar for doing the bulk of the work around this.

This PR introduces the concept of `BuilderConfig`, which is a `ClientConfigBuilder`-specific configuration object wherein the individual configuration sections take the shape of a `Record<string, SectionType>`.

This allows us to configure the client in a very readable fashion.
Let's take `config.envs` as an example.
Before, we used to write:

```typescript
const envs = [
  {
    uri: Uri.from("wrap://ens/hello-world.polywrap.eth"),
    env: {
      foo: "bar",
    },
  },
  {
    uri: Uri.from("ens/ethereum.polywrap.eth"),
    env: {
      connection: {
        node: "https://mainnet.infura.io/v3/some_api_key",
        networkNameOrChainId: "mainnet",
      },
    },
  },
];

  .addDefaults()
  .addEnvs(envs)
  .build();

const client = new PolywrapClient(config);
```

Now we can write a much more succint form of the configuration:
```typescript
const envs = {
      "ens/ethereum.polywrap.eth": {
        connection: {
          networkNameOrChainId: "mainnet",
          node: "https://mainnet.infura.io/v3/some_api_key",
        },
      },
      "ens/hello-world.polywrap.eth": { foo: "bar" },
    };

  .addDefaults()
  .addEnvs(envs)
  .build();

const client = new PolywrapClient(config);
```

This can also be seen as precursor work to #1283 as it lays the foundation of the "ideal" layout of our final config objects.

Thank you @krisbitney for the note that this PR does a few more things than the description initially explained, these are:

- replace Uri | string with Uri
- remove tracing from core
- fix http server infra build


packages/cli/src/commands/build.ts
```diff
@@ -174,11 +173,13 @@ async function run(options: Required<BuildCommandOptions>) {
   } = options;
   const logger = createLogger({ verbose, quiet, logFile });
 
-  const envs = await parseWrapperEnvsOption(wrapperEnvs);
+  const envs = await parseWrapperEnvsOption(
+    wrapperEnvs ? wrapperEnvs : undefined
```
cbrzn: nit: what's the reason for adding this ternary if? we can just pass `wrapperEnvs`, right?

cbrzn (commented):
thanks for pushing this! looks great :-)

packages/cli/src/commands/build.ts
pileks: `wrapperEnvs` in the options is a `WrapperEnvs | false` type (thus making it essentially required, as per @dOrgJelli's CLI-JS PR). `parse` takes `WrapperEnvs | undefined`, so it's just a simple transform.

Niraj-Kamdar: Then parse should take `WrapperEnvs  | undefined | bool` I guess and do proper transformation inside the function.

pileks: Done!

packages/js/client-config-builder/src/bundles/getDefaultConfig.ts
```diff
@@ -25,147 +26,73 @@ export const defaultWrappers = {
   sha3: "wrap://ens/goerli/sha3.wrappers.eth",
   uts46: "wrap://ens/goerli/uts46-lite.wrappers.eth",
   graphNode: "wrap://ens/goerli/graph-node.wrappers.eth",
-  ensTextRecordResolver:
-    "wrap://ipfs/QmfRCVA1MSAjUbrXXjya4xA9QHkbWeiKRsT7Um1cvrR7FY",
+  concurrentInterface: "wrap://ens/goerli/interface.concurrent.wrappers.eth",
 };
 
-export const defaultPackages = {
```
krisbitney: Why are we removing these constants? We should instead be using them everywhere instead of string literals.

packages/js/client-config-builder/src/types/IClientConfigBuilder.ts
```diff
@@ -0,0 +1,43 @@
+import { BuilderConfig } from "./configs/BuilderConfig";
+import { ClientConfig } from "./configs/ClientConfig";
+
+import { CoreClientConfig, Wrapper, IWrapPackage } from "@polywrap/core-js";
+import { UriResolverLike } from "@polywrap/uri-resolvers-js";
+
+export type TUri = string;
```
krisbitney: The names TUri and TEnv could be misleading since they are aliases for specific types but names like generic types. I think we should instead just use the actual types (e.g. `string`) directly.

packages/js/client-config-builder/src/types/configs/PolywrapClientConfig.ts
```diff
@@ -0,0 +1,23 @@
...
+  IUriRedirect,
+  IUriWrapper,
+  UriResolverLike,
+} from "./types";
+
+import { IWrapperCache } from "@polywrap/uri-resolvers-js";
+import { TracerConfig } from "@polywrap/tracing-js";
+import { Uri } from "@polywrap/core-js";
+
+export interface PolywrapClientConfig<TUri extends Uri | string = string> {
```
krisbitney: I think this belongs with the PolywrapClient, not with the config builder. What was the thought process behind the move?

packages/js/client-config-builder/src/types/configs/PolywrapCoreClientConfig.ts
```diff
@@ -0,0 +1,15 @@
+import { Env, InterfaceImplementations } from "./types";
+
+import { IUriResolver, Uri } from "@polywrap/core-js";
+import { TracerConfig } from "@polywrap/tracing-js";
+import { IWrapperCache } from "@polywrap/uri-resolvers-js";
+
+export interface PolywrapCoreClientConfig<
```
krisbitney: I think this belongs with the core client, not with the config builder. What was the thought process behind the move?

packages/js/client/src/PolywrapClient.ts
```diff
@@ -1,36 +1,277 @@
...
+import {
+  compareSignature,
+  DeserializeManifestOptions,
+  ImportedModuleDefinition,
+  WrapManifest,
+} from "@polywrap/wrap-manifest-types-js";
+import { Tracer, TracerConfig } from "@polywrap/tracing-js";
 
-export class PolywrapClient extends PolywrapCoreClient {
+export class PolywrapClient<
```
krisbitney, line 35: Please add doc comments to the methods in PolywrapClient and update the readme snippets

packages/js/core-client/src/PolywrapCoreClient.ts
```diff
@@ -18,72 +16,35 @@ import {
...
   WrapErrorCode,
 } from "@polywrap/core-js";
 import { msgpackEncode, msgpackDecode } from "@polywrap/msgpack-js";
 import {
-  compareSignature,
   DeserializeManifestOptions,
   WrapManifest,
-  ImportedModuleDefinition,
 } from "@polywrap/wrap-manifest-types-js";
-import { Tracer, TracerConfig, TracingLevel } from "@polywrap/tracing-js";
```
krisbitney, line 33: While I think it might be correct to removing tracing from the core client and core, I think it should be a separate PR so it can be evaluated in isolation

krisbitney (dismissed):
There is a lot of cool stuff happening in this PR. Thanks for the amazing work!

There are five things happening in this PR:

1. add `BuilderConfig` and changes to `client-config-builder` UX
2. replace `Uri | string` with `Uri`
3. remove tracing from core
4. remove constants from default config
5. fix http server infra build

In the future, I think it would be better to try to do separate PRs. It's easier to review smaller PRs, and fewer issues will be missed by reviewers.

But again, there is a ton of value here. I'm excited to see this get merged!

packages/js/client-config-builder/src/bundles/getDefaultConfig.ts
krisbitney: The constants solve a real user problem. Before removing them, I'd like to see an alternative solution to the problem.

krisbitney:
If tracing is only implemented in `PolywrapClient`, how does that affect the usefullness of the tracing? If it's no longer effective, should we discuss removing tracing altogether?

packages/js/client-config-builder/src/bundles/getDefaultConfig.ts
pileks: Agreed, and there is a non-destructive path here, so I took it.

packages/js/client-config-builder/src/types/IClientConfigBuilder.ts
pileks: Agreed!

packages/js/client-config-builder/src/bundles/getDefaultConfig.ts
krisbitney: Awesome. Thanks!!

packages/js/client-config-builder/src/types/configs/PolywrapClientConfig.ts
krisbitney: To clarify, I'm not saying the decision is wrong at all. I don't know. I'm just wondering what the benefit of placing it here is, and if there is a cost.

packages/js/client-config-builder/src/types/configs/PolywrapCoreClientConfig.ts
krisbitney: Same as I said above: To clarify, I'm not saying the decision is wrong at all. I don't know.

packages/js/client/src/PolywrapClient.ts
krisbitney: I know this is a pain but I think you can just copy/paste them from the PolywrapCoreClient

packages/js/client-config-builder/src/types/configs/PolywrapClientConfig.ts
pileks: https://github.com/polywrap/toolchain/pull/1498

This PR tackles these issues, as I completely agree with what you're saying here.

`ClientConfigBuilder` should really only concern itself with outputting a single `CoreClientConfig`, and the Client should accept that as its primary source of config. The other config objects should only be handled by the appropriate client.

krisbitney: Super cool! I'm excited about these changes. You and Niraj did some amazing work.

packages/js/client-config-builder/src/types/IClientConfigBuilder.ts
```diff
@@ -1,43 +1,27 @@
...
-
-  // $start: IClientConfigBuilder-buildCoreConfig
   /**
    * Build a sanitized core client configuration that can be passed to the PolywrapClient or PolywrapCoreClient constructors
    *
-   * @returns CoreClientConfig<Uri> that results from applying all the steps in the builder pipeline
+   * @returns CoreClientConfig that results from applying all the steps in the builder pipeline
    */
-  buildCoreConfig(): CoreClientConfig<Uri>;
+  build(): CoreClientConfig;
```
nerfZael, line 15: `IClientConfigBuilder` seems to be missing a getter for config

packages/js/client-config-builder/src/BaseClientConfigBuilder.ts
```diff
@@ -1,54 +1,60 @@
...
+    redirects: {},
+    wrappers: {},
+    packages: {},
     resolvers: [],
   };
 
-  abstract addDefaults(): ClientConfigBuilder;
-  abstract buildCoreConfig(): CoreClientConfig<Uri>;
+  abstract addDefaults(): IClientConfigBuilder;
+  abstract build(
```
nerfZael: I don't think the `BaseClientConfigBuilder` should be aware of a wrapperCache

nerfZael (commented):
Most of this looks great.
I left two comments.
Additionally, I find the `Generic**` types really awkward

pileks:
> Most of this looks great. I left two comments. Additionally, I find the `Generic**` types really awkward

We will delete those in a future PR:
https://github.com/polywrap/toolchain/issues/1506

packages/js/client-config-builder/package.json
```diff
@@ -32,6 +32,7 @@
     "@polywrap/ipfs-plugin-js": "0.10.0-pre.7",
     "@polywrap/ipfs-resolver-plugin-js": "0.10.0-pre.7",
     "@polywrap/logger-plugin-js": "0.10.0",
+    "@polywrap/tracing-js": "0.10.0-pre.7",
```
cbrzn: this is not being used in this package

cbrzn (commented):
looking solid :clap: - i share the sentiment of Jure of having the two types but I think it's fine to fix this in another PR so we don't get things bigger. 

The only idea I would give is that, since we want to fix the `TUri` generic not going in through the entire client, we maybe have this generic only in the builder, and this takes care of sanitizing the URIs, the client would only expect an URI and wont know about any generics. Making the core types just expect `Uri` instead of `TUri` 

Just one minor thing to change (an unused dependency) and we're good to go

Niraj-Kamdar:
> If tracing is only implemented in `PolywrapClient`, how does that affect the usefullness of the tracing? If it's no longer effective, should we discuss removing tracing altogether?

It will still be beneficial to trace subinvocation traces, tryResolverUri, loadWrapper, or any other top-level functions.

cbrzn:
> > If tracing is only implemented in `PolywrapClient`, how does that affect the usefullness of the tracing? If it's no longer effective, should we discuss removing tracing altogether?
> 
> It will still be beneficial to trace subinvocation traces, tryResolverUri, loadWrapper, or any other top-level functions.

tbh i feel we're not really using the tracing - i think discussing to improve/remove it is worth it

packages/js/core/src/types/Uri.ts
```diff
@@ -194,4 +179,16 @@ export class Uri {
       throw Error(`Unknown uri type, cannot convert. ${JSON.stringify(uri)}`);
     }
   }
+
+  // $start: Uri-toString
+  /** @returns Uri string representation */
+  public toString(): string /* $ */ {
```
Niraj-Kamdar, line 185: Why do we need this?

Niraj-Kamdar (approved):
LGTM! Let's get this merged and then cleanup legacy

cbrzn (approved):

packages/js/core/src/types/Uri.ts
pileks: I'll assume that this would render the URI as a string within interpolated strings, but that's as far as I can fathom this.

TBH I've only now seen this and it's something that's been in the Uri class before, it's just been moved from lines 104-115 to here.

Niraj-Kamdar: Yeah I noticed that but not sure we need this functions in the first place

PR #1482, opened 2023-01-13
chore: separate deploy from polywrap project

### LINKED ISSUE #1426: Able To Run `polywrap deploy` w/o `polywrap.yaml` Manifest

cbrzn:
Creates a new `PolywrapDeploy` class which takes care of the deploy logic. Now user does not need to have `polywrap.yaml` to deploy a wrapper.

TODO:
- [x] Remove `deploy` attribute from `extensions` in `polywrap.yaml`

block: #1476 
closes #1426


cbrzn:
one open question: now that this pr removes the dependency of having a `polywrap.yaml` to do a deploy, would it make sense to remove [`deploy` from the `extensions` attribute of the `polywrap.yaml` manifest](https://github.com/polywrap/toolchain/blob/chore/separate-deploy-from-polywrap-project/packages/manifests/polywrap/formats/polywrap/0.3.0.json#L70-L74)?

pileks:
> one open question: now that this pr removes the dependency of having a `polywrap.yaml` to do a deploy, would it make sense to remove [`deploy` from the `extensions` attribute of the `polywrap.yaml` manifest](https://github.com/polywrap/toolchain/blob/chore/separate-deploy-from-polywrap-project/packages/manifests/polywrap/formats/polywrap/0.3.0.json#L70-L74)?

I think there's precedent for this (`polywrap.test.yaml`), and `deploy` is a very similar manifest to `test`, so I'm in 100% agreement with this.

cbrzn:
> > one open question: now that this pr removes the dependency of having a `polywrap.yaml` to do a deploy, would it make sense to remove [`deploy` from the `extensions` attribute of the `polywrap.yaml` manifest](https://github.com/polywrap/toolchain/blob/chore/separate-deploy-from-polywrap-project/packages/manifests/polywrap/formats/polywrap/0.3.0.json#L70-L74)?
> 
> I think there's precedent for this (`polywrap.test.yaml`), and `deploy` is a very similar manifest to `test`, so I'm in 100% agreement with this.

@pileks perfect, working on this now

packages/cli/src/lib/deploy/PolywrapDeploy.ts
```diff
@@ -0,0 +1,96 @@
...
+import nodePath from "path";
+
+interface PolywrapDeployConfig {
+  cache: CacheDirectory;
+  logger: Logger;
+  defaultModulesCached: boolean;
+}
+
+export class PolywrapDeploy {
+  public static cacheLayout = {
```
pileks: Any particular reason for `public static` instead of just a file-level `const` here?

packages/cli/src/lib/deploy/PolywrapDeploy.ts
```diff
@@ -0,0 +1,96 @@
...
+import path from "path";
+import nodePath from "path";
+
+interface PolywrapDeployConfig {
+  cache: CacheDirectory;
+  logger: Logger;
+  defaultModulesCached: boolean;
+}
+
+export class PolywrapDeploy {
```
pileks: We usually have `lib` classes like `Compiler`, `SchemaComposer`, etc. that the commands call.

Would it make sense to name this `ProjectDeployer` or simply `Deployer`?

I know `Deployer` already exists, although it does seem like that one should be called `DeployModule` rather than `Deployer`.

A small nitpick, just for clarity's sake 😄

packages/cli/src/lib/deploy/PolywrapDeploy.ts
```diff
@@ -0,0 +1,96 @@
...
+  ): Promise<PolywrapDeploy> {
+    const deployManifest = await loadDeployManifest(manifest, logger);
+    const cache = new CacheDirectory({
+      rootDir: nodePath.dirname(manifest),
+      subDir: PolywrapDeploy.cacheLayout.root,
+    });
+    return new PolywrapDeploy(deployManifest, cache, logger);
+  }
+
+  public async getDeployModule(
```
pileks: Related to my naming comment. We're literally returning a `Deployer` (a.k.a. DeployModule) here.

While thinking about this, we can also just open an issue for this and do it separately as to not overly increase the scope of this PR.

packages/js/manifests/polywrap/src/formats/polywrap/migrators/0.3.0_to_0.4.0.ts
```diff
@@ -0,0 +1,15 @@
+import { ILogger } from "@polywrap/logging-js";
+import { PolywrapManifest as OldManifest } from "../0.3.0";
+import { PolywrapManifest as NewManifest } from "../0.4.0";
+
+export function migrate(migrate: OldManifest, logger?: ILogger): NewManifest {
+  if (migrate.extensions?.deploy) {
+    logger?.warn(
+      `The Polywrap Deploy Manifest does not needs to be defined in the Polywrap Manifest, please remove it`
```
pileks: I think it would make more sense to just
```typescript
delete migrate.extensions.deploy;
```
instead of warning the user? Or you can do both!

If you still want to keep the warning text, there's an English grammar error in your sentence:
`needs` -> `need`

An alternative sentence too:
```
The Polywrap Deploy Manifest ('extensions.deploy') is no longer used within the Polywrap Manifest. Please remove it from your manifest file.
```

pileks (changes requested):
Overall this is a great change and a step forward in a direction we've already taken with `build` and `codegen`!

One thing I'd note - When migrating the templates, it would be a good excercise to migrate them using `polywrap manifest migrate`, so that you make sure that you are creating a migration that can truly migrate our manifests (of course, there can always be breaking changes which prevent you from this).

Apart from some nitpicks, I'd say that I'd personally go with the `delete` approach in the migration.

packages/cli/src/lib/deploy/PolywrapDeploy.ts
cbrzn: not really, just following the same pattern as `PolywrapProject` - I also think that makes sense since these are constant- maybe adding a `readonly` would be enough?

packages/cli/src/lib/deploy/PolywrapDeploy.ts
cbrzn: hey i opened the changes in other pr on top of this https://github.com/polywrap/toolchain/pull/1483

packages/cli/src/lib/deploy/PolywrapDeploy.ts
cbrzn: will resolve and conversation will follow in the other comment related to this

packages/js/manifests/polywrap/src/formats/polywrap/migrators/0.3.0_to_0.4.0.ts
cbrzn: just pushed changes based on your feedback, please feel free to review and resolve if you feel it's enough :smile:

packages/cli/src/lib/deploy/PolywrapDeploy.ts
pileks: yeah, that makes sense!

pileks (dismissed):
Looking great!

dOrgJelli:
Hey @cbrzn @pileks, some thoughts on how I think we should treat the project manifest (`polywrap.yaml`) and what we currently call "extensions" manifests (build, infra, test, deploy, etc).

First thought is, I think we should remove the concept of a project "extension" and create a much simpler & uniform experience for the CLI. Instead I think we should call them **command manifests**, which can be defined for the various CLI commands: `polywrap.build.yaml`, `polywrap.infra.yaml`, `polywrap.test.yaml`, `polywrap.deploy.yaml`.

Second, we should have the command manifests back-link to the project they are referencing (if needed). Currently only the `polywrap.build.yaml` manifest needs to reference a project that is to be built. It should do this through a property within itself, for example `project: ./polywrap.yaml`.

Lastly, commands will have default naming conventions for their manifests (i.e. `polywrap.build.yaml`, `polywrap.deploy.yaml`, etc).

The result of doing these things is that we:
- Create a much simpler experience that can be explained easily.
- Remove the unnecessary concept of "project extensions", which doesn't apply well to all manifest types (as we see here with deploy).

Thoughts?

packages/js/manifests/polywrap/src/formats/polywrap/index.ts
```diff
@@ -35,18 +40,20 @@ export const PolywrapManifestSchemaFiles: Record<string, string> = {
   "0.1.0": "formats/polywrap/0.1.0.json",
   "0.2.0": "formats/polywrap/0.2.0.json",
   "0.3.0": "formats/polywrap/0.3.0.json",
+  "0.4.0": "formats/polywrap/0.4.0.json",
```
dOrgJelli: Hey @cbrzn, check the CHANGELOG.md file, we've actually added version `0.3.0` within toolchain version `0.10.0`, so we don't need to increment this to `0.4.0` since `0.3.0` is still in a pre-release state.

Here's the entry from the CHANGELOG for your reference:
image

dOrgJelli (changes requested):

packages/js/manifests/polywrap/src/formats/polywrap/index.ts
dOrgJelli: FYI I'm working on making this change, and the removal of extensions as described above in another branch, will PR into this branch soon. Will report back soon.

dOrgJelli:
Okay, spent some time digging into this https://github.com/polywrap/toolchain/pull/1482#issuecomment-1386325805 and have a bit more of an appreciation for the concept of a "project exetension". I think project extensions make sense for things like the build & codegen command, since they both require a project manifest (`polywrap.yaml`), and can **optionally** have additional configuration options set within another manifest (`polywrap.build.yaml`), which is the extension manifest.

So, this means that the following will be true:
- Project commands like `polywrap build` and `polywrap codegen` will require a `polywrap.yaml` project manifest, and will look for an optional extension manifest for added configuration options. For example: `polywrap.build.yaml`, `polywrap.codegen.yaml`.
- Stand-alone commands like `polywrap infra`, `polywrap test`, and `polywrap deploy` will not require a `polywrap.yaml` manifest be present, and will instead look for their command manifests. For example: `polywrap.infra.yaml`, `polywrap.test.yaml`, `polywrap.deploy.yaml`.

I know this is right back where we started, but just wanted to fully explore the options here and explain the rationale.

dOrgJelli:
Here's the PR @cbrzn: https://github.com/polywrap/toolchain/pull/1486

cbrzn:
> I know this is right back where we started, but just wanted to fully explore the options here and explain the rationale.

@dOrgJelli what you said makes sense, and I also thought that we should only have `build` as an extension in `polywrap.yaml`. Thanks for explaining the thing about the version, nice catch there. 

I will merge and fix the tests

packages/cli/src/lib/deploy/Deployer.ts
```diff
@@ -0,0 +1,229 @@
...
+    });
+
+    if (unrecognizedPackages.length) {
+      throw new Error(
+        `Unrecognized packages: ${unrecognizedPackages.join(", ")}`
+      );
+    }
+  }
+
+  private async _getDeployModule(
```
Niraj-Kamdar, line 146: rename: `getDeployPackage`

packages/cli/src/lib/deploy/Deployer.ts
```diff
@@ -0,0 +1,229 @@
...
+    );
+
+    return {
+      // eslint-disable-next-line @typescript-eslint/no-require-imports
+      deployModule: require(cachePath).default as DeployModule,
+      manifestExt,
+    };
+  }
+
+  private async _cacheDeployModules(modules: string[]): Promise<void> {
```
Niraj-Kamdar, line 174: cacheDeployPackages

Niraj-Kamdar (changes requested):

packages/cli/src/lib/deploy/Deployer.ts
dOrgJelli: Private properties & methods should be prefixed with an underscore. We do this in other places within the toolchain:
https://github.com/polywrap/toolchain/search?q=private+_

packages/cli/src/lib/deploy/Deployer.ts
dOrgJelli: Same comment as above.

dOrgJelli (dismissed):

packages/cli/src/lib/deploy/Deployer.ts
Niraj-Kamdar: It's not about `_` I meant renaming module -> package 😅

packages/cli/src/lib/deploy/Deployer.ts
Niraj-Kamdar: Same here

PR #1483, opened 2023-01-17
cli: deployer rename

cbrzn:
addresses  https://github.com/polywrap/toolchain/pull/1482#discussion_r1071582421 & https://github.com/polywrap/toolchain/pull/1482#discussion_r1071577207

### DIFF
--- a/packages/cli/src/lib/deploy/Deployer.ts
+++ b/packages/cli/src/lib/deploy/Deployer.ts
@@ -1,28 +1,32 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 
-import { Deployer } from "./deployer";
 import { loadDeployManifest, loadDeployManifestExt } from "../project";
 import { CacheDirectory } from "../CacheDirectory";
 import { Logger } from "../logging";
 
+import { Uri } from "@polywrap/core-js";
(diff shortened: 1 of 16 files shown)


packages/cli/src/lib/deploy/Deployer.ts
```diff
@@ -1,28 +1,32 @@
...
 import { Logger } from "../logging";
 
+import { Uri } from "@polywrap/core-js";
 import { DeployManifest } from "@polywrap/polywrap-manifest-types-js";
 import { Schema as JsonSchema } from "jsonschema";
 import path from "path";
 import nodePath from "path";
 
-interface PolywrapDeployConfig {
+export interface DeployModule {
```
pileks: One nitpick - would it make sense to export this `DeployModule` into its own file?

It usually makes searching through the projects/files trivial, and it's the general approach I take when refactoring/coding new stuff.

pileks (approved):
LGTM!

I did have one nitpick but it's not blocking anything.

packages/cli/src/lib/deploy/Deployer.ts
dOrgJelli: Agreed, going to push an update to make this the case within the `lib/deploy/` directory.

dOrgJelli (approved):

PR #1484, opened 2023-01-17
`ClientConfigBuilder` - move wrapperCache and resolver from constructor to buildCoreConfig

pileks:
This PR aims to move ClientConfigBuilder constructor arguments into the only place where they are actually used - the `buildCoreConfig` method.

Context from Discord:

@krisbitney mentioned:
I noticed the ClientConfigBuilder constructor accepts a IWrapperCache argument, but the cache is ignored when calling build(). It's only used for buildCoreConfig(). Could this be misleading?

I'm thinking someone could call build() and then not realize they need to put their wrapper cache in a PolywrapClientConfig.

```typescript
// config in constructor will be lost
const clientConfig = new ClientConfigBuilder(new CustomWrapperCache())
  .add(...)
  .build();

// my client doesn't use my custom cache!
const wrongClient = new PolywrapClient(clientConfig);

// Ohhh, it's because I forgot this step
const polywrapClientConfig = {
  ...clientConfig
   wrapperCach: new CustomWrapperCache(),
}

// this client will use the custom cache
const rightClient = new PolywrapClient(polywrapClientConfig);
```

After discussion between @pileks and @nerfZael it was decided that this is likely the best approach.

### DIFF
--- a/packages/js/client-config-builder/README.md
+++ b/packages/js/client-config-builder/README.md
@@ -13,12 +13,6 @@ Initialize a ClientConfigBuilder using the [constructor](#constructor)
 ```typescript
   // start with a blank slate (typical usage)
   const builder = new ClientConfigBuilder();
-
-  // instantiate a builder with a custom cache and/or resolver
-  const _builder = new ClientConfigBuilder(
-    new WrapperCache(),
-    RecursiveResolver.from([])
-  );
 ```
 
 ### Configure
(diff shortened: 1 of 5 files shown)


krisbitney (approved):
LGTM! Thanks for doing this!

Niraj-Kamdar (approved):

PR #1487, opened 2023-01-18
fix: handle null return values from uri resolver extensions

dOrgJelli:
@krisbitney this actually helps debug the issue where the ENS text records were not resolving correctly. I'm working on getting that fixed right now, and came across this, as it was masking the problem occurring a few invocation layers deep.

### DIFF
--- a/packages/js/core/src/interfaces/uri-resolver.ts
+++ b/packages/js/core/src/interfaces/uri-resolver.ts
@@ -4,8 +4,8 @@ import { Tracer } from "@polywrap/tracing-js";
 import { Result } from "@polywrap/result";
 
 export interface MaybeUriOrManifest {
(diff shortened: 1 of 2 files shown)


krisbitney (approved):

krisbitney:
> @krisbitney this actually helps debug the issue where the ENS text records were not resolving correctly. I'm working on getting that fixed right now, and came across this, as it was masking the problem occurring a few invocation layers deep.

Thanks! This is a much-needed fix. I actually made a similar change in the PR that adds the ethereum wrapper.

I think the text records issue I experienced was caused by me trying to resolve the ethereum wrapper from an ENS text record ENS resolution depends on the ethereum wrapper, so it was never going to work.

PR #1493, opened 2023-01-24
Update PackageToWrapperCacheResolver to prevent unnecessary IO

krisbitney:
Currently, many asynchronous invocations will trigger many IO requests. For example, in the Uniswap V3 front end demo dApp, the Uniswap v3 Wasm wrapper is downloaded several times whenever the page is refreshed. This occurs because multiple asynchronous calls to `invoke` are triggered before the wrapper is cached.

This update to the `PackageToWrapperCacheResolver` uses mutexes to synchronize URI resolution on a per-URI basis until the first time the URI is resolved. Once a URI is resolved for the first time, its mutex is destroyed and parallel calls are allowed for that URI.


packages/js/client-config-builder/src/ClientConfigBuilder.ts
```diff
@@ -4,11 +4,11 @@ import { BaseClientConfigBuilder } from "./BaseClientConfigBuilder";
 import { CoreClientConfig, Uri, IUriResolver } from "@polywrap/core-js";
 import {
   IWrapperCache,
-  PackageToWrapperCacheResolver,
   RecursiveResolver,
   StaticResolver,
   WrapperCache,
 } from "@polywrap/uri-resolvers-js";
+import { EcoCacheResolver } from "wraplib";
```
Niraj-Kamdar: Why don't we move the resolver to uri-resolver-js?

Niraj-Kamdar: Can we also do the same for the other utility functions and move it to its relevant package in toolchain?

Niraj-Kamdar (changes requested):

Niraj-Kamdar:
This is really awesome improvement!

packages/js/client-config-builder/src/ClientConfigBuilder.ts
krisbitney: I think `uri-resolvers-js` is for core/fundamental resolvers. I imagine in the future there will be many resolvers developed by a large Polywrap community. If it makes sense to put the `EcoCacheResolver` in `uri-resolvers-js`, I'm happy to do it.

I created the wraplib package for my own exploration and enjoyment. I had some ideas for things I wanted to try out. It was motivating when I was exhausted. 

I can move the wraplib repo to the Polywrap organization so long as I can continue to freely develop in it as I please, but I'm not going to move everything out of it and delete it.

krisbitney: I added `EcoCacheResolver` to `uri-resolvers-js` and removed `wraplib` as a dependency. Instead of having two similar resolvers, would it make sense to instead update `PackageToWrapperCacheResolver` with the new features and refactor? Or should they remain distinct resolvers?
 
I would really like your input on this @nerfZael

Niraj-Kamdar: I thought we are going to improve `PackageToWrapperCacheResolver` with `EcoCacheResolver`, Isn't that the case?

krisbitney: I didn't know that's what you were requesting.

I pushed a commit that removes `EcoCacheResolver` completely and instead modifies `PackageToWrapperCacheResolver`.

dOrgJelli (approved):
This is really well done!

packages/js/client-config-builder/src/ClientConfigBuilder.ts
nerfZael: I don't think this should be part of PackageToWrapperCacheResolver.
It's better to separate it into its own resolver, so that it can be used with any other resolver (just like PackageToWrapperCacheResolver or RecursiveResolver can)

krisbitney: It can be a distinct resolver, but it would be a drop-in alternative for the `PackageToWrapperCacheResolver`. Is that okay?

It needs to be able to check if a URI is cached. So I think it either has to be
- integrated with the `PackageToWrapperCacheResolver`
- take a `PackageToWrapperCacheResolver` as the resolver argument in its constructor/factory method,
- be a distinct resolver that has the same cache logic as the `PackageToWrapperCacheResolver`

What do you think?

nerfZael: I'm thinking something along the lines of a distinct resolver that has it's own cache and takes any IUriResolver as argument (In the default config that would be the PackageToWrapperCacheResolver).
The cache could be Uri -> UriResolutionResult (Result<UriPackageOrWrapper, Error>)

nerfZael: Actually, in the default config, PackageToWrapperCacheResolver could take this new resolver as argument. That might be better

nerfZael: Not a fan of EcoResolver name though, maybe call it RequestSynchronizationResolver? I dunno :D

krisbitney: > I'm thinking something along the lines of a distinct resolver that has it's own cache and takes any IUriResolver as argument (In the default config that would be the PackageToWrapperCacheResolver). The cache could be Uri -> UriResolutionResult (Result<UriPackageOrWrapper, Error>)

This is an interesting idea. I'll work on this.

krisbitney: I don't think it makes

> Actually, in the default config, PackageToWrapperCacheResolver could take this new resolver as argument. That might be better

The idea is that the lock is in place until the wrapper is cached, so I think it makes more sense for it to take the `PackageToWrapperCacheResolver` as an argument.

krisbitney: > I'm thinking something along the lines of a distinct resolver that has it's own cache and takes any IUriResolver as argument (In the default config that would be the PackageToWrapperCacheResolver). The cache could be Uri -> UriResolutionResult (Result<UriPackageOrWrapper, Error>)

I'm implementing this and I am a little worried that having a separate cache will effectively double the amount of data cached and eat up memory. Do you think that's an issue?

nerfZael: No, but I also messed up when talking about the type of cache.
It should not be Uri -> Result<UriPackageOrWrapper, Error>
It should be Uri -> Promise<Result<UriPackageOrWrapper, Error>>
And ofc, the items are temporary and should be deleted after promise resolution.

krisbitney: Okay, gotcha

krisbitney: I thought I understood what you were saying but now I don't think I do.

Let's say there are N invocations waiting to obtain the mutex and 1 resolving. The resolving invocation finishes resolution and returns a `Result<UriPackageOrWrapper, TError>`. The lock is destroyed. The remaining 9 need to return the `Result<UriPackageOrWrapper, TError>`. They should not be forwarded to the next resolver.

How do I know when the `Result<UriPackageOrWrapper, TError>` is safe to delete?

In other words, how do I know that the waiting "threads" have finished returning the `Result` and the wrapper has been successfully cached by the `PackageToWrapperCacheResolver`?

I don't think storing a `Promise` is helpful because we need to see if the result is an error before we can destroy the lock.

krisbitney: I could set a timeout to delete the `Result<UriPackageOrWrapper, TError>` but that isn't reliable.

Also, it really makes more sense to store the `UriPackageOrWrapper`. If a result is an error then we can't cache anything and we don't want to destroy the mutex yet.

packages/js/uri-resolvers/src/__tests__/package-to-wrapper-cache-resolver.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+      ]
+    )
+    .buildCoreConfig();
+
+  return new PolywrapClient(config, { noDefaults: true });
+};
+
+describe("PackageToWrapperCacheResolver", () => {
+
+  it("repeated calls with same uri trigger only one network request", async () => {
```
nerfZael, line 70: If we're using Mutexes, I think we should have tests that use workers to confirm it all works as expected.
Otherwise we could use Mutex-less request caching

PR #1494, opened 2023-01-26
Move http interface and plugin to its own repo

krisbitney:
As part of an effort to move all the interfaces and plugins outside of the toolchain repo, I moved the http plugin and interface to their own repo here: https://github.com/polywrap/http

There is a temporary extra redirect to avoid breaking changes with wrappers that use the old package URI:
```ts
{
  from: new Uri("wrap://ens/http.polywrap.eth"),
  to: new Uri(defaultInterfaces.http),
},
```


Niraj-Kamdar (approved):
LGTM!

cbrzn (approved):

PR #1495, opened 2023-01-26
move fs interface and plugin to new repo

krisbitney:
As part of an effort to move all the interfaces and plugins outside of the toolchain repo, I moved the file-system plugin and interface to their own repo here: https://github.com/polywrap/file-system

There is a temporary import redirect added to the file-system resolver plugin. It's necessary due to circular dependencies with the CLI being used to build the package. The import redirect can be removed once the CLI in the dependencies folder incorporates the change.

There is also a temporary extra redirect added to avoid breaking changes:
```ts
{
  from: new Uri("wrap://ens/fs.polywrap.eth"),
  to: new Uri(defaultInterfaces.fileSystem),
},
```


Niraj-Kamdar (dismissed):
LGTM!

cbrzn (dismissed):

cbrzn (approved):

dOrgJelli (approved):
@krisbitney can we make a quick issue for the remaining work of removing these workarounds so we don't forget? Also should we create a new pre-release ASAP so we can do this?

krisbitney:
> @krisbitney can we make a quick issue for the remaining work of removing these workarounds so we don't forget? Also should we create a new pre-release ASAP so we can do this?

I created an issue here: https://github.com/polywrap/toolchain/issues/1504

I don't think we necessarily need a pre-release ASAP. I don't see it as urgent. But if it makes sense to do a release soon anyway, I'm for it.

PR #1496, opened 2023-01-27
update concurrent plugin interface uri

krisbitney:
This PR updates the concurrent plugin to use an ENS text record URI


cbrzn (approved):
cool! should we also update the wrappers that are in goerli to a text record in mainnet? following the same pattern as interfaces?

krisbitney:
> cool! should we also update the wrappers that are in goerli to a text record in mainnet? following the same pattern as interfaces?

Yeah, for sure. I'll do it in another PR.

dOrgJelli (approved):

PR #1505, opened 2023-02-01
Make wrapper manifest getter immutable

### LINKED ISSUE #1442: A 2nd instance of SchemaComposer fails to generate a valid ABI
**Describe the bug**
For a reproduction, please check out `pileks/reproduction/schema-composer-second-call-issue` and see code in `packages/cli/commands/codegen.ts`.

1. `yarn reset` and make sure everything's built.
2. Create a new wrapper (wasm-as is fine) using `polywrap create wasm assemblyscript test-wrapper`
3. Edit the wrapper's `schema.graphql` to include multiple imports, some of which point to a plugin:
   ```graphql
   #import { Module } into Fs from "wrap://ens/fs.polywrap.eth"
   #import { Module } into Pwr from "wrap://ens/pwr-app.eth"
   #import { Module } into Eth from "wrap://ens/ethereum.polywrap.eth"
   #import { Module } into HelloWorld from "wrap://ipfs/QmebzauKAXoYbywLAYdBvKyPhWsDoHfkUC4wffWvjighKT"
   ```
   A minimal reproduction can be achieved using:
   ```graphql
   #import { Module } into Fs from "wrap://ens/fs.polywrap.eth"
   ```
4. Run `polywrap codegen`
5. The usual `codegen` run executes smoothly.
6. The code within `generateProjectCache` fails during `schemaComposer.getComposedAbis()` with the following error:
   ```bash
   Error: Unknown property type found: type Fs_Module { readFileAsString: Fs_FileSystem_Encoding }
       at cleanup (/home/pileks/dev/polywrap/toolchain/packages/schema/parse/build/validate/types.js:159:31)
       at validate (/home/pileks/dev/polywrap/toolchain/packages/schema/parse/build/index.js:95:17)
       at Object.parseSchema (/home/pileks/dev/polywrap/toolchain/packages/schema/parse/build/index.js:43:9)
       at Object.<anonymous> (/home/pileks/dev/polywrap/toolchain/packages/schema/compose/build/resolve.js:210:58)
       at step (/home/pileks/dev/polywrap/toolchain/packages/schema/compose/build/resolve.js:47:23)
       at Object.next (/home/pileks/dev/polywrap/toolchain/packages/schema/compose/build/resolve.js:28:53)
       at fulfilled (/home/pileks/dev/polywrap/toolchain/packages/schema/compose/build/resolve.js:19:58)
       at processTicksAndRejections (node:internal/process/task_queues:96:5)
   ```

**Expected behavior**
The 2nd instance of the SchemaComposer should run `getComposedAbis()` smoothly, just like the 1st one.

- OS: Ubuntu 20.04.4 LTS running in WSL
- Polywrap CLI version `0.10.0-pre.5`

**Additional context**
I have managed to trace the problem down to somewhere within the `resolveImportsAndParseSchemas` function within `packages/schema/compose/src/resolve.ts`, Line 130 onwards.
It seems that on the 2nd run, `Fs_FileSystem_Encoding` doesn't get imported properly, and this is visible when we try to inspect the `newSchema` string found within this function on the 1st and 2nd runs.

1st run:

```graphql
# ...
type Module @imports(
    types: [
      \"Fs_Module\",
    \"Fs_FileSystem_Encoding\",
    \"Pwr_Module\",
    \"Eth_Module\",
    \"Eth_Connection\",
    \"Eth_TxOverrides\",
    \"Eth_StaticTxResult\",
    \"Eth_TxRequest\",
    \"Eth_TxReceipt\",
    \"Eth_Log\",
    \"Eth_EventNotification\",
    \"Eth_Network\",
    \"Eth_TxResponse\",
    \"Eth_Access\",
    \"HelloWorld_Module\"
    ]
    ) {
    sampleMethod(arg: String!): SampleResult!
}
# ...
```

2nd run (notice the missing `Fs_FileSystem_Encoding` in @imports)
```graphql
# ...
type Module @imports(
    types: [
      \"Fs_Module\",
    \"Pwr_Module\",
    \"Eth_Module\",
    \"Eth_Connection\",
    \"Eth_TxOverrides\",
    \"Eth_StaticTxResult\",
    \"Eth_TxRequest\",
    \"Eth_TxReceipt\",
    \"Eth_Log\",
    \"Eth_EventNotification\",
    \"Eth_Network\",
    \"Eth_TxResponse\",
    \"Eth_Access\",
    \"HelloWorld_Module\"
    ]
    ) {
    sampleMethod(arg: String!): SampleResult!
}
# ...
```

namesty:
NOTICE - this PR has been migrated here: https://github.com/polywrap/javascript-client/pull/4

Closes #1442

This PR aims to make the resulting manifest from `getManifest` getters in Wrappers and Packages not mutate the internal state of the Wrapper.

In what we had before, the manifest returned by `getManifest` is a reference to the wrapper's manifest private property. So, mutating it, modified the internal state of the wrapper. And that was the reason #1442 was happening:

- A wrapper was resolved and its manifest was mutated
- It was cached
- In the 2nd run, the same instance is retrieved from the cache
- The codegen process threw a runtime error because of the previous manifest mutation

The solution here would be making all manifest getters return a `Readonly` manifest. However, in Typescript, `Readonly` doesn't work as you would expect, take a look here:

- https://www.banterly.net/2021/08/20/the-readonly-lie-of-typescript/
- https://stackoverflow.com/questions/68056810/why-is-the-readonly-property-in-typescript-not-working

So, in the actual wrapper getters, I am returning a deep clone to ensure that the returned object cannot modify the wrapper's internals in an uncontrolled way. This is, unless we may wanna use a new compiler flag like discussed here: https://github.com/microsoft/TypeScript/issues/13002


packages/js/plugin/src/PluginPackage.ts
```diff
@@ -50,8 +50,8 @@ export class PluginPackage<
     }
   }
 
-  async getManifest(): Promise<Result<WrapManifest, Error>> {
-    return ResultOk(this._manifest);
+  async getManifest(): Promise<Result<Readonly<WrapManifest>, Error>> {
+    return ResultOk(JSON.parse(JSON.stringify(this._manifest)));
```
krisbitney: There might be a faster way to deep clone the manifest. Here are some benchmarks: http://jsben.ch/bWfk9

I found the benchmarks in this StackOverflow post: https://stackoverflow.com/questions/122102/what-is-the-most-efficient-way-to-deep-clone-an-object-in-javascript/5344074#5344074

krisbitney: I don't know if there is a node equivalent, but the JS Web API also has [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone).

namesty: I agree with you. Found this snippet which seems to be extremely efficient (see `c7x43t`'s `deepcopy`): https://jsben.ch/gtKdJ

Adapted it slightly for Typescript and used that instead of JSON.stringify + parse.

I benchmarked `structuredClone` and it seems to be even less efficient than Stringify + Parse

krisbitney: Amazing!

krisbitney (approved):
Fantastic work!

packages/js/plugin/src/PluginPackage.ts
cbrzn: while i agree that we should think of performance, I am not sure at which point we should stop thinking about it. My main thought is: Is the deep clone's speed really impactful when building or invoking a wrapper?

my concern is that here we are adding custom code which we're committing to maintaining; I think that we might appreciate in the future going with the approach of `structuredClone` or `JSON.parse(JSON.stringify(..))` because it is **way** much easier to debug/maintain. In cases like this, we will need to decide whether we want performance or maintainability, and in this scenario, I would go with the latter.

These are my two cents hehe no strong opinion on this. (My only strong opinion is that if we want to add that `deepCopy` function, we should add tests to it :P)

krisbitney: That's a good point. There is a tradeoff here. How often is `getManifest` called?

krisbitney: I thought about it more and you're right. I think it makes sense to use the simpler code. Sorry for complicating this!

nerfZael:
I'm not sure about the use of `deepCopy` here.
While `Readonly` is not perfect, it does communicate the intent to other developers and saves on performance.
`deepCopy` on the other hand makes it safer against misuse by unaware devs in certain cases, while sacrificing performance.
It's true that we (the PolywrapClient) does not use the `getManifest` much (or at all) right now, but that doesn't mean it won't be used by the user, maybe even on every invocation.

dOrgJelli:
Migrating this PR to the javascript-client repository here: https://github.com/polywrap/javascript-client/pull/4

PR #1518, opened 2023-02-03
js/client-config-builder: build method now support one argument

### LINKED ISSUE #1507: `clientConfigBuilder.build(wrapperCache, resolver)` should accept a Resolver OR WrapperCache, not both
**Describe the bug**
As the title describes, the `clientConfigBuilder.build(wrapperCache, resolver)` method should accept a Resolver OR WrapperCache, not both.
The reason for this is the method is opaque to outside developers, and it's unclear what it does with these two arguments. 
Essentially, the `build` method will only use the `wrapperCache` parameter if a `resolver` parameter is not provided, otherwise it will use the `resolver` parameter or build a completely new Resolver for the `CoreClientConfig`.

**Expected behavior**
This, in my view, is a bug, and can be easily mitigated by having the `build` method accept a single argument of type:
```ts
type BuildOptions = { resolver: IUriResolver<unknown> } | { wrapperCache: IWrapperCache };
```

This would make sure that developers are explicit in their intent, and that the `build` method behaves in accordance with the developer's intent.
Having a `BuildOptions` type instead of something simple like an `IUriResolver | IWrapperCache` is a preference, but it will allow us to add additional options in the future (maybe even for both cases) while not breaking the method signature.

**Additional context**
Found while working on the #1480 refactor.

cbrzn:
closes https://github.com/polywrap/toolchain/issues/1507

### DIFF
--- a/packages/js/client-config-builder/README.md
+++ b/packages/js/client-config-builder/README.md
@@ -64,7 +64,6 @@ Finally, build a ClientConfig or CoreClientConfig to pass to the PolywrapClient 
   // or build with a custom resolver
   coreClientConfig = builder.build({
     resolver: RecursiveResolver.from([]),
-    wrapperCache: new WrapperCache(),
   });
 ```
 
--- a/packages/js/client-config-builder/README.md
+++ b/packages/js/client-config-builder/README.md
@@ -56,11 +56,16 @@ Finally, build a ClientConfig or CoreClientConfig to pass to the PolywrapClient 
   // accepted by either the PolywrapClient or the PolywrapCoreClient
   let coreClientConfig = builder.build();
 
-  // build with a custom cache and/or resolver
-  coreClientConfig = builder.build(
-    new WrapperCache(),
-    RecursiveResolver.from([])
-  );
+  // build with a custom cache
+  coreClientConfig = builder.build({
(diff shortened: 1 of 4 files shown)


dOrgJelli (dismissed):
LGTM, would defer to @pileks and @nerfZael though.

packages/js/client-config-builder/README.md
```diff
@@ -56,11 +56,16 @@ Finally, build a ClientConfig or CoreClientConfig to pass to the PolywrapClient
...
-  );
+  // build with a custom cache
+  coreClientConfig = builder.build({
+    wrapperCache: new WrapperCache(),
+  });
+
+  // or build with a custom resolver
+  coreClientConfig = builder.build({
+    resolver: RecursiveResolver.from([]),
+    wrapperCache: new WrapperCache(),
```
pileks: It seems like you missed rebuilding the readme once here, looking at quickstart.ts 😄

pileks (changes requested):
LGTM other than an apparent error in the README, so I'm requesting changes 🙂

packages/js/client-config-builder/README.md
cbrzn: lol nice catch

dOrgJelli (approved):

PR #1525, opened 2023-02-07
cli/create: remove node & react from supported langs in app template

cbrzn:
in a previous PR the `typescript-node` and `typescript-react` templates have been removed from `app` (https://github.com/polywrap/toolchain/pull/1500); this aims to update the create command with these changes.

the ci did not catch this because we're directly downloading the latest version from npm (0.9.4) when running the create command. this makes me think that decoupling templates from toolchain might be a necessary thing to do sooner than later (so these things don't go to production without us noticing)

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -176,8 +176,8 @@ polywrap create wasm assemblyscript my-wrapper
 # Create an interface project using assemblyscript called "my-project"
 polywrap create wasm interface my-interface
 
-# Create a React app project using Typescript called "my-react-app"
-polywrap create app typescript-react my-react-app
+# Create a project using Typescript called "my-react-app"
+polywrap create app typescript my-app
 
 # Create a Plugin wrapper project using Typescript called "my-plugin"
 polywrap create plugin typescript my-plugin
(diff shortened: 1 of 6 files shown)


krisbitney (changes requested):
This is a great PR. Thanks looking at this.

There seems to be a bug where the full template isn't generated. I guess it's because Polywrap version 0.9.4 of the templates package doesn't have a template called "typescript"?

Screenshot 2023-02-08 at 12 27 04 PM

krisbitney:
Maybe the PR is fine as is and we should merge it assuming it will work when 0.10.0 is released. We can do a patch fix if there are problems. What do you think?

cbrzn:
> Maybe the PR is fine as is and we should merge it assuming it will work when 0.10.0 is released. We can do a patch fix if there are problems. What do you think?

@krisbitney yes, that's the exact reason - actually, it amazes me that the CI went green (I was expecting it to fail lol). but yah, I feel comfortable with merging this so we're ready for 0.10 release

krisbitney (approved):

dOrgJelli (approved):

PR #1526, opened 2023-02-07
`InterfaceRedirectResolver` - an `IUriResolver` that treats interface implementations as implied redirects

### LINKED ISSUE #1501: IUriResolver that treats interface implementations as implied redirects
An implementation of the interface is implicitly invocable at the interface URI, but not directly.

An `IUriResolver` implementation that can redirect invocations to an interface to the first implementation registered to it. In the case that there are multiple implementations available, it could be nice to be able to change how implementations are selected for invocation.

```ts
const resolver = new InterfaceRedirectsResolver();
                    .addDefaults()
                    .addResolver(resolver)
                    .buildCoreConfig();
```

```ts
const coreConfig = {
      envs: this.config.envs,
      interfaces: [{
        interface: new Uri(defaultInterfaces.uriResolver),
        implementations: [new Uri(defaultPackages.httpResolver)],
      }],
      resolver: RecursiveResolver.from(
          PackageToWrapperCacheResolver.from(
            [
              {
                uri: defaultPackages.httpResolver,
                package: httpResolverPlugin({}),
              },
              new ExtendableUriResolver(),
              new InterfaceRedirectsResolver(), // now I can invoke the uri resolver interface directly
            ],
            new WrapperCache(),
          )
        ),
    };
```

**Describe alternatives you've considered**
TODO

**Additional context**
TODO

krisbitney:
The `InterfaceRedirectResolver` is a simple `IUriResolver` implementation that redirects from a registered interface URI to the first registered implementation for that interface.

This PR also fixes a bug in https://github.com/polywrap/toolchain/pull/1518 where the `build()` method in the `IClientConfigBuilder` interface did not accept the new options as an argument.

Closes https://github.com/polywrap/toolchain/issues/1501

### DIFF
--- a/packages/js/client-config-builder/README.md
+++ b/packages/js/client-config-builder/README.md
@@ -430,7 +430,7 @@ A complete example using all or most of the available methods.
    *
    * @returns CoreClientConfig that results from applying all the steps in the builder pipeline
    */
-  build(): CoreClientConfig;
+  build(options?: BuildOptions): CoreClientConfig;
 ```
 
 ## Bundles
--- a/packages/js/client-config-builder/src/ClientConfigBuilder.ts
+++ b/packages/js/client-config-builder/src/ClientConfigBuilder.ts
@@ -1,7 +1,6 @@
 import { getDefaultConfig } from "./bundles";
 import { BaseClientConfigBuilder } from "./BaseClientConfigBuilder";
-import { IClientConfigBuilder } from "./types/IClientConfigBuilder";
-import { BuilderConfig } from "./types";
+import { IClientConfigBuilder, BuilderConfig, BuildOptions } from "./types";
 
 import {
   CoreClientConfig,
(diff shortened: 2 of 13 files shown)


Niraj-Kamdar (commented):
I wonder if we need a separate uri resolver for this, Can't we just use a normal redirect resolver? I agree that it may help save a few line of code for users but they can create something similar themselves. This is also not very configurable. It just redirects interface to first implementation which wouldn't be a behavior that everyone wants

krisbitney:
> I wonder if we need a separate uri resolver for this, Can't we just use a normal redirect resolver? I agree that it may help save a few line of code for users but they can create something similar themselves. This is also not very configurable. It just redirects interface to first implementation which wouldn't be a behavior that everyone wants

Fair enough! It's also true that when using embedded packages (plugins or wrappers) it is often possible to use the interface URI as the implementation URI:
```ts
// in interfaces config
[defaultInterfaces.concurrent]: new Set([defaultInterfaces.concurrent]),

// in packages config
[defaultInterfaces.concurrent]: concurrentPromisePlugin({}),
```

But if you're using non-embedded Wasm wrappers for implementations, or if the interface URI can't readily be assigned to a specific embedded package (e.g. the uri resolver interface), then this new resolver is useful.

This resolver is syntax sugar that lets you do something like this without adding a redirect:
```ts
// in interfaces config
[defaultInterfaces.concurrent]: new Set(["wrap://package/concurrent"]),

// in packages config
["wrap://package/concurrent"]: concurrentPromisePlugin({}),
```

krisbitney:
I now know we can remove all of the global redirects without this resolver. I didn't realize that would be possible at the time Jordan and I talked about the issue. We were just thinking about ways to remove global redirects. (In retrospect, I think Jordan was alluding to the idea of using the interface URI as the implementation URI, but it went over my head.)

So I'm not sure whether this resolver is immediately useful to us, even as syntax sugar.

krisbitney:
IIRC, we decided this was not needed

PR #1527, opened 2023-02-07
update dependencies CLI version to 0.10.0-pre.7

krisbitney:
This will make it possible to remove import redirects from the plugins that still use them


krisbitney:
This will be irrelevant very soon, once we remove the last of the plugins

PR #1528, opened 2023-02-08
Request synchronizer resolver

nerfZael:
Rework of: https://github.com/polywrap/toolchain/pull/1493/files

With URI resolvers, multiple requests for the same URI, in most cases, needlessly repeat the same process (usually a network request).
Using a cache resolver (like PackageToWrapperCacheResolver) helps if the resolution requests are synchronous (one after another).
This new `RequestSynchronizerResolver` can be used to reuse parallel requests for the same URI, this way, only the first one needs to do the work (e.g. a network request) while others will await that same promise.
e.g. 
```typescript
await Promise.all([
  client.invoke("wrap://ipfs/Qmdssd"),
  client.invoke("wrap://ipfs/Qmdssd")
]);
```
Since the URIs are the same, if using the `RequestSynchronizerResolver`, this will result in only one call to the IPFS resolver.

In case of a thrown exception the resolver will also throw (for all listeners)
In case of a resolution error, by default, the resolver will use the cached error for all listeners.
If the user wants to modify that behavior, they can use the `shouldIgnoreCache` handler to use cache in case of a specific resolution error type (or all errors)


packages/js/uri-resolvers/src/cache/RequestSynchronizerResolver.ts
```diff
@@ -0,0 +1,87 @@
...
+            description: "RequestSynchronizerResolver (Cache)",
+          });
+
+          return result;
+        },
+        () => {
+          // In case of an error, we try to resolve the URI again.
+          // This is because the error might be caused by a network issue or something similar,
+          // and we don't want all the requests to fail.
+          return this.tryResolveUri(uri, client, resolutionContext);
```
krisbitney: At first glance, it looks like this could recurse many times if the URI is bad and there are many repeated calls

packages/js/uri-resolvers/src/__tests__/package-to-wrapper-cache-resolver.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+        defaultPackages.ipfsResolver,
+        defaultPackages.httpResolver,
+      ]
+    )
+    .buildCoreConfig();
+
+  return new PolywrapClient(config, { noDefaults: true });
+};
+
+describe("PackageToWrapperCacheResolver", () => {
```
krisbitney: filename and test names should probably read `"RequestSyncrhonizerResolver"`

yarn.lock
```diff
@@ -4867,6 +4867,13 @@ async-limiter@~1.0.0:
   resolved "https://registry.yarnpkg.com/async-limiter/-/async-limiter-1.0.1.tgz#dd379e94f0db8310b08291f9d64c3209766617fd"
   integrity sha512-csOlWGAcRFJaI6m+F2WKdnMKr4HhdhFVBk0H/QbJFMCr+uO2kwohwXQPxw/9OCxp05r5ghVBFSyioixx3gfkNQ==
 
+async-mutex@0.4.0:
```
krisbitney: it looks like you forgot to remove the `async-mutex` dependency or possibly just to update `yarn.lock`

krisbitney (changes requested):
Amazing work!

packages/js/uri-resolvers/src/cache/RequestSynchronizerResolver.ts
krisbitney: It would be nice to see a test that demonstrates this isn't an issue. 

My thought is that a React app, for example, can sometimes render frequently and each render can involve many invocations (e.g., as between-render effects).

nerfZael: Yep, this was an oversight.
Now, in case of a thrown exception the resolver will also throw (for all listeners)
In case of a resolution error, by default the resolver will ignore the cache for all but the first listener.
If the user wants to modify that behavior, they can use the `shouldUseCache` handler to use cache in case of a specific resolution error type (or all errors)

packages/js/uri-resolvers/src/cache/RequestSynchronizerResolver.ts
```diff
@@ -0,0 +1,121 @@
...
+          )
+          .finally(() => {
+            // After every listener has been notified with the above resolve or reject, remove the request from the cache.
+            this.requestCache.delete(uri.uri);
+          });
+      }
+    );
+
+    this.requestCache.set(uri.uri, resolutionRequest);
+
```
krisbitney, line 148: Do we need to track resolution history for the first request?

nerfZael: We should, I added it

dOrgJelli (dismissed):

dOrgJelli (approved):

PR #1529, opened 2023-02-08
`polywrap create template` command

### LINKED ISSUE #768: CLI Create Command: --template URL Option
Support the creation of new projects from externally hosted template project repositories like so:
`polywrap create --template ${url}`

Initially, the first supported `${url}` types can be git. In the future we can support other folder-based-repository protocols (like IPFS for ex).

krisbitney:
This PR adds the `template` command to the CLI's `polywrap create`. The `template` command lets users provide a url and downloads the template for them.

As described in https://github.com/polywrap/toolchain/issues/768 , the initial implementation only supports `.git` URLs. I implemented it with the assumption that additional URL formats will be added now or in the future.

Closes https://github.com/polywrap/toolchain/issues/768


packages/cli/src/lib/project/templates/downloadProjectTemplate.ts
```diff
@@ -0,0 +1,68 @@
...
+  const command = "git clone";
+  const args = ["--depth", "1", "--single-branch", url];
+  const repoName = path.basename(url, ".git");
+  const repoDir = path.join(cacheDir, repoName);
+  const dotGitPath = path.join(repoDir, "/.git/");
+
+  return new Promise((resolve, reject) => {
+    // clone repo
+    runCommand(command, args, logger, undefined, cacheDir)
+      .catch(() => reject({ command: `${command} ${args.join(" ")}` }))
```
cbrzn: i know this is still draft but it would be awesome if we use `await` and `try/catch` instead of `then/catch`

krisbitney: I can do that, but why? I actually wrote it both ways and imo it looks cleaner this way. It's just a style preference, right?

krisbitney: I refactored to use `await` with `try/catch`

cbrzn: yes! its just a style preference - I think we should follow one pattern in the codebase hehe, just for the sake of consistency

krisbitney: sounds good!

packages/cli/src/__tests__/e2e/create.spec.ts
```diff
@@ -134,4 +144,82 @@ describe("e2e tests for create command", () => {
...
+          args: ["create", "template", "lang", "demo"],
+          cli: polywrapCli,
+        });
+
+        expect(code).toEqual(1);
+        expect(error).toContain(`URL 'lang' uses an invalid format. Valid URL formats: .git`);
+        expect(output).toBe("");
+      });
+
+      for (const format of Object.values(UrlFormat)) {
```
dOrgJelli, line 182: Future proof testing 🔥 🔥 🔥

packages/cli/src/lib/project/templates/downloadProjectTemplate.ts
```diff
@@ -0,0 +1,73 @@
...
+    const message = intlMsg.commands_create_error_badUrl({
+      url: `'${url}'`,
+      formats,
+    });
+    throw Error(message);
+  }
+}
+
+function createCacheDir(): string {
+  const cacheDir = path.resolve(`./.polywrap/${createUUID()}`);
```
dOrgJelli: Since we need access to a cache directory w/o a valid `polywrap.yaml` project present, I think this gives a good reason to create a "user-level cache directory". For example, the cross-platform equivalent of `~/.polywrap/`.

Checkout the PWR codebase, they've done this there. Also we could export a global class instance of the `Cache` class we already have defined (used by project classes).

dOrgJelli (changes requested):
This is super clean @krisbitney, amazing & quick work. Only thing I'd like to be different is the treatment of the cache directory. Let's quickly make that more uniform and use a user-level cache within the operating system. That way other functionality can make use of it in the future too! (ex: import caching, etc)

packages/cli/src/lib/project/templates/downloadProjectTemplate.ts
krisbitney: I added a user-level cache directory in the same dir that pwr uses. I am also now using the `CacheDirectory` class.

dOrgJelli (approved):
Amazing 👏

pileks (approved):
:ship:

PR #1534, opened 2023-02-09
Remove legacy config types from PolywrapClient

### LINKED ISSUE #1506: Remove "legacy" Client config object types
With the latest changes to the ClientConfigBuilder and its new BuilderConfig object outlined in #1480, we are converging onto a unified system of configuration, which consists of:

- `BuilderConfig` as the config object used by the `ClientConfigBuilder` internally
- `ClientConfigBuilder` which can spit out a `CoreClientConfig`, and
- `CoreClientConfig`, which is now used both within the `PolywrapCoreClient` and the `PolywrapClient`

With this convergence, we have 3 "vestigial" config objects which the `PolywrapClient` accepts and transforms into a `CoreClientConfig` using the `ClientConfigBuilder`:
- `PolywrapClientConfig`
- `ClientConfig`
- `PolywrapCoreClientConfig`

We could (and likely should) remove these 3 vestigial config types before `0.10` in order to facilitate a less confusing configuration experience. This would include removing everything from the `legacy` folder in `packages/js/client`, including the now public-facing generic types for Envs, etc.

My proposal is that `PolywrapClient`'s constructor only accepts `CoreClientConfig`, which can be built using the `ClientConfigBuilder`.
I'm on the fence about whether we'd want `PolywrapClient`'s constructor to accept a `BuilderConfig` object, as this seems to be the last place where there is a dependency between `@polywrap/client-js` and `@polywrap/client-config-builder-js`. Separating these two isn't a top priority, but it would be nice to have, as we could have each of them then be 100% focused on their own concerns.

**Describe alternatives you've considered**
We can keep this `legacy` folder alive for however long we'd like, but we're then nudging developers into 4 separate ways of setting up the PolywrapClient - not ideal IMO.

**Additional context**
/

pileks:
Resolves #1506 

This PR aims to remove the `PolywrapCoreClientConfig`, `PolywrapClientConfig`, `ClientConfig`, and the associated types found in the `legacy` folder within `@polywrap/client-js`.

As a result, the `PolywrapClient`'s constructor would only need to accept an optional `CoreClientConfig` parameter, which would override the default config:

```typescript
const clientWithDefaults = new PolywrapClient();

  .addDefaults()
  .add(...)
  .build();

const clientWithConfig = new PolywrapClient(config);

```

Remaining work:

* [x] ~~Workflows (`test` command) are currently using the `PolywrapClientConfig` type as a basis for the configuration structure of each `Step` of a `Job`. This should be changed into the same structure used for `--client-config` within the CLI - pointing to a module which exposes a `configure(config)` function which returns a `ClientConfigBuilder`. This will be handled as part of a separate PR targeting this one.~~ The consensus is that we should **remove** the `config` property of the `Step`, as it's not used and untested. This is done in #1545


packages/cli/src/lib/defaults/deploy-modules/ens/index.ts
```diff
@@ -49,11 +52,22 @@ class ENSPublisher implements DeployModule {
...
 
-    const client = new PolywrapClient({
-      packages: [
-        {
-          uri: defaultInterfaces.ethereumProvider,
-          package: ethereumProviderPlugin({
+    const clientConfig = new ClientConfigBuilder()
+      .addDefaults()
+      .add({
+        envs: {
```
krisbitney, line 58: The redirects aren't needed here anymore (or in the other ENS deploy module).

packages/js/client/src/PolywrapClient.ts
```diff
@@ -31,58 +25,18 @@ import {
...
-      }
-
-      Tracer.startSpan("PolywrapClient: constructor");
-    } catch (error) {
-      Tracer.recordException(error);
-      throw error;
-    } finally {
-      Tracer.endSpan();
-    }
+  constructor(config?: CoreClientConfig) /* $ */ {
```
krisbitney, line 37: I really love how simple this becomes!

packages/js/client/src/PolywrapClient.ts
```diff
@@ -1,11 +1,5 @@
 import { InvokerOptions, TryResolveUriOptions } from "./types";
-import {
-  PolywrapClientConfig,
-  PolywrapCoreClientConfig,
-  buildPolywrapCoreClientConfig,
-  sanitizeUri,
-  ClientConfig,
-} from "./legacy";
+import { sanitizeUri } from "./legacy";
```
krisbitney, line 2: Can we delete the legacy config types and rename the folder to something like "helpers" or "util"? Or are we not ready for that yet?

krisbitney (changes requested):
This is awesome!! It really simplifies the configs. Thanks for doing this.

packages/js/client/src/PolywrapClient.ts
krisbitney: okay, i'm just now reviewing your other PR and I see this is done there

packages/js/client/src/PolywrapClient.ts
cbrzn: GG

packages/js/plugins/uri-resolvers/ens-resolver/src/__tests__/helpers/getClient.ts
```diff
@@ -17,82 +17,81 @@ import {
   defaultInterfaces,
   defaultPackages,
   defaultIpfsProviders,
-  defaultWrappers
+  defaultWrappers,
 } from "@polywrap/client-config-builder-js";
 import { httpPlugin } from "@polywrap/http-plugin-js";
 import { Uri } from "@polywrap/core-js";
 
 export const getClient = () => {
```
cbrzn, line 25: i guess these `getClient` functions can be improved to use the builder with default? And just adding what we need

cbrzn (dismissed):
:_) thanks for this!

cbrzn:
btw, just my two cents, i would merge this one first and then #1545, just for the sake of smaller PR reviews :smile:

packages/js/plugins/uri-resolvers/ens-resolver/src/__tests__/helpers/getClient.ts
pileks: For sure. My main focus here was just to "get this working". I'd say any other refactors such as these will happen as we see the need for them arise. I know some like to set up their config manually, without builder

packages/cli/src/lib/defaults/deploy-modules/ens/index.ts
pileks: Removed them. Let's let the test turn green and we're good!

cbrzn (dismissed):

krisbitney (approved):
Looks good!

cbrzn (approved):

PR #1545, opened 2023-02-13
Remove `config` section from test manifest & remove legacy config types from Client

pileks:
~~This PR targets #1534, and should be merged into its branch before #1534 is merged.~~
As per @cbrzn’s comment, merge this after #1534 is merged.

This PR aims to do two things:

1. Remove the `config` property from the `steps` in the Test Manifest.
2. Remove the entire `legacy` folder from `@polywrap/client-js`, as it was last referenced within the Test Manifest.


packages/js/manifests/polywrap/src/formats/polywrap.test/migrators/0.1.0_to_0.2.0.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Jobs as OldJobs, PolywrapWorkflow as OldManifest } from "../0.1.0";
+import { Jobs as NewJobs, PolywrapWorkflow as NewManifest } from "../0.2.0";
+
+export function migrate(manifest: OldManifest, logger?: ILogger): NewManifest {
+  const [jobsSanitized, jobsHaveStepsWithConfig] = sanitizeJobsSteps(
+    manifest.jobs
+  );
+
+  if (jobsHaveStepsWithConfig) {
+    logger?.warn(
```
krisbitney, line 11: This warning is great

krisbitney (dismissed):
LGTM!

cbrzn (dismissed):
lgtm :green_apple:

packages/js/manifests/polywrap/src/formats/polywrap.test/migrators/0.1.0_to_0.2.0.ts
cbrzn: this warning made me think we probably want to have translations for these messages? since they're thrown through the CLI :thinking: anyways, a topic for another PR in any case :yum:

pileks: The 1st time I issued such a warning I also thought of this. However, it's a bit hard to reason around this without extracting intl into its own package, as currently only the CLI has intl support. That's why I went this route and said "We'll cross that bridge when we get there."

We're also not using intl right now (as both `en` and `es` are in English), so I think it's a low-priority issue right now.

cbrzn (dismissed):

krisbitney (approved):

cbrzn (approved):

PR #1546, opened 2023-02-13
Add `RetryResolver` to retry resolution attempts on failure

krisbitney:
This `IUriResolver` implementation will retry resolution when it fails. The resolver can be configured to work with specific URIs or URI authorities. Custom timeouts can be set to pause between retries.

TODO:
- [ ] Update URI resolution tests in the `@polywrap/client-js` package


krisbitney:
This resolver doesn't solve the problem of retries on `getFile`. Need to rethink this!

dOrgJelli:
I think I'm onboard with this, but can we please add more context as to why this route was taken? For example, maybe we wanted to "simplify the ipfs resolver wrapper" or "apply retry logic to other resolvers".

@nerfZael we should also get your review on this before merging.

krisbitney:
> I think I'm onboard with this, but can we please add more context as to why this route was taken? For example, maybe we wanted to "simplify the ipfs resolver wrapper" or "apply retry logic to other resolvers".
> 
> @nerfZael we should also get your review on this before merging.

I think it can go either way. The `RetryResolver` adds more flexibility and makes it possible to add a timeout between retries.

krisbitney:
I think it makes sense to keep retries in the ipfs resolver wrapper so the Rust and Python clients can use the feature. I wasn't thinking about other clients when I was making this PR.

krisbitney:
@dOrgJelli @nerfZael Do we want this resolver or do we want to keep retries in the ipfs wrapper?

nerfZael:
> @dOrgJelli @nerfZael Do we want this resolver or do we want to keep retries in the ipfs wrapper?

I think we want both.
It's useful to keep retries in the ipfs resolver for the reasons you listed above, and this resolver can be used with any other URI resolver

packages/js/uri-resolvers/src/helpers/RetryResolver.ts
```diff
@@ -0,0 +1,113 @@
...
+import {
+  CoreClient,
+  IUriResolutionContext,
+  IUriResolver,
+  Result,
+  Uri,
+  UriPackageOrWrapper,
+} from "@polywrap/core-js";
+
+/** configure how additional resolution attempts are handled after an initial resolution attempt fails */
```
nerfZael: Nit: lets be consistent with casing. Either start each comment with uppercase or lowercase.
Personally, I prefer uppercase.

packages/js/uri-resolvers/src/helpers/RetryResolver.ts
```diff
@@ -0,0 +1,113 @@
...
+    uri: Uri,
+    client: CoreClient,
+    resolutionContext: IUriResolutionContext
+  ): Promise<Result<UriPackageOrWrapper, TError>> {
+    const subContext = resolutionContext.createSubHistoryContext();
+
+    const result = await this.resolver.tryResolveUri(uri, client, subContext);
+
+    const isChange = !(
+      result.ok &&
```
nerfZael, line 57: So this will retry if (and only if) a URI is not found?
Is that what we want? I thought we want to retry in case of an error (like a request timeout=

packages/js/uri-resolvers/src/helpers/RetryResolver.ts
```diff
@@ -0,0 +1,113 @@
...
+
+/** A map or uri or authority strings to retry options */
+export type RetryResolverOptions = {
+  [uriOrAuthority: string]: RetryOptions;
+};
+
+export class RetryResolver<TError = undefined> implements IUriResolver<TError> {
+  constructor(
+    private resolver: IUriResolver<TError>,
+    private options: RetryResolverOptions
```
nerfZael, line 34: It'd be nice to not have to rely on authority for all cases here.
Like having a handler to decide if the request should retry or not.
Maybe both (e.g. a union type) ?

packages/js/uri-resolvers/src/__tests__/retry-resolver.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+  ipfsProviders,
+  plugins,
+} from "@polywrap/client-config-builder-js/build/bundles/default";
+
+jest.setTimeout(200000);
+
+const getClientWithRetryResolver = (retryOptions: RetryResolverOptions): PolywrapClient => {
+  const ipfsResolverUri = Uri.from("wrap://package/ipfs-resolver");
+  const resolver = RecursiveResolver.from(
+    PackageToWrapperCacheResolver.from(
```
nerfZael: We don't need all these resolvers just to test the RetryResolver.
A simple test resolver instead should do the trick.
Checkout this for inspiration: https://github.com/polywrap/toolchain/pull/1586/files#diff-57a872ca435da8f15bb68754ef07491fa13b9b2d043ffed209e474f495c7b662

packages/js/client-config-builder/src/ClientConfigBuilder.ts
```diff
@@ -52,8 +53,10 @@ export class ClientConfigBuilder extends BaseClientConfigBuilder {
                   ...this.buildWrappers(),
                   ...this.buildPackages(),
                 ]),
-                ...this._config.resolvers,
-                new ExtendableUriResolver(),
+                new RetryResolver(new ExtendableUriResolver(), {
```
nerfZael, line 56: Not sure about this being a part of the default resolver setup.
Feels very custom to me.

nerfZael (changes requested):

packages/js/uri-resolvers/src/helpers/RetryResolver.ts
```diff
@@ -0,0 +1,113 @@
...
+        result.value.type === "uri" &&
+        result.value.uri.uri === uri.uri
+      );
+
+      if (isChange) {
+        resolutionContext.trackStep({
+          sourceUri: uri,
+          result,
+          subHistory: subContext.getHistory(),
+          description: "RetryResolver (Retry)",
```
nerfZael: I dont think this case is covered in tests

packages/js/uri-resolvers/src/helpers/RetryResolver.ts
```diff
@@ -0,0 +1,113 @@
...
+
+/** configure how additional resolution attempts are handled after an initial resolution attempt fails */
+export type RetryOptions = {
+  /** the number of additional resolution attempts */
+  retries: number;
+  /** the duration (in ms) to pause between attempts */
+  interval: number;
+};
+
+/** A map or uri or authority strings to retry options */
```
nerfZael: Typo: `A map of URI or authority...`
`of` not `or`

packages/js/uri-resolvers/src/helpers/RetryResolver.ts
```diff
@@ -0,0 +1,113 @@
...
+  Uri,
+  UriPackageOrWrapper,
+} from "@polywrap/core-js";
+
+/** configure how additional resolution attempts are handled after an initial resolution attempt fails */
+export type RetryOptions = {
+  /** the number of additional resolution attempts */
+  retries: number;
+  /** the duration (in ms) to pause between attempts */
+  interval: number;
```
nerfZael: A good practice for time variables is to put the unit in the name. E.g. intervalMs/intervalInMilliseconds/intervalMilliseconds

packages/js/uri-resolvers/src/helpers/RetryResolver.ts
krisbitney: no, it's `or`

They key can be a URI or a URI authority

packages/js/uri-resolvers/src/helpers/RetryResolver.ts
krisbitney: If there is an error in a uri resolver extension, such as a timeout, it is currently returned as no change with `result.ok === true`

packages/js/uri-resolvers/src/helpers/RetryResolver.ts
krisbitney: Good idea! it could accept a callback

packages/js/uri-resolvers/src/helpers/RetryResolver.ts
krisbitney: good catch!

packages/js/uri-resolvers/src/helpers/RetryResolver.ts
krisbitney: Nice catch

krisbitney:
Since this resolver probably shouldn't be part of the default resolver configuration, it probably shouldn't be added to the uri resolvers package where it would become another item for us to maintain. I'm closing this PR. We can revisit this feature concept if a user need arises.

PR #1547, opened 2023-02-14
Move WebSocket plugin to https://github.com/polywrap/WebSocket

krisbitney:
I moved the WebSocket plugin to its own repo at https://github.com/polywrap/WebSocket. I created an interface wrapper, based on the plugin's schema, and made the plugin implement the interface.

I've noticed the WebSocket plugin was never in the default client config. Was this intentional or an oversight? Should we add it?


cbrzn:
> I've noticed the WebSocket plugin was never in the default client config. Was this intentional or an oversight? Should we add it?

this was done on purpose because we don't use the ws plugin, so there wasn't a need to add this dependency

krisbitney:
> > I've noticed the WebSocket plugin was never in the default client config. Was this intentional or an oversight? Should we add it?
> 
> this was done on purpose because we don't use the ws plugin, so there wasn't a need to add this dependency

Okay, cool. I assumed it was on purpose but I just wanted to check.

Niraj-Kamdar (dismissed):

cbrzn (dismissed):
with this PR, the `build:plugins` command has been removed :smile: 

arigato @krisbitney for pushing this forward!!! really happy to finally see this :yum:

dOrgJelli (approved):
👏 🎬 that's a warp!

PR #1556, opened 2023-02-16
Fix: WrapError now correctly parsing Rust unwrap errors

krisbitney:
When calling `.unwrap()` on a Rust result that contains an error, Rust will panic with an error message that contains the `Err`. For fidelity to the original `Err`, Rust inserts escape characters in the string. For example, `"\n"` becomes `"\\n"`. This behavior was not being handled correctly by `WrapError`'s string parsing logic.


dOrgJelli (approved):

dOrgJelli:
Great catch @krisbitney, should this also be a patch for `0.9`?

PR #1557, opened 2023-02-16
JS client: change `envs` and `interfaces` into readonly maps

### LINKED ISSUE #1283: JS Client: Client Config Should Use Unique Maps Instead Of Unique Arrays

pileks:
Closes #1283 

This PR aims to change the `envs` and `interfaces` collections inside `CoreClientConfig` into readonly unique maps/map-like structures.

`envs` is now essentially:
```typescript
Record<string, Record<string, unknown>>
```

`interfaces` is now essentially:
```typescript
Record<string, string[]>
```

Everything that can be is marked as `readonly` in order to relay the intended readonly nature of the CoreClientConfig.

Some comments/feedback needed:
- [ ] I'm not 100% sure I like having **everything** as `string`s, but implementing some custom `URI-Map` seems like a worse idea. This does leave us vulnerable to people mixing `wrap://xyz/...` and `xyz/...` as two separate objects within the maps, but that's something that should be handled by the ClientConfigBuilder.
  - Again, I feel like this is something to think about as the `ClientConfigBuilder` is not **always** used yet, and I know some devs want to simply build their own config.
- [ ] We might want to think about `interfaces` being a `Record<string, Uri[]>` instead. Thoughts?


krisbitney (dismissed):
Everything looks good. This is well done! The use of objects instead of arrays looks really clean.

Before merging, I think the question about using `string` vs. `Uri` should be resolved, just because of the recent decision to use `Uri` in the core.

I think making the implementations array of type `Uri[]` instead of `string[]` makes sense.

packages/js/client/src/__tests__/e2e.spec.ts
```diff
@@ -1,15 +1,15 @@
...
+// import { subinvokeCase } from "./core/wrap-features/subinvoke-case";
 import { envTestCases } from "./core/wrap-features/env-case";
-import { interfaceInvokeCase } from "./core/wrap-features/interface-implementation-case";
+// import { interfaceInvokeCase } from "./core/wrap-features/interface-implementation-case";
 
 export const supportedImplementations = ["as", "rs"];
 describe.each(supportedImplementations)(
   "client <-> wrappers end to end",
   (i) => {
-    typeTestCases(i);
```
dOrgJelli, line 10: Oops, need to revert this

packages/js/core/src/types/CoreClient.ts
```diff
@@ -60,22 +67,22 @@ export interface CoreClient extends Invoker, UriResolverHandler<unknown> {
    *
    * @returns an array of interfaces and their registered implementations
    */
-  getInterfaces(): readonly InterfaceImplementations[] | undefined;
```
dOrgJelli, line 63: I noticed some inconsistencies in this file when it comes to `readonly` and `ReadOnly<...>` usage. Could we please use it the same in all methods / properties? If I'm missing some nuances please inform :)

dOrgJelli (changes requested):

packages/js/core-client/src/PolywrapCoreClient.ts
```diff
@@ -86,15 +85,13 @@ export class PolywrapCoreClient implements CoreClient {
...
-    const uriUri = Uri.from(uri);
-
+  public getEnvByUri(uri: Uri): Readonly<WrapperEnv> | undefined /* $ */ {
     const envs = this.getEnvs();
     if (!envs) {
       return undefined;
     }
 
-    return envs.find((environment) => Uri.equals(environment.uri, uriUri));
+    return envs[uri.uri];
```
nerfZael, line 94: This will probably treat "wrap://ens/" and "ens/" as different URIs

nerfZael: We could maybe sanitize this stuff in the constructor

nerfZael (changes requested):

PR #1559, opened 2023-02-17
0.9 feat: add retries to ipfs resolver plugin

krisbitney:
This non-breaking change would be a huge improvement for the uniswap v3 front end, and I imagine the EthForAll hackers


Niraj-Kamdar (approved):
LGTM!

Niraj-Kamdar:
How about also adding auth support with env?

dOrgJelli (approved):
Do we have the equivalent of this in 0.10?

krisbitney:
> Do we have the equivalent of this in 0.10?

Yeah, we do. I know you saw the other PRs related to this since your comment. It's a really great feature!

PR #1560, opened 2023-02-17
fix: add BuildOptions to build method in IClientConfigBuilder

krisbitney:
This makes it possible to add a custom cache or resolver without casting.

Before:
```ts
const builder: IClientConfigBuilder = new ClientConfigBuilder().addInterface(...);
const config = (builder as ClientConfigBuilder).build(resolver);
```

After:
```ts
const builder: IClientConfigBuilder = new ClientConfigBuilder().addInterface(...);
const config = builder.build(resolver);
```

### DIFF
--- a/packages/js/client-config-builder/README.md
+++ b/packages/js/client-config-builder/README.md
@@ -428,9 +428,10 @@ A complete example using all or most of the available methods.
   /**
    * Build a sanitized core client configuration that can be passed to the PolywrapClient or PolywrapCoreClient constructors
    *
+   * @param options - Use a custom wrapper cache or resolver
    * @returns CoreClientConfig that results from applying all the steps in the builder pipeline
    */
-  build(): CoreClientConfig;
+  build(options?: BuildOptions): CoreClientConfig;
 ```
 
 ## Bundles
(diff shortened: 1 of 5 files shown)


Niraj-Kamdar (approved):
LGTM!

cbrzn (commented):
i thought about this, and I am not sure if we want this. the reason is that (imo) we don't want the interface to be aware of our implementation details. If someone wants to implement its builder, he probably needs to implement the build method with our options (or with no options at all). 

for example (iiuc), this piece of code will throw error
```typescript
export class MyBuilder extends BaseClientConfigBuilder {
  addDefaults(): IClientConfigBuilder {
    return this;
  }
  build(_options?: { custom: string }): CoreClientConfig {
    return {} as CoreClientConfig;
  }
}
```
edit: I might be wrong tho but just wanted to share my thoughts :smile:

krisbitney:
> i thought about this, and I am not sure if we want this. the reason is that (imo) we don't want the interface to be aware of our implementation details. If someone wants to implement its builder, he probably needs to implement the build method with our options (or with no options at all).
> 
> for example (iiuc), this piece of code will throw error
> 
> ```ts
> export class MyBuilder extends BaseClientConfigBuilder {
>   addDefaults(): IClientConfigBuilder {
>     return this;
>   }
>   build(_options?: { custom: string }): CoreClientConfig {
>     return {} as CoreClientConfig;
>   }
> }
> ```
> 
> edit: I might be wrong tho but just wanted to share my thoughts 😄

I think this is a bad UX. I thought it was a bug. Maybe we should reconsider how we add a custom cache or custom resolver?

cbrzn:
> I think this is a bad UX. I thought it was a bug. Maybe we should reconsider how we add a custom cache or custom resolver?

yup, I think this is a code smell that is indicating us that

dOrgJelli (approved):

dOrgJelli (approved):

PR #1561, opened 2023-02-17
fix(cli): remove builder by default when build image strategy with buildx

### LINKED ISSUE #1479: buildx docker container still running after build
**Describe the bug**
The docker container with a name like `buildx_buildkit_polywrap-build-env-804329be-892e-4979-bdc7-627685ee8fb20` remains running after building a wrapper. It is not automatically stopped.

I think it can be replicated by building any wrapper with the test env's `buildWrapper` function.

**Expected behavior**
I would expect the container to be stopped when the build comletes.

**Screenshots / logs**
Screenshot 2023-01-11 at 3 28 32 PM

- Mac OS with Apple cpu

Niraj-Kamdar:
- Closes: #1479 
### Breaking Change:
- Added new `keepBuilder` option which is default to false and removed the previous: `removeBuilder` option from the build manifest


pileks (approved):
LGTM!

dOrgJelli (approved):

PR #1565, opened 2023-02-18
[0.9] feat(ipfs-plugin-js): add support for auth

Niraj-Kamdar:


dOrgJelli (approved):
Ah my mistake, just saw this after having started the 0.9 patch release process, will get this in.

Also are we in agreement that technically adding "optional plugin env properties" is still considered a patch it's technically backward & forward compatible? It could be considered a "feature" if you think about it...

PR #1576, opened 2023-02-21
fix: updated ethereum wrapper and ethereum provider plugin to fix bug preventing waiting for transactions

krisbitney:
This PR updates the ethereum wrapper, the ethereum provider interface wrapper, and the ethereum provider plugin. The updates fix a bug where transactions were not being waited for when users made calls to `sendTransactionAndWait`, `callContractMethodAndWait`, or `awaitTransaction`.


packages/js/plugins/uri-resolvers/ens-resolver/README.md
```diff
@@ -37,7 +37,7 @@ export async function foo({
   client = new PolywrapClient({
     plugins: [
       {
-        uri: "wrap://ens/wraps.eth:ethereum@1.0.0",
+        uri: "wrap://ens/wraps.eth:ethereum@1.1.0",
```
cbrzn, line 40: i think this readme is outdated right?

krisbitney: It is, yeah. We will remove the ens resolver from the repo soon, anyway.

dOrgJelli (changes requested):
Hey @krisbitney I think we should make further updates to the default config. First I don't think we should stop support for invoking the ethereum-provider @ `1.0.0`. Currently this PR would cause that URI resolution to fail.

Additionally I don't think we need to update all of our test-cases, this seems like a needless change, as the old `1.0.0` URI should still work.

Lastly after looking at the default config, I found a few things I think we can simplify, so I'm going to make a new PR with those larger changes. Will comment back here once that's complete.

dOrgJelli:
Created a new PR here: https://github.com/polywrap/toolchain/pull/1582

I've also removed all built-in uri-resolvers in favor of their uri-resolver-ext equivalents.

PR #1577, opened 2023-02-21
fix: rebuild wrap-test-harness when necessary

cbrzn:
In this PR, we've added support for rebuilding the test-harness wrappers @ https://github.com/polywrap/wrap-test-harness when necessary. This rebuilding is required when software within the toolchain that may effect the `polywrap build` command is changed.

Additionally, we've added support for rebuilding from a specific branch within the `polywrap/wrap-test-harness` repository. This will be necessary if there are breaking changes within the toolchain, that require updates to the test wrappers. In the event of this happening, the following steps would need to be taken:
1. Open a PR in this repo (toolchain) with breaking changes.
2. Create a branch within the `polywrap/wrap-test-harness` repo.
3. Update wrappers based on latest toolchain changes.
4. Create a 2nd PR in the test-harness repo w/ changes.
5. Modify the `WRAP_TEST_HARNESS` file within the toolchain repo w/ the branch you've created in the wrap-test-harness.

Once the PR is merged in toolchain, then the wrap-test-harness PR can be merged, and the `WRAP_TEST_HARNESS` branch pointer can be reset.


.github/workflows/ci-javascript.yaml
```diff
@@ -44,36 +44,41 @@ jobs:
...
     uses: ./.github/workflows/generate-modified-test-wrappers.yaml
     with:
       cli-path: /toolchain/packages/cli
       wasm-path: /toolchain/packages/wasm
-
+      harness-ref: ${{ needs.Should-Generate-Test-Wrappers.outputs.wrappers_branch }}
 
   Test-Core:
     runs-on: ubuntu-latest
-    needs: Generate-Test-Wrappers
```
dOrgJelli, line 76: IDK this seems off... we only test the core wrappers if we "tell" the CI to do so via the special `/` command in the description? Seems like we'd want to do it if the core files change.

.github/workflows/generate-modified-test-wrappers.yaml
```diff
@@ -82,8 +87,13 @@ jobs:
         if: ${{ inputs.wasm-path }}
         run: echo "POLYWRAP_WASM_PATH=${{ github.workspace }}/${{ inputs.wasm-path }}" >> $GITHUB_ENV
 
-      - name: Run wrappers build
-        run: cargo run -- -w
+      - uses: nick-fields/retry@v2
```
dOrgJelli, line 111: Let's move this bug fix into a separate PR? This fix seems like a non-controversial one, but the other workflow changes in `ci-javascript` and `ci-rust` are a bit more radicle.

dOrgJelli (changes requested):

.github/workflows/ci-javascript.yaml
cbrzn: > we only test the core wrappers if we "tell" the CI to do so via the special / command in the description?

yes

what I am trying to achieve here is to remove the need to create a new branch in the test harness when one of these files has changed.

it's worth mentioning that the tests will still run with the released wrappers (the ones that are generated via the `fetchWrappers` function).

based on your feedback, I would like to propose some approaches:

1.- (with the current changes) if these files are changed, we can make the bot comment something like: "Looks like you changed some files in the packages CLI, Schema, or WASM; remember to add the /generate-test-cases branch_name in your PR description.".

2.- another approach we might take is to keep the flow as it currently is, but the workflow will check if the branch exists in the test harness (remember, this branch will have the same name as the one in the toolchain); if it doesn't, then we can just build against master branch (this will require a custom script)

cbrzn: after thinking a bit more, a third one (and maybe the easier one) might be a mix of both worlds: we generate the wrappers every time the files change (as it currently is), BUT we can define the branch that's going to be used in the test harness using the /generate-test-case command? meaning that you don't **need** to create the branch in the test harness if it is not necessary :thinking:

dOrgJelli: Okay I'm better understanding the nuances here, hmmmm 🤔

dOrgJelli: Just got off a call w/ Cesar, we feel we have a good plan laid out, will detail it here along with the changes incomming.

dOrgJelli (approved):

PR #1581, opened 2023-02-22
Fix startup of eth-ens-ipfs infra module on WSL

### LINKED ISSUE #1580: CLI - `infra` command fails on latest WSL
**Describe the bug**
When running the `infra up` command with `--modules=eth-ens-ipfs`, docker fails setting the test env up with an issue regarding the compose file.

run `npx polywrap infra up --modules=eth-ens-ipfs` within WSL.

**Expected behavior**
The infra environment should start.

**Screenshots / logs**

```bash
pileks@HACKSTATION:~/dev/polywrap/toolchain/packages/cli$ npx polywrap -v
0.10.0-pre.8
pileks@HACKSTATION:~/dev/polywrap/toolchain/packages/cli$ npx polywrap infra up --modules=eth-ens-ipfs
Using infra modules: 
- eth-ens-ipfs

> docker stats --no-stream
The Compose file '/home/pileks/dev/polywrap/toolchain/packages/cli/.polywrap/infra/modules/local/eth-ens-ipfs/docker-compose.yaml' is invalid because:
Unsupported config option for services.ganache: 'platform'

The Compose file '/home/pileks/dev/polywrap/toolchain/packages/cli/.polywrap/infra/modules/local/eth-ens-ipfs/docker-compose.yaml' is invalid because:
Unsupported config option for services.ganache: 'platform'
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "#<Object>".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

- OS: WSL Ubuntu 20.x

**Additional context**
Seems to be a simple fix, will deploy a PR with a potential fix along this bug.

pileks:
Closes #1580 

It seems that the `platform` property isn't needed for `ganache` to start up, so I removed it and everything works!


packages/cli/src/lib/defaults/infra-modules/eth-ens-ipfs/docker-compose.yaml
```diff
@@ -1,7 +1,6 @@
 version: '3'
 services:
   ganache:
-    platform: linux/x86_64
```
dOrgJelli, line 4: Yes this seems like the correct approach, remove the specific platform specification, it's unneeded. This is because the docker image being used (`node:16-alpine`, see `./eth/Dockerfile`), is a multi-platform image. This means that docker will automatically detect what CPU architecture it is running within (amd64, arm, etc) and download the correct build of the image from docker hub: https://hub.docker.com/_/node/tags

dOrgJelli (approved):

krisbitney (approved):
LGTM!

packages/cli/src/lib/defaults/infra-modules/eth-ens-ipfs/docker-compose.yaml
pileks: Thank you for the detailed explanation!

krisbitney:
@dOrgJelli @pileks 

I think I prematurely approved this PR. I didn't actually test it locally. For me, it doesn't work without the platform tag. I can add the tag and build the CLI, and deploy tests pass. When I remove the tag and build the CLI, I get the `"Ganache failed to start"` error thrown by `@polywrap/test-env-js`.

I apologize for not checking this properly before!

pileks:
@krisbitney what OS are you using? Seems like something we should figure and see how to support all platforms.

krisbitney:
> @krisbitney what OS are you using? Seems like something we should figure and see how to support all platforms.

I'm using MacOS with the Apple cpu (arm architecture)

krisbitney:
Maybe we need to increase the version of Node to support the architecture? Let me try that and I'll comment again with the result

krisbitney:
@pileks Support for arm was added in node 16, so that wasn't the issue. It turns out the issue was the `node-alpine` image, which is a slimmed down version of node that uses some different libraries. A PR is up here with a fix: https://github.com/polywrap/toolchain/pull/1595

From https://hub.docker.com/_/node:
> "The main caveat to note is that it does use [musl libc](https://musl.libc.org/) instead of [glibc and friends](https://www.etalabs.net/compare_libcs.html), so software will often run into issues depending on the depth of their libc requirements/assumptions."

PR #1582, opened 2023-02-23
Integrate URI Resolver Extension Wrappers + Remove All Plugins

dOrgJelli:


dOrgJelli:
It looks like the final issue to solve here is that we're hitting an OOM (out of memory) error `WebAssembly.Memory(): could not allocate memory`. To address this, I think we should:
1. Create a reproducible test-case that limits the heap size of the process to something very low.
2. Run a test that produces this OOM error.
3. Create a garbage collection mechanism within the client to hopefully solve this.

dOrgJelli:
After more digging, I think the problem is much simpler, and can be solved by this existing PR: https://github.com/polywrap/toolchain/pull/1528

packages/cli/package.json
```diff
@@ -47,8 +47,7 @@
     "@polywrap/client-config-builder-js": "0.10.0-pre.8",
     "@polywrap/client-js": "0.10.0-pre.8",
     "@polywrap/core-js": "0.10.0-pre.8",
-    "@polywrap/ens-resolver-plugin-js": "0.10.0-pre.8",
-    "ethereum-provider-js": "0.1.7",
+    "ethereum-provider-js": "0.2.0",
```
Niraj-Kamdar: Can we deploy this under polywrap org?

packages/js/client-config-builder/package.json
```diff
@@ -24,20 +24,17 @@
...
     "@polywrap/ipfs-plugin-js": "~0.10.0-pre",
-    "@polywrap/ipfs-resolver-plugin-js": "~0.10.0-pre",
     "@polywrap/logger-plugin-js": "0.10.0",
+    "@polywrap/plugin-js": "0.10.0-pre.8",
     "@polywrap/uri-resolver-extensions-js": "0.10.0-pre.8",
     "@polywrap/uri-resolvers-js": "0.10.0-pre.8",
     "@polywrap/wasm-js": "0.10.0-pre.8",
     "base64-to-uint8array": "1.0.0",
-    "concurrent-plugin-js": "0.1.2"
+    "concurrent-plugin-js": "0.1.2",
```
Niraj-Kamdar: We also need to publish concurrent-plugin under polywrap.

Niraj-Kamdar (changes requested):

packages/cli/package.json
dOrgJelli: On it!

packages/js/client-config-builder/package.json
dOrgJelli: Will fix this in another PR, thank you for catching this!

packages/js/uri-resolver-extensions/src/ExtendableUriResolver.ts
```diff
@@ -25,7 +25,7 @@ export class ExtendableUriResolver extends UriResolverAggregatorBase<
   // $start: ExtendableUriResolver-extInterfaceUri-static
   /** The default interface uri to which implementations should be registered */
   public static extInterfaceUri: Uri = new Uri(
-    "wrap://ens/uri-resolver.core.polywrap.eth"
+    "wrap://ens/wraps.eth:uri-resolver-ext@1.0.0"
```
krisbitney: Although it doesn't really matter what this URI is, if we are using a versioned URI I think we should use version 1.1.0

packages/cli/src/lib/test-env/client-config.ts
```diff
@@ -29,14 +27,21 @@ export function getTestEnvClientConfig(): Partial<BuilderConfig> {
 
   return {
     envs: {
-      [defaultPackages.ipfsResolver]: {
+      [DefaultBundle.uriResolverExts[0].uri.uri]: {
```
krisbitney: Should we use a named URI here so we don't break the config if we change the order of the resolver extensions? I know we aren't planning to change the order anytime soon, but as a precaution

packages/cli/src/lib/test-env/client-config.ts
```diff
@@ -29,14 +27,21 @@ export function getTestEnvClientConfig(): Partial<BuilderConfig> {
...
-        fallbackProviders: defaultIpfsProviders,
+        fallbackProviders: DefaultBundle.ipfsProviders,
         retries: { tryResolveUri: 1, getFile: 1 },
       },
+      "proxy/testnet-ens-uri-resolver-ext": {
+        registryAddress: ensAddress,
+      },
+    },
+    redirects: {
+      "proxy/testnet-ens-uri-resolver-ext":
```
krisbitney, line 40: why do we need this proxy uri and redirect?

packages/js/client-config-builder/src/bundles/default.ts
```diff
@@ -0,0 +1,171 @@
...
+    package: ipfsResolver.wasmPackage,
+    source: Uri.from("ens/wraps.eth:async-ipfs-uri-resolver-ext@1.0.0"),
+  },
+};
+
+type UriResolverExtBootloader = [IDefaultEmbed, IUriRedirect, ...Uri[]];
+
+export const uriResolverExts: UriResolverExtBootloader = [
+  embeds.asyncIpfsResolver,
+  {
```
krisbitney, line 54: I think this redirect is a bit out of place. I ended up writing out my thought process weighing the pros and cons of this strategy:

**For ipfs uri and object**
I think there is an argument to be made for using the ipfs uri and avoiding the redirect altogether. This would probably coincide with using an object instead of an array for the resolver uris.

One point is that the redirect makes it a bit harder for users to copy the default config (as a basis for their own custom configuration) because they can't just copy the whole array like `...uriResolverExts`.

An advantage of using an object with named keys, instead of an array, is that it's more clear what you're accessing. A key named `ensTextRecordResolver` can indicate what the ipfs uri represents. And there is no need to worry about values changing order or anything like that. Users can always do `Object.values(...)` if they want an array.

**For ens uri and array**
On the other hand, the ens uri is going to be much nicer for debugging because the uri resolution history will contain the uri and not an object key. Seeing an ipfs hash instead of a named URI could confuse anyone who isn't familiar with the default config. 

Using an array also means we don't have to worry about key names. Refactoring the resolvers, or reordering them, will not generally lead to breaking changes. The exception would be changes to the ipfs resolver or ens text record resolver, since their order in the array is used.

krisbitney (dismissed):
Amazing work! This is a really exciting step for us. Using wrappers instead of plugins shows that we believe in our product.

packages/js/uri-resolver-extensions/src/ExtendableUriResolver.ts
dOrgJelli: Good catch, we just chatted via DM and agree we should just rename version `1.1.0` to `1.0.0` since the changes are benign, and makes sense to have within a v1 release.

packages/cli/src/lib/test-env/client-config.ts
dOrgJelli: yes that's a great point, will do!

packages/cli/src/lib/test-env/client-config.ts
dOrgJelli: So, this is a bit of "hack", but effectively we want to re-configure the `ens-uri-resolver-ext` with a new environment that uses the test network's ENS address. This was the only work around I could think of. Open to other potential options!

packages/js/client-config-builder/src/bundles/default.ts
dOrgJelli: Okay this all sounds super valid, but I'm having a bit of a hard time picturing what this would look like. Would you mind creating another PR on top of this work to show what you mean here? 🙏

packages/js/client-config-builder/src/bundles/default.ts
```diff
@@ -0,0 +1,171 @@
...
+  http: IDefaultPlugin;
+  fileSystem: IDefaultPlugin;
+  concurrent: IDefaultPlugin;
+  ethereumProvider: IDefaultPlugin;
+}
+
+export const plugins: IDefaultPlugins = {
+  logger: {
+    uri: Uri.from("plugin/logger@1.0.0"),
+    plugin: (loggerPlugin({}) as unknown) as PluginPackage<unknown>,
```
namesty, line 82: Minor nit: I don't think we should do this casting here; as it erases the factory function's return type to `unknown` and `unknown` can be then cast to anything (it's as good as using `any`). Since `IDefaultPlugin` already states the `plugin` property is of type `PluginPackage<unknown>`, I think doing just `plugin: loggerPlugin({})` should automatically coerce the result of the factory function correctly to `PluginPackage<unknown>`. Same goes for the other plugins below in this structure.

Another thought is that if the generic in `PluginPackage` disallows polymorphic plugin lists (as we need to deterministically know which plugin package to properly pass the correct generic) and we need to frequently cast the generic as unknown; we may wanna review that interface definition.

namesty (dismissed):

packages/js/client-config-builder/src/bundles/default.ts
dOrgJelli: Yes I agree @namesty, should have added a `// TODO: ` to these casts, we need to add `@polywrap/core-js` as a `peerDependency` within the plugin wrappers, to help with package -> type resolution in typescript. This will be removed.

packages/js/uri-resolver-extensions/src/ExtendableUriResolver.ts
```diff
@@ -23,31 +23,32 @@ export class ExtendableUriResolver extends UriResolverAggregatorBase<
...
   /**
    * Create an ExtendableUriResolver
    *
-   * @param extInterfaceUri - URI Resolver Interface URI
+   * @param extInterfaceUris - URI Resolver Interface URIs
    * @param resolverName - Name to use in resolution history output
    * */
   constructor(
-    extInterfaceUri: Uri = ExtendableUriResolver.extInterfaceUri,
+    extInterfaceUris: Uri[] = ExtendableUriResolver.extInterfaceUris,
```
cbrzn: nit: seeing that we have two variables with the same name (one static variable and the other instance variable) feels weird - should we define the static one as the default (explicitly adding something like `defaultExtInterfaceUris`)? iiuc, it's only used as a default value in the constructor

packages/cli/src/lib/test-env/client-config.ts
```diff
@@ -29,14 +27,21 @@ export function getTestEnvClientConfig(): Partial<BuilderConfig> {
 
   return {
     envs: {
-      [defaultPackages.ipfsResolver]: {
+      [DefaultBundle.embeds.asyncIpfsResolver.uri.uri]: {
```
cbrzn: ultra nit: i am not sure about having `asyncIpfsResolver`, I feel we should just have `ipfsResolver`, the `async` is an implementation detail that the client config shouldn't need to be aware of?

cbrzn (commented):
looks solid :moyai: 

wouldn't it be better if the `wasm` files are generated when the client is built? that way we don't push it (and need to re-build it if modified)? i am ok with both ways tho but just wondering if you thought about this

cbrzn:
also, how did you fix the wasm memory errors when resolving with many wrappers?

packages/js/uri-resolver-extensions/src/ExtendableUriResolver.ts
```diff
@@ -23,31 +23,32 @@ export class ExtendableUriResolver extends UriResolverAggregatorBase<
...
+  public static extInterfaceUris: Uri[] = [
+    Uri.from("wrap://ens/wraps.eth:uri-resolver-ext@1.1.0"),
+    Uri.from("wrap://ens/wraps.eth:uri-resolver-ext@1.0.0"),
+  ];
   // $end
 
   // $start: ExtendableUriResolver-extInterfaceUri
-  /** The active interface uri to which implementations should be registered */
-  public readonly extInterfaceUri: Uri;
+  /** The active interface uris to which implementations should be registered */
```
nerfZael: We should keep the casing consistent: URI

packages/js/uri-resolver-extensions/src/ExtendableUriResolver.ts
```diff
@@ -23,31 +23,32 @@ export class ExtendableUriResolver extends UriResolverAggregatorBase<
   Error
 > /* $ */ {
   // $start: ExtendableUriResolver-extInterfaceUri-static
-  /** The default interface uri to which implementations should be registered */
-  public static extInterfaceUri: Uri = new Uri(
-    "wrap://ens/uri-resolver.core.polywrap.eth"
-  );
+  /** The supported interface uris to which resolver-ext implementations should be registered */
```
nerfZael: We should keep the casing consistent: URIs

packages/js/uri-resolver-extensions/src/ExtendableUriResolver.ts
nerfZael: `defaultExtInterfaceUris` sounds good to me

packages/cli/src/lib/test-env/client-config.ts
nerfZael: I dont know how the asyncIpfsResolver is implemented,but, ideally, it'd use an ipfsResolver under the hood.
That way the logic about ipfs resolution could be contained in that wrapper while you can build other resolvers on top.
E.g someone could implement a parallelIpfsResolver (that uses workers) and then users could choose which wrapper to use

That all said, we could just have a polywrapIpfsResolver where async resolution is just an implementation detail

packages/js/uri-resolver-extensions/src/ExtendableUriResolver.ts
nerfZael: Also, while I like the ability for ExtendableUriResolver to use multiple interfaces, I wonder why this was required right now (what changed in the new interface version).
The extendable resolver follows the resolver extension interface and if the interface changes then it should change, but in this case it didn't?

packages/cli/src/lib/test-env/client-config.ts
krisbitney: Yeah, that totally makes sense. I think we used to configure this client using the same URI that was used for the ENS resolver. Are we able to do the same here and use the `ens/wraps.eth:ens-uri-resolver-ext@1.0.0` URI?

dOrgJelli:
> wouldn't it be better if the `wasm` files are generated when the client is built? that way we don't push it (and need to re-build it if modified)? i am ok with both ways tho but just wondering if you thought about this

@cbrzn yes definitely, I think this should be something that is done through the CLI. For now it's just done within this custom script, but this should be easier to do with a single command.

packages/js/uri-resolver-extensions/src/ExtendableUriResolver.ts
dOrgJelli: There was a minor modification to the interface that was made, which caused the version to be bumped to `1.1.0`.

The change was that each method was given an optional environment, see here: https://wrappers.io/v/ipfs/QmSAXrSLcmGUkQRrApAtG5qTPmuRMMX2Zf1wihpguDQfbm

This was discussed up here: https://github.com/polywrap/toolchain/pull/1582#discussion_r1121424976

So we have to account for old & new interface URIs that wrappers could have implemented, since both will work as there are no breaking changes being made.

dOrgJelli: Effectively it's just backwards compatibility.

packages/cli/src/lib/test-env/client-config.ts
dOrgJelli: The IPFS resolver implementations can be found here: https://github.com/polywrap/uri-resolver-extensions/tree/master/implementations/ipfs

While I agree in principle that async should use sync resolver under-the-hood, it's easier to just build them as 2 separate wrappers as it helps conserve memory.

I agree with simplifying the naming @cbrzn so happy to make this change.

krisbitney: The async IFPS resolver uses the IPFS HTTP client wrapper. The IPFS resolver wrapper implements the URI resolver interface, while the IPFS HTTP client is a more general IPFS HTTP client utility. 

This is somewhat necessary due to the way our concurrency feature works. The concurrency interface and plugin are designed for concurrent invocations. That implies there must be two wrappers: one that implements the Uri Resolver interface and another that is called with concurrency. The initial call to the URI Resolver cannot be parallel, because it must work the same way that other URI resolvers do. The call to the IPFS `cat` function is what needs to be parallelized. 

The reason it has `async` in the name is because there is a synchronous IPFS resolver that was made for use in clients that don't yet have concurrency plugins available. We can absolutely change the name in the default config (and elsewhere). That makes sense to me!

dOrgJelli: Also we could certainly implement different ways of parallelizing the ipfs resolution, but would leave this for the future on an as-need basis.

dOrgJelli: benefit of using wrappers is they're super easy to swap out, just change the URI :D

krisbitney: The IPFS resolver wrapper could call itself recursively, and it could all be done with one wrapper. But it would require a hard-coded URI in the wrapper. Let me know what you think!

dOrgJelli: > The IPFS resolver wrapper could call itself recursively, and it could all be done with one wrapper. But it would require a hard-coded URI in the wrapper. Let me know what you think!

This has been a feature that's been requested a lot, allowing wrappers to know about their URI. IIRC @Niraj-Kamdar @nerfZael had a plan for how to implement this in wrap 0.2?

dOrgJelli:
> also, how did you fix the wasm memory errors when resolving with many wrappers?

@cbrzn upgrading to node > 17 helped a lot. In previous versions of node, `WebAssembly.Memory` instances aren't freed from memory very aggressively, leading to the heap blowing up over time. In order to help with memory related problems in the browser, we have plans to add a `limits` section to the client config, so that we can limit (for example) the number of parallel invocations of a given wrapper, effectively load-balancing invocations.

packages/js/uri-resolver-extensions/src/ExtendableUriResolver.ts
dOrgJelli: will fix 👌

krisbitney (approved):

packages/js/client-config-builder/src/bundles/default.ts
dOrgJelli: Hey @namesty this is fixed here, had to update all plugins to properly use `peerDependency`s: https://github.com/polywrap/toolchain/pull/1599

Helpful reading: https://nodejs.org/es/blog/npm/peer-dependencies/

PR #1584, opened 2023-02-23
CLI - codegen command now supports the `--watch` option

### LINKED ISSUE #1459: Implement a polywrap codegen --watch  CLI command

pileks:
Resolves #1459

You can now run `npx polywrap codegen --watch` which will automatically watch files within your project directory and re-run codegen whenever changes are detected.


packages/cli/src/commands/codegen.ts
```diff
@@ -139,12 +144,37 @@ async function run(options: Required<CodegenCommandOptions>) {
         project,
       });
 
-  const result = await codeGenerator.generate();
+  const execute = async (): Promise<boolean> => {
+    let result = false;
+
+    try {
```
krisbitney, line 150: There is already a try-catch block within `codeGenerator.generate()` that catches and logs any errors, so the try-catch block in `codegen.ts` will log the errors a second time. 

https://github.com/polywrap/toolchain/blob/origin-dev/packages/cli/src/lib/codegen/CodeGenerator.ts

krisbitney (changes requested):
This is a great feature! Very cool stuff. 😁

packages/cli/src/commands/codegen.ts
dOrgJelli: Hmmm, this try/catch logic is also in the build command. Honestly I think it's alright to log more than once, and it might be best to remove the try/catch from within the `lib` layer. Thoughts?

dOrgJelli (approved):
Looks great! Waiting for @krisbitney's approval regarding logging practices.

packages/cli/src/commands/codegen.ts
krisbitney: Yeah, either way is good. I prefer removing the try-catch from the lib layer if there are no side effects from doing that, but I'd also be happy merging this as-is.

krisbitney (approved):

PR #1586, opened 2023-02-27
PackageToWrapperCacheResolver refactor and resolver tests

nerfZael:
This PR separates the PackageToWrapperCacheResolver into two resolvers:
- PackageToWrapperResolver
- WrapperCacheResolver
This accomplishes the separation of concerns between turning packages into wrappers and actually caching the wrappers. With this PR the resolvers can be used separately and combined with other resolvers.
 
 In addition to the refactor, this PR removes all except a couple of tests from the client URI resolution set.
 The remaining tests are:
 - `sanity` (which serves to confirm that the default PolywrapClient configuration resolves URIs as expected)
 - `can resolve uri with custom resolver` (which serves to confirm that the default PolywrapClient configuration works with custom resolvers)
 
 Other tests have been either removed, since they're no longer applicable (such as `can resolve uri with custom resolver at invoke-time`) or moved into unit tests in the `uri-resolvers-js` and `uri-resolver-extensions-js` repo.
 Additional unit tests have also been added to those repos.
 
 The tests also exposed an issue with the WrapperCacheResolver (as well as the previous PackageToWrapperCacheResolver) not caching the final resolved URI, since it only cached the resolution path which does not contain it.
 This issue has also been fixed in this PR.


packages/js/uri-resolver-extensions/examples/quickstart.ts
```diff
@@ -27,17 +25,14 @@ export function example(): CoreClientConfig {
...
-          StaticResolver.from([
-            ...redirects,
-            ...wrappers,
-            ...packages,
-          ]),
-          new ExtendableUriResolver(),
-        ],
-        new WrapperCache()
-      )
+      [
```
pileks, line 28: Considering this is an example (and used within the README), are you missing something like the following here?

```typescript
WrapperCacheResolver.from(
  PackageToWrapperResolver.from([
    //...
  ])
)
//...
```

pileks (changes requested):

yarn.lock
```diff
@@ -2171,14 +2171,22 @@
...
-"@polywrap/file-system-plugin-js@~0.10.0-pre":
+"@polywrap/file-system-plugin-js@0.10.0-pre.10", "@polywrap/file-system-plugin-js@~0.10.0-pre":
   version "0.10.0-pre.10"
   resolved "https://registry.yarnpkg.com/@polywrap/file-system-plugin-js/-/file-system-plugin-js-0.10.0-pre.10.tgz#93e796d4c25203f05605e7e36446facd6c88902d"
   integrity sha512-rqiaHJQ62UoN8VdkoSbpaI5owMrZHhza9ixUS65TCgnoI3aYn3QnMjCfCEkEiwmCeKnB9YH/0S2+6NWQR17XJA==
   dependencies:
     "@polywrap/core-js" "0.10.0-pre.10"
     "@polywrap/plugin-js" "0.10.0-pre.10"
 
+"@polywrap/fs-resolver-plugin-js@0.10.0-pre.10":
```
dOrgJelli: Not sure how this made its way it, this plugin should be deprecated, and the file-system-uri-resolver-ext should be used in its place.

packages/js/uri-resolvers/src/helpers/PackageToWrapperResolver.ts
```diff
@@ -0,0 +1,97 @@
+import { UriResolutionResult, UriResolver, UriResolverLike } from ".";
```
dOrgJelli: Why is this resolver within the "helpers" folder? Seems like ti should be within a "resolvers" folder along with the others, no?

packages/js/uri-resolvers/src/helpers/PackageToWrapperResolver.ts
```diff
@@ -0,0 +1,97 @@
...
+   * @param client - a CoreClient instance that may be used to invoke a wrapper that implements the UriResolver interface
+   * @param resolutionContext - the current URI resolution context
+   * @returns A Promise with a Result containing either a wrap package, a wrapper, or a URI if successful
+   */
+  async tryResolveUri(
+    uri: Uri,
+    client: CoreClient,
+    resolutionContext: IUriResolutionContext
+  ): Promise<Result<UriPackageOrWrapper, TError | Error>> /* $ */ {
+    // Resolve uri if not in cache
```
dOrgJelli: I'm a bit confused by this, when are we looking in the cache here? If this resolver doesn't "know" about a cache, then I don't think we should make mention of it within this comment.

packages/js/uri-resolvers/src/cache/WrapperCacheResolver.ts
```diff
@@ -8,57 +8,49 @@ import {
...
-import { DeserializeManifestOptions } from "@polywrap/wrap-manifest-types-js";
 import { Result } from "@polywrap/result";
 
-// $start: PackageToWrapperCacheResolver
+// $start: WrapperCacheResolver
 /**
  * An IUriResolver implementation that caches wrappers once they are resolved.
- * The PackageToWrapeprCacheResolver wraps one or more IUriResolver
- * implementations and delegates resolution to them.
+ * It does not cache URIs and packages.
```
dOrgJelli: Can you explain further here why this is? Maybe adding something like "the caching or URIs and packages is a job for parent resolvers" or something like that.

packages/js/uri-resolvers/src/helpers/PackageToWrapperResolver.ts
```diff
@@ -0,0 +1,97 @@
...
+  implements IUriResolver<TError | Error> /* $ */ {
+  // $start: PackageToWrapperResolver-constructor
+  /**
+   * Creates a PackageToWrapperResolver
+   *
+   * @param _resolver - a resolver to delegate resolution to
+   * @param _options - control wrapper manifest deserialization
+   * */
+  constructor(
+    private _resolver: IUriResolver<TError>,
```
dOrgJelli: NIT: I think we should start using names that convey the hierarchy of resolvers that are being created here. For example, `_resolver` could be renamed to `_innerResolver` or `_childResolver`.

packages/js/uri-resolvers/src/helpers/PackageToWrapperResolver.ts
```diff
@@ -0,0 +1,97 @@
...
+  IUriResolutionContext,
+  UriPackageOrWrapper,
+} from "@polywrap/core-js";
+import { DeserializeManifestOptions } from "@polywrap/wrap-manifest-types-js";
+import { Result } from "@polywrap/result";
+
+// $start: PackageToWrapperResolver
+/**
+ * An IUriResolver implementation that initalizes wrappers from resolved packages.
+ * The PackageToWrapperResolver wraps an IUriResolver implementation and delegates resolution to it.
```
dOrgJelli: NIT: there are a lot of resolvers that "wrap" 1 or more inner resolvers. This functionality could be put into a base-class to avoid duplicate code. Might be more work than it's worth though, as we don't want to over complicate this.

packages/js/uri-resolvers/src/cache/WrapperCacheResolver.ts
```diff
@@ -8,57 +8,49 @@ import {
...
   /**
-   * Create a PackageToWrapperCacheResolver
+   * Creates a WrapperCacheResolver
    *
    * @param _resolverToCache - a resolver to delegate resolution to
    * @param _cache - a wrapper cache
-   * @param _options - control wrapper manifest deserialization
    * */
   constructor(
     private _resolverToCache: IUriResolver<TError>,
```
dOrgJelli: NIT: this naming is confusing to me... I would prefer if this was named `_resolver`, `_innerResolver`,  `_subResolver`, or `_childResolver`.

packages/js/uri-resolvers/src/cache/WrapperCacheResolver.ts
```diff
@@ -8,57 +8,49 @@ import {
...
-   * @returns a PackageToWrapperCacheResolver
+   * @returns a WrapperCacheResolver
    * */
   static from<TResolverError = unknown>(
     resolver: UriResolverLike,
-    cache: IWrapperCache,
-    options?: { deserializeManifestOptions?: DeserializeManifestOptions }
-  ): PackageToWrapperCacheResolver<TResolverError> /* $ */ {
-    return new PackageToWrapperCacheResolver(
+    cache: IWrapperCache
```
dOrgJelli, line 46: We should create a "local storage" `IWrapperCache` implementation, which could cache wrappers in the browser and desktop application!

packages/js/uri-resolvers/src/cache/WrapperCacheResolver.ts
```diff
@@ -82,55 +74,36 @@ export class PackageToWrapperCacheResolver<TError>
       resolutionContext.trackStep({
```
dOrgJelli, line 75: I think the logic for "maintaining" a valid context could be something that's "lifted" into a base class. This way resolvers don't have to be "trusted" to properly manipulate the context, and can simply worry about the resolution logic they're responsible for.

packages/js/uri-resolvers/src/cache/WrapperCacheResolver.ts
```diff
@@ -82,55 +74,36 @@ export class PackageToWrapperCacheResolver<TError>
...
-        result = UriResolutionResult.ok(result.value.uri, wrapper);
-      } else if (result.value.type === "wrapper") {
-        const wrapper = result.value.wrapper;
-        const resolutionPath: Uri[] = subContext.getResolutionPath();
-
-        for (const uri of resolutionPath) {
-          await this._cache.set(uri, wrapper);
-        }
+      for (const uri of resolutionPath) {
+        await this._cache.set(uri, wrapper);
```
dOrgJelli, line 97: We've talked about this before, but aren't we loosing information here about the URI resolution chain?

For example, we're turning `A->B->C->Wrapper` into `A->Wrapper, B->Wrapper, C->Wrapper`, so we no longer know about how `A B C` are linked.

packages/js/uri-resolvers/src/__tests__/wrapper-resolver/histories/can-resolve-a-wrapper.json
```diff
@@ -0,0 +1,3 @@
+[
+  "wrap://test/wrapper => Wrapper (wrap://test/wrapper) => wrapper (wrap://test/wrapper)"
```
dOrgJelli, line 2: Love these smaller more specific test files! 👏

dOrgJelli (dismissed):
Very minor NITs, everything looks great 👏

packages/js/uri-resolvers/src/__tests__/wrapper-resolver/histories/can-resolve-a-wrapper.json
nerfZael: Much easier to maintain and they even caught a minor bug!

packages/js/uri-resolvers/src/cache/WrapperCacheResolver.ts
nerfZael: Yep, this PR doesn't change anything regarding that and leaves it for the other two issues we have in origin-qa to tackle

packages/js/uri-resolvers/src/cache/WrapperCacheResolver.ts
nerfZael: Good idea! I created an issue to track this: https://github.com/polywrap/toolchain/issues/1625

packages/js/uri-resolvers/src/helpers/PackageToWrapperResolver.ts
nerfZael: It's a leftover comment from the PackageToWrapperCacheResolver, will remove

yarn.lock
nerfZael: This PR was before the deprecation, I'll update it

packages/js/uri-resolver-extensions/examples/quickstart.ts
nerfZael: No? This is an example for resolver extensions. It's independent from caching (which was IMO mistakenly included here before)

packages/js/uri-resolvers/src/helpers/PackageToWrapperResolver.ts
nerfZael: Moved it to the `packages` directory

packages/js/uri-resolvers/src/cache/WrapperCacheResolver.ts
nerfZael: IMO, it is the responsibility of the resolver to not only apply resolution logic, but to build a resolution context matching that resolution logic, I'm not sure we can separate it in most cases.
That said, if you have a good example how this can work in a base class (for all use cases) then we can create an issue for that to be solved in another PR

dOrgJelli: Okay I can understand that perspective, but I'm still not 100% clear on what the different cases we want to support here. I was thinking they'd all be the same, but I can understand if that's not the case.

packages/js/uri-resolvers/src/cache/WrapperCacheResolver.ts
dOrgJelli: Okay great, nice to know we're tracking this issue elsewhere :pray:

dOrgJelli (dismissed):

PR #1592, opened 2023-03-01
Add `primaryJob` to deploy manifest and output `URI.txt` when running `polywrap deploy`

krisbitney:
This PR adds a new option to the deploy manifest, `primaryJob`. The option identifies the name of the job that is used for the primary deployment. The URI of the final step of the primary job is printed to `polywrap.deployment.txt`, which is output to the folder that contains the deploy manifest.

The primary deployment URI is also recorded in a historical log, with a timestamp, located at the `.polywrap/deploy/deploy.log`.

This PR enables the UX described in https://github.com/polywrap/uri-resolver-extensions/pull/19


packages/js/manifests/polywrap/src/formats/polywrap.deploy/0.3.0.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export interface DeployManifest {
+  /**
+   * Polywrap deployment manifest format version.
+   */
+  format: "0.3.0";
+  /**
+   * Name of the job that runs the primary deployment sequence.
+   */
+  primaryJob?: string;
```
pileks: Would it make more sense to call this `primaryJobName` for the sake of clarity? Just a nitpick!

packages/cli/src/lib/defaults/infra-modules/eth-ens-ipfs/docker-compose.yaml
```diff
@@ -1,6 +1,7 @@
 version: '3'
 services:
   ganache:
+    platform: linux/x86_64
```
pileks: Please drop this, as `ganache` doesn’t require this by default, while it causes trouble inside WSL.
For more details see: https://github.com/polywrap/toolchain/pull/1581

pileks (changes requested):
Overall looks good! I’d love to see the `platform` dropped from the docker compose file as it was causing issues for me.

One thing to note - Instead of having the `deploy` output text file be in the same folder as the deploy manifest, we should have a `deployments` folder inside the `.polywrap` folder, with each deployment output being recorded in a datetime-named file, similar to how we export logs using the `-l, --log-file` option.
Also, it would be prudent to let the user know that a file has been output by logging something in the lines of “The output of the primary job has been logged into “./polywrap/deploys/primary-job-20230301122234.txt” or however we name them.

We’d have historic logs inside a folder that can be safely deleted and is gitignored by default, which could prove useful in the future.

packages/cli/src/lib/defaults/infra-modules/eth-ens-ipfs/docker-compose.yaml
krisbitney: Yes, 100%! Adding it back in was a mistake. I didn't mean to commit it.

I was using it because I am actually having issues without it: https://github.com/polywrap/toolchain/pull/1581#issuecomment-1446268122

packages/js/manifests/polywrap/src/formats/polywrap.deploy/0.3.0.ts
krisbitney: Good call!

krisbitney:
> Overall looks good! I’d love to see the `platform` dropped from the docker compose file as it was causing issues for me.
> 
> One thing to note - Instead of having the `deploy` output text file be in the same folder as the deploy manifest, we should have a `deployments` folder inside the `.polywrap` folder, with each deployment output being recorded in a datetime-named file, similar to how we export logs using the `-l, --log-file` option. Also, it would be prudent to let the user know that a file has been output by logging something in the lines of “The output of the primary job has been logged into “./polywrap/deploys/primary-job-20230301122234.txt” or however we name them.
> 
> We’d have historic logs inside a folder that can be safely deleted and is gitignored by default, which could prove useful in the future.

I like the idea of keeping a historic log. 

The deployment file should ideally be in the root of the repo. The reason for adding is so that it can be retrieved by a URI resolver extension wrapper. That way users can import a wrapper from a URI like `github.com/polywrap/toolchain`. Ideally, users will soon be able to copy an past a repo URL from their browser into their schema and it will work without modification. That's what I'm working towards at the moment.

krisbitney:
To elaborate, we want people to commit the `polywrap.deployment.txt` file. If it's in the `.polywrap` folder, it will be ignored.

krisbitney:
I added a historical deployment log stored at `./polywrap/deploy/deploy.log`. Now "primary deployment" URIs are timestamped and recorded there.

pileks (dismissed):

packages/js/client-config-builder/src/bundles/getDefaultConfig.ts
```diff
@@ -34,6 +34,7 @@ export const defaultWrappers = {
   graphNode: "wrap://ens/wraps.eth:graph-node@1.0.0",
   ensTextRecordResolver:
     "wrap://ipfs/QmbqeVAhSzTtSmdVjrPMK42pX1sFs8t5MUB741T7nxSs1p",
+  githubResolver: "wrap://ipfs/QmTNukpDR6HHtZnUKX88ytJTfqRaeLoy6bpDLRXAQX42sW",
```
dOrgJelli: This should probably be done within another PR

packages/cli/src/lib/deploy/Deployer.ts
```diff
@@ -34,14 +34,14 @@ export class Deployer {
     deployModulesDir: "modules/",
   };
   public manifest: DeployManifest;
-  private _config: DeployerConfig;
+  public readonly config: DeployerConfig;
```
dOrgJelli: Instead of making the config public, requiring us to "lock down" the structure by adding readonly specifiers, I think we should instead just expose getter methods to the `Deployer` class for these internal config properties.

packages/cli/src/commands/deploy.ts
```diff
@@ -63,8 +64,47 @@ async function run(options: Required<DeployCommandOptions>): Promise<void> {
...
+  // get deployment uri
+  const primaryJob = jobResults.find(
+    (job) => job.name === primaryJobName
+  ) as DeployJobResult;
+  const deploymentStep = primaryJob.steps.length - 1;
+  const deploymentUri = primaryJob.steps[deploymentStep].result.uri;
+
+  // write deployment uri to file
+  const manifestDir = path.dirname(manifestFile);
+  const deploymentFilepath = path.join(manifestDir, "polywrap.deployment.txt");
```
dOrgJelli: I'm not a fan of using the `polywrap` name here, I think we should reserve the `polywrap` file name for toolchain configuration files. Similarly, we reserve the `wrap` file name for wrap standard files.

This file to me is super similar to GitHub page's "CNAME" file, which redirects to the external domain the page is being hosted at. Here we're redirecting to an external URI where the package is being hosted at.

It's for this reason that I'd be preferable to both the names `WRAP` and `URI`, where they're simple text files containing 1 line of text, which is the URI to be used for wrapper resolution.

Thoughts?

dOrgJelli (changes requested):

packages/js/client-config-builder/src/bundles/getDefaultConfig.ts
krisbitney: Yes, for sure. This is a mistake. I initially added it but then I thought I removed it in this commit https://github.com/polywrap/toolchain/pull/1592/commits/97df57a58df49d4a12f7d2700094842305f174e7 . But it looks like I only removed it from the uri resolver ext implementations array

krisbitney: fixed!

packages/cli/src/commands/deploy.ts
krisbitney: I renamed it to `URI.txt`

krisbitney: and yeah, that all makes sense and sounds good!

krisbitney: let me know if you actually preferred `WRAP.txt`. It's an easy change to make now, but later it will be harder if we start using the github resolver

packages/cli/src/lib/deploy/Deployer.ts
krisbitney: yeah, that makes way more sense. thanks!!

dOrgJelli (approved):

pileks (approved):

PR #1593, opened 2023-03-01
Fix: display decoded args when errors are thrown in subinvocations

krisbitney:
Args used to be displayed as-is in error messages when errors were thrown in Wasm wrapper invocations. This meant args for subinvocations were byte arrays. Now the args are always decoded for errors so they are human-readable.


cbrzn (approved):
awesome :)

dOrgJelli (approved):
👏

PR #1594, opened 2023-03-02
replace UriResolverExtBootLoader array with object

krisbitney:


krisbitney:
PR was just to show something. It was never intended to be merged.

PR #1595, opened 2023-03-02
Fix: ganache now works properly on Apple CPU

krisbitney:
Fixed by using full node image instead of node-alpine in ganache Docker image file


pileks (approved):
Tested locally on WSL, working as intended!
:ship: it!

dOrgJelli (approved):

PR #1601, opened 2023-03-06
refactor(plugin-js): move env to plugin method params

Niraj-Kamdar:


dOrgJelli (changes requested):
@Niraj-Kamdar how can we make this a change backwards-compatible? We should also have tests to ensure this the case.

PR #1612, opened 2023-03-08
fix: align @types/node package version, fixing infra command in monorepo

dOrgJelli:
This should fix CI, where the CLI's `infra` command is failing.


dOrgJelli:
I'm really not sure why the CLI tests have an exit code of non-zero after finishing all tests. There is no error output to help debug this, will dig into in the future. Going to merge this PR as it fixes another test suite.

PR #1618, opened 2023-03-10
Fix: --no-codegen option is now recognized in `polywrap build`

krisbitney:
fix for issue where `noCodegen` option name is automatically changed to `codegen` by Commander package in build command

### DIFF
--- a/packages/cli/src/__tests__/e2e/build.plugin.spec.ts
+++ b/packages/cli/src/__tests__/e2e/build.plugin.spec.ts
@@ -35,7 +35,6 @@ describe("e2e tests for build command - plugin project", () => {
       test(testCaseName, async () => {
         const { exitCode: code, stdout: output, stderr: error } = await Commands.build({
           ...args,
-          codegen: true,
         }, {
           cwd: testCaseDir,
         });
--- a/packages/cli/src/__tests__/e2e/build.wasm.spec.ts
+++ b/packages/cli/src/__tests__/e2e/build.wasm.spec.ts
@@ -192,7 +192,6 @@ describe("e2e tests for build command", () => {
     const logFileAbsPath = path.join(testCaseDir, logFilePath);
     const { exitCode: code } = await Commands.build({
       logFile: logFilePath,
-      codegen: true
     }, {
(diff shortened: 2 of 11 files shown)


pileks (dismissed):
Let's get this merged ASAP, as it also fixes the IPFS hash issue! :D

Full disclosure: I was the one that changed this `codegen` stuff, and I left it as such since, with Commander, specifying a `--codegen` flag also adds the option of `--no-codegen` as an alternative flag. I did exclude the `noCodegen` option for the CLI-JS, as I didn't think of it at the time. Good catch @krisbitney!

cbrzn (dismissed):

pileks (dismissed):
Bravo @cbrzn!

cbrzn:
i've (in theory) fixed all tests except for `create app typescript` command. 

this last test, is failing in this line https://github.com/polywrap/toolchain/blob/kris/fix-codegen-option-recognition/packages/cli/src/lib/project/templates/generateProjectTemplate.ts#L142-L149 and the reason is that we've done a change in `.pre-*` in the app templates (now we only have `typescript`). and it is pulling the latest in npm (which points to `0.9.6`); hence, it is trying to copy a folder that doesn't exist.

I've seen this problem before, but I do not remember what fixed it... Also, it's worth mentioning that this error is **totally** unrelated to this PR, and it was introduced two weeks ago in this PR https://github.com/polywrap/toolchain/pull/1529 (`origin-dev` has been broken since then).

i think a hotfix might be to hard code the current version of [`@polywrap/templates`](https://github.com/polywrap/toolchain/blob/kris/fix-codegen-option-recognition/packages/cli/src/lib/project/templates/generateProjectTemplate.ts#L96) (we can just do `@polywrap/templates@0.10.0-pre.11` and add a todo to remove it once `0.10` is relased), but I feel this is no bueno tho - open to any suggestions :smile: 

@krisbitney @pileks

dOrgJelli (approved):

dOrgJelli:
Thank you @krisbitney! Let's fix the `create ...` CLI test in another PR, no need to hold this up longer.

PR #1638, opened 2023-03-22
Resolver extension tests

nerfZael:
This PR adds no extra functionality.
It's focus is on covering different use cases of resolver extensions with tests.
The following tests are implemented:
- can resolve URI with plugin extension
- can resolve a package with plugin extension
- shows the plugin resolver extension error
- does not resolve a URI when not a match with plugin extension
- can resolve URI with wasm extension
- can resolve a package with wasm extension
- shows the wasm resolver extension error
- does not resolve a URI when not a match with wasm extension
- does not cause infinite recursion when resolved at runtime when an extension is not found


pileks (approved):
LGTM!

dOrgJelli (approved):

PR #1684, opened 2023-04-13
Fix: Map type argument parsing for `WrapError`

### LINKED ISSUE #1637: WrapError: Map types arguments are not parsed correctly
**Describe the bug**
When an invocation fails and `WrapError` is returned, any argument that is a `map` type, is shown as `{}`.

the reason of this is that `null` is being passed as the second argument of the `JSON.stringify` that is being called in [the return of the `WrapError` arguments ](https://github.com/polywrap/toolchain/blob/origin-dev/packages/js/wasm/src/WasmWrapper.ts#L225).

This has happened before in the workflows, and this PR fixed it: https://github.com/polywrap/toolchain/pull/1438/files

I think a fix for this would be moving the `typesHandler` method to `core-js` and importing it in the cli & wasm-js package

krisbitney:
I moved the typesHandler method to core-js and imported it in the cli & wasm-js packages. I also added unit tests for it.

Closes https://github.com/polywrap/toolchain/issues/1637

### DIFF
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -16,10 +16,10 @@ import {
   parseManifestFileOption,
   parseLogFileOption,
   parseWrapperEnvsOption,
-  typesHandler,
 } from "../lib";
 import { createLogger } from "./utils/createLogger";
 
+import { typesHandler } from "@polywrap/core-js";
 import path from "path";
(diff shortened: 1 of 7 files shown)


cbrzn (approved):
thanks for pushing this ser :pray:

dOrgJelli (approved):

PR #1718, opened 2023-04-23
fix: enum in python plugin bindings

Niraj-Kamdar:


dOrgJelli (approved):
LGTM! But we really need to add support for managing these app/plugin/wasm bindings outside of the toolchain so we don't have to keep rev'ing versions of the CLI. Will try to work on that ASAP.

PR #1732, opened 2023-05-03
feat: add msgpack bindings for base types for wasm-as

Niraj-Kamdar:


packages/schema/bind/src/bindings/assemblyscript/wasm/templates/msgpack/uint32-ts.mustache
```diff
@@ -0,0 +1,20 @@
+{{> serialization_imports}}
+
+export class UInt32 {
```
dOrgJelli, line 3: Why is this being added directly to the template, and not re-exported from the `@polywrap/wasm-as` package?

packages/test-cases/cases/bind/sanity/output/wasm-as/msgpack/uint32.ts
```diff
@@ -0,0 +1,31 @@
...
+  WriteSizer,
+  WriteEncoder,
+  Box,
+  BigInt,
+  BigNumber,
+  JSON,
+  Context
+} from "@polywrap/wasm-as";
+
+export class UInt32 {
```
dOrgJelli, line 14: All of this new code should have tests added (as-pect tests).

dOrgJelli (changes requested):
Also I think we should use these new functions within the rest of the serialization bindings. This will help serve as an extra stress-test to make sure they all function as expected, and it will reduce the size of the bindings.

dOrgJelli:
This PR has grown pretty stale, going to close for now

PR #1735, opened 2023-05-09
Docs manifest 0.1.0

pileks:
This PR introduces the Documentation Manifest, an extension manifest for the WASM/Interface project manifests.

An example documentation manifest:
```yaml
format: 0.1.0
title: A test wrap docs
logo: ./docs/logo2.png
pages: # the keys of this object are used as URI slugs for the readmes
  home:
    title: Home page
    path: ./docs/pages/first.md
  second-page:
    title: Second page
    path: ./docs/pages/second.md
homePage: second-page # we can specify any page to be our landing page
examples: # the keys of this object are used as URI slugs for the examples
  foo:
    title: Foo
    steps:
      - uri: $$WRAP_URI
        method: sampleMethod
        args:
          arg: Foo 1
        description: This is a foo 1 on URI $$WRAP_URI
      - uri: $$WRAP_URI
        method: sampleMethod
        args:
          arg: $0.value.result
        description: This is a foo 2 description $0.value.result value injection!
```

I have decided against merging workflow jobs and doc examples at this time for the following reasons:
- Workflows don't have the concept of a "primary job" in the same way deploy manifests do. Adding this just to satisfy the needs of the docs manifest seems counter-intuitive
  - [ ] Now that I say it out loud, adding a `jobName` field to the workflow might work. Thoughts? (Still doesn't answer other points)
- There is no real use for `step.description` in Workflows themselves, it would be a Docs addition to a manifest that doesn't belong to Docs
- Easier overall to work on them separately first - we can decide to merge them later or even diverge if we see it go that way

If `polywrap.docs.yaml` is present and mentioned in `exts` of the Project Manifest, `polywrap build` outputs all files related to documentation inside a `docs` directory within the build output dir.

This directory can then be consumed by `wrap-docs`, as is described in the following PR:
https://github.com/polywrap/wrap-docs/pull/8


packages/js/manifests/polywrap/src/formats/polywrap/migrators/0.3.0_to_0.4.0.ts
```diff
@@ -0,0 +1,15 @@
+import { PolywrapManifest as OldManifest } from "../0.3.0";
+import { PolywrapManifest as NewManifest } from "../0.4.0";
+
+import { ILogger } from "@polywrap/logging-js";
+
+export function migrate(migrate: OldManifest, logger?: ILogger): NewManifest {
+  logger?.info(
+    `A new Docs manifest extension is available! Use 'polywrap docs init' to set up documentation for your Wrap.`
```
krisbitney, line 8: Nice touch!

krisbitney:
I think the way you handled the docs examples is great. The execution of the examples can use same backend as `polywrap.test.yaml` workflows, right? If not, we should generalize the back end so it can run both `polywrap.test.yaml` examples and `polywrap.docs.yaml` workflows. Essentially, the docs manifest lets users define a workflow with metadata.

An alternative would be to only put the metadata in the docs manifest and ask users to provide a link to their workflow file, but I think the UX is better the way you've done it.

krisbitney (dismissed):

krisbitney:
One thought - I think users should also be able to reference assets stored in the `resources` directory, if that's not already possible. That way users never feel a need to have duplicate copies of files. Or maybe the `docs` directory should be stored in the `resources` folder? I'm not sure.

That kind of feature could come in a future PR

pileks:
> I think the way you handled the docs examples is great. The execution of the examples can use same backend as `polywrap.test.yaml` workflows, right? If not, we should generalize the back end so it can run both `polywrap.test.yaml` examples and `polywrap.docs.yaml` workflows. Essentially, the docs manifest lets users define a workflow with metadata.
> 
> An alternative would be to only put the metadata in the docs manifest and ask users to provide a link to their workflow file, but I think the UX is better the way you've done it.

This was actually the original idea!
I've detailed out the reasons against going this path in the description, and it's something that I'll be keeping in mind for the future!

> One thought - I think users should also be able to reference assets stored in the `resources` directory, if that's not already possible. That way users never feel a need to have duplicate copies of files. Or maybe the `docs` directory should be stored in the `resources` folder? I'm not sure.
> 
> That kind of feature could come in a future PR

This is not a bad idea at all. I'll see that we have something like this added.

krisbitney (dismissed):

dOrgJelli (changes requested):
I think we should remove the "pages" entry in the manifest, to help simplify this new feature. Let's only keep the features we need for initial MVP of wrapscan.

To me this is:
- description
- logo
- repository
- website
- examples

packages/js/manifests/polywrap/src/formats/polywrap.app/validate.ts
```diff
@@ -12,6 +12,7 @@ import {
 import AppManifestSchema_0_1_0 from "@polywrap/polywrap-manifest-schemas/formats/polywrap.app/0.1.0.json";
 import AppManifestSchema_0_2_0 from "@polywrap/polywrap-manifest-schemas/formats/polywrap.app/0.2.0.json";
 import AppManifestSchema_0_3_0 from "@polywrap/polywrap-manifest-schemas/formats/polywrap.app/0.3.0.json";
+import AppManifestSchema_0_4_0 from "@polywrap/polywrap-manifest-schemas/formats/polywrap.app/0.4.0.json";
```
dOrgJelli: Why are we adding version 0.4.0 of app, plugin, and wrap manifests? I don't see the need.

packages/manifests/polywrap/formats/polywrap.docs/0.1.0.json
```diff
@@ -0,0 +1,104 @@
...
+                "type": "object",
+                "additionalProperties": false,
+                "required": ["uri", "method"],
+                "properties": {
+                  "description": {
+                    "type": "string",
+                    "description": "A description of the step rendered above the invocation code."
+                  },
+                  "uri": {
+                    "description": "URI of the wrap being invoked.",
```
dOrgJelli, line 59: I would make this URI field optional. If it isn't specified it is assumed it is a self-invocation. If it is it will be used.

No need to introduce a special "$$WRAP_URI" or "$$THIS_URI", just leave it empty.

dOrgJelli (changes requested):

packages/manifests/polywrap/formats/polywrap.docs/0.1.0.json
pileks: One thing we can't do without `$$WRAP_URI` is add it to the `description` of the step, but I don't think it's important, so off it goes.

pileks:
> I think we should remove the "pages" entry in the manifest, to help simplify this new feature. Let's only keep the features we need for initial MVP of wrapscan.
> 
> To me this is:
> 
> * description
> * logo
> * repository
> * website
> * examples

I believe the correct fields for the Wrapscan MVP are:
- Description
  - this one is obvious - each Wrap should have a description rendered in the list and above its readme page
  - We *would* be duplicating this information within Wrapscan's homepage for starters, and later moving it into some form of incremental cache
- Logo
  - this one is necessary too, same reason as description, same caveats
- Readme
  - a path to the readme file. I'd keep this so that we can later add multiple readme pages easily, and docs manifest migrations will be easier to do if we immediately have paths instead of writing to a default readme path (we're not renaming files, we're just placing them in the appropriate place ahead-of-time, and the code for this is already completed).
- Repository URL - the wrap info widget renders this
- Website URL - the wrap info widget renders this

We don't need the Examples field actually, as it's not part of the MVP. Nevertheless, I'd keep it as the sturcture seems good enough, and we'll be able to add examples much quicker if we just keep it.

packages/manifests/polywrap/formats/polywrap.docs/0.1.0.json
dOrgJelli: Nice :D

packages/js/manifests/polywrap/src/formats/polywrap.docs/0.1.0.ts
```diff
@@ -0,0 +1,73 @@
...
+   */
+  description?: string;
+  /**
+   * Path to the wrap's logo.
+   */
+  logo?: string;
+  /**
+   * Website related to the wrap.
+   */
+  websiteUrl?: string;
```
dOrgJelli: NIT: I think we can do without the `Url` at the end. Other popular manifests (`Cargo.toml`, `package.json`) don't do this.

Also if we follow this naming convention, then we'd also have `logoPath` and `readmePath`. Better to just omit it IMO.

dOrgJelli: Same for `repositoryUrl`

packages/js/manifests/polywrap/src/formats/polywrap/migrators/0.3.0_to_0.4.0.ts
dOrgJelli: Agreed, nice touch! I was initially thinking we would want to use something like `polywrap manifest init docs` for this, but I can get behind having a `polywrap docs` root command, just like we have for `polywrap build`.

dOrgJelli: I do think we could also suggest the user use `polywrap manifest schema docs` to see what properties are available.

packages/manifests/polywrap/formats/polywrap.docs/0.1.0.json
```diff
@@ -0,0 +1,81 @@
...
+                "required": ["method"],
+                "properties": {
+                  "description": {
+                    "type": "string",
+                    "description": "A description of the step rendered above the invocation code."
+                  },
+                  "uri": {
+                    "description": "URI of the wrap being invoked.",
+                    "type": "string",
+                    "tsType": "any"
```
dOrgJelli, line 61: Why is this needed? Don't we want this to be typed a "string"?

packages/cli/src/lib/Compiler.ts
```diff
@@ -46,6 +48,10 @@ export class Compiler {
 
         // Copy: Resources folder
         await this._copyResourcesFolder();
+
+        // Output docs if any
+
```
dOrgJelli: NIT: extra space

dOrgJelli (changes requested):

packages/js/manifests/polywrap/src/formats/polywrap/migrators/0.3.0_to_0.4.0.ts
dOrgJelli: I've updated the `polywrap docs init` command to emit a fully commented yaml version of the docs manifest, so no need to suggest the user "read the docs" via the `polywrap manifest schema docs` command.

dOrgJelli (dismissed):

PR #1769, opened 2023-05-12
CLI - `docs init` command

### LINKED ISSUE #1650: CLI - `docs init` command
`docs init` should set up a `polywrap.docs.yaml` file and add it as an extension to `polywrap.yaml`.
It should additionally set up a basic folder structure with all files required/available for our documentation (intro md, logo).

pileks:
Closes #1650 

This PR introduces the `docs` CLI command with its `init` subcommand.

The `init` command initializes your Wrap's docs manifest and a directory with a sample readme file.


packages/cli/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "polywrap",
   "description": "Polywrap CLI",
-  "version": "0.10.3",
+  "version": "0.10.3aaa",
```
krisbitney: Is this intentional?

krisbitney (changes requested):
This looks really great!

I have a request: could you add a test for the new command to the `cli-js` package?

Also, I think should we remove the `polywrap docgen` command. But it makes sense to do that in a separate PR, so that this one remains easy to review. I can create the PR to remove it if you agree. Or you can do it in this PR or another, however you'd like. What do you think?

packages/cli/package.json
pileks: Whoops!

pileks:
> This looks really great!
> 
> I have a request: could you add a test for the new command to the `cli-js` package?
> 
> Also, I think should we remove the `polywrap docgen` command. But it makes sense to do that in a separate PR, so that this one remains easy to review. I can create the PR to remove it if you agree. Or you can do it in this PR or another, however you'd like. What do you think?

Both good ideas, will get to working on the request as part of this PR.

PR #1776, opened 2023-05-30
HTTP deployment of redirect and added headers to http deployer

nerfZael:


pileks (dismissed):

namesty (dismissed):

packages/cli/src/lib/defaults/deploy-modules/http/index.ts
```diff
@@ -34,28 +32,43 @@ const dirToFormData = (baseDirPath: string) => {
...
+  ): Promise<Uri> {
     if (!config?.postUrl) {
       throw new Error(
         `HTTP Deployer error: No postUrl provided. Please provide a postUrl in the deploy manifest's config object`
       );
     }
 
-    const formData = dirToFormData(uri.path);
+    let response;
+    if (uri.authority === "fs") {
```
dOrgJelli: "file" is also being treated as a valid file-system URI authority, going to add that here.

PR #1786, opened 2023-06-19
fix: python-plugin bindings

Niraj-Kamdar:


packages/cli/lang/en.json
```diff
@@ -14,7 +14,7 @@
   "commands_build_keypressListener_exit": "Exit: [CTRL + C], [ESC], or [Q]",
   "commands_build_keypressListener_watching": "Watching",
   "commands_build_options_h": "Show usage information",
-  "commands_build_options_m": "Path to the Polywrap manifest file (default: {default})",
+  "commands_build_options_m": "Path to the Polywrap Build manifest file (default: {default})",
```
Niraj-Kamdar: This was missing for some reason

packages/js/cli/src/__tests__/commands.spec.ts
```diff
@@ -227,6 +227,7 @@ const testData: CommandTestCaseData<CommandTypings> = {
     init: [{
       cwd: fs.mkdtempSync(path.join(os.tmpdir(), "docs-init")),
       before: async (test) => {
+        fs.writeFileSync(`${test.cwd}/polywrap.yaml`,"");
```
Niraj-Kamdar: `polywrap docs init` doesn't create polywrap manifest so we need to have polywrap manifest to create docs manifest

packages/cli/lang/en.json
dOrgJelli: This was actually correct before. It's the path to the `polywrap.yaml` manifest, not `polywrap.build.yaml`.

dOrgJelli (approved):

PR #1795, opened 2023-06-29
feat: add wrap-abi-bindgen support + upgrade js client to v0.11

dOrgJelli:
Kris: This PR integrates `wrap-abi-bindgen`, which replaces the current bindings generation with wrap-based bindings.

See https://github.com/polywrap/wrap-abi-bindgen

A separate PR will add `wrap-rust` wrap-based bindings.


todo
```diff
@@ -0,0 +1,3 @@
+- when codegen for wrap/assemblyscript, use wrap URI
```
dOrgJelli: this file can be removed

packages/cli/src/__tests__/e2e/helpers/testCliOutput.ts
```diff
@@ -19,6 +19,12 @@ export const testCliOutput = (
     )
   );
 
+  if (expected.exitCode && exitCode !== expected.exitCode) {
```
dOrgJelli: I think the proper condition here would be `if (exitCode !== expected.exitCode)`, no need for the `else if` below if you use this.

packages/schema/bind/src/bindings/index.ts
```diff
@@ -1,30 +1,42 @@
...
 export function getGenerateBindingFn(
   bindLanguage: BindLanguage
 ): GenerateBindingFn {
   switch (bindLanguage) {
-    case "wasm-as":
-      return AssemblyScript.Wasm.generateBinding;
-    case "wasm-rs":
+    case "wrap-as":
+      return WrapBindgen.getGenerateBindingFn(
+        "https://github.com/polywrap/wrap-abi-bindgen/tree/dev/implementations/wrap-assemblyscript"
```
dOrgJelli: Can we update this to be a "main" branch instead of "dev"? Might I suggest we call this branch `wrap-0.1`?

packages/schema/bind/src/bindings/wrap-bindgen.ts
```diff
@@ -0,0 +1,70 @@
...
+      type: options.wrapInfo.type,
+      abi: JSON.stringify(options.wrapInfo.abi),
+    },
+    context: options.config ? JSON.stringify(options.config) : undefined,
+  };
+
+  const result = await client.invoke<Output>({
+    uri,
+    method: "generateBindings",
+    args: (args as unknown) as Record<string, unknown>,
```
dOrgJelli, line 42: Unrelated: Is this casting really required? If so I think we should fix this because it isn't a very good developer experience.

todo
krisbitney: done!

packages/cli/src/__tests__/e2e/helpers/testCliOutput.ts
krisbitney: fixed

packages/schema/bind/src/bindings/index.ts
krisbitney: done!

packages/schema/bind/src/bindings/wrap-bindgen.ts
krisbitney: Yeah, it's required when using statically typed args objects. TypeScript complains without the cast because `args` is of type `Record<string, unknown> | Uint8Array | undefined`.

PR #1796, opened 2023-06-29
fix: properly remove wasm-bindgen from rust wasm modules

dOrgJelli:
NOTE: the main difference is when we call `wasm-snip` we explicitly instruct it to remove the nefarious wasm-bindgen import that remains, even if no other wasm-bindgen imports are being used.
`wasm-snip input.wasm -o output.wasm -p .*__wbindgen_throw`

I've tested this using the ethers-core wrap, which was experiencing a problem of having left-over wasm-bindgen imports, and this fix is fully functional from my testing.


pileks (dismissed):
I like the new, clearer steps with reasoning!

dOrgJelli:
NOTE: @cbrzn is investigating wrap-test-harness CI issues

cbrzn (dismissed):

PR #1808, opened 2023-07-10
update IPFS Deployer module

krisbitney:
The IPFS Deployer Module was rewritten to use the `PolywrapClient` and IPFS HTTP Client wrap. This is compatible with Node 18.

### DIFF
--- a/packages/cli/src/lib/defaults/deploy-modules/ipfs/index.ts
+++ b/packages/cli/src/lib/defaults/deploy-modules/ipfs/index.ts
@@ -1,13 +1,72 @@
 import { DeployModule } from "../../../deploy";
 
 import { Uri } from "@polywrap/core-js";
-
-// eslint-disable-next-line @typescript-eslint/no-require-imports,@typescript-eslint/no-var-requires,@typescript-eslint/naming-convention
-const IPFSClient = require("ipfs-http-client");
-const { globSource } = IPFSClient;
+import { PolywrapClient } from "@polywrap/client-js";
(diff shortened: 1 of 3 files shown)


packages/cli/src/lib/defaults/deploy-modules/ipfs/index.ts
```diff
@@ -18,26 +77,34 @@ class IPFSDeployer implements DeployModule {
...
+      timeout: 10000,
     };
 
-    const addOptions = {
-      wrapWithDirectory: false,
-    };
+    const client = new PolywrapClient();
 
-    let rootCID = "";
+    const result = await client.invoke<AddResult[]>({
```
dOrgJelli, line 96: @krisbitney I'm good with merging this now, but can you please make a TODO for an A/B test locally? I want to make sure that deploying a wrap using this new ipfs deployer yields the same URI as the old deployer.

dOrgJelli (approved):

packages/cli/src/lib/defaults/deploy-modules/ipfs/index.ts
krisbitney: Sure thing. It produces the same IPFS URI in the CLI test cases. I'll make sure it also works with a real wrap.

krisbitney: I confirmed that it produces the same IPFS hash for a real wrap (tested on wrap-abi-bindgen plugin-kotlin wrap)

PR #1818, opened 2023-07-13
feat: add go templates for create command

Niraj-Kamdar:


dOrgJelli (approved):
:fire: :fire_engine: :clap:

Issue #1846, opened 2023-08-02
IPFS Deployer possibly duplicates subdirectory paths when deploying


pileks:
When deploying using the IPFS deployer, dirs within dirs possibly has an issue wherein subpaths are "multiplied":

Structure of build dir:
```
├── docs
│   ├── pages
│   │   └── readme.md
│   └── polywrap.docs.json
├── wrap.info
└── wrap.wasm
```

Structure of deployment to IPFS:
https://ipfs.io/ipfs/QmaG1butqzL7vXqs5hZCkba4LAiGStrh3vx6eKz8sdFfDo/
```
├── docs
|   |
|   └──docs
│      ├── pages
│      │   └── readme.md
│      └── polywrap.docs.json
├── wrap.info
└── wrap.wasm
```

Notice the difference in the docs dir - the path to the wrap readme is different:
docs/pages/readme.md - local build dir
vs
docs/docs/pages/readme.md - what's on IPFS

Further investigation warranted.

dOrgJelli:
This should be fixed within `polywrap` CLI version `0.11.1`

PR #1847, opened 2023-08-03
Fix IPFS deployer subdirectory duplication issue

pileks:
This PR aims to:

- Fix the IPFS deployer subdirectory issue described in #1846 
- Lightly refactor the IPFS deployer to allow unit tests to be written for this case


packages/test-cases/cases/cli/lib/deployers/mock-build-output/wrap.info
cbrzn, line 1: do we need these generated files? or they can be ignored?

cbrzn (approved):
lgtm :green_apple: just left a minor nit

packages/test-cases/cases/cli/lib/deployers/mock-build-output/wrap.info
pileks: These are not actually generated, they are empty mock files intended to test whether the IPFS deployer properly reads the folder structure.

cbrzn: gotcha, and would it be possible to fetch those from this already existing folder https://github.com/polywrap/cli/tree/46719d78d7a990b281ca5d359b060282a2942131/packages/test-cases/cases/cli/deploy/wrapper? or we can even delete this folder and use the wraps from the generated `cases` folder (generated from this script https://github.com/polywrap/cli/blob/origin-dev/packages/test-cases/package.json#L9)

krisbitney (approved):
LGTM

krisbitney:
@CodiumAI-Agent /review

packages/test-cases/cases/cli/lib/deployers/mock-build-output/wrap.info
pileks: Unfortunately no, that case (just `wrap.info` and `wrap.wasm`) won't cover the exact needs of this test.

The test needs this dir within a dir structure to determine whether the IPFS deployer properly reads the dir and subdirs.

This is essentially an edge-case here, and I'd like to keep the test for it separate from the other cases, especially since I created a new **unit** test for this, not an e2e test.

PR #1855, opened 2023-08-15
Feat: update typescript/app template to latest bindings

dOrgJelli:


packages/templates/app/typescript/src/ethereum.ts
```diff
@@ -0,0 +1,15 @@
+import { BaseEthereum } from "./wrap";
```
Niraj-Kamdar: This would go in npm sdk

packages/templates/app/typescript/src/index.ts
```diff
@@ -1,36 +1,31 @@
-import {
-  Logging_Module,
-  Ethereum_Module,
-} from "./wrap";
-
-import { PolywrapClient } from "@polywrap/client-js";
-
-const client = new PolywrapClient();
+import { Ethereum } from "./ethereum";
+import { Logger } from "./logger";
```
Niraj-Kamdar: This would be npm wrap sdk

packages/templates/app/typescript/src/ethereum.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { CoreClient, PolywrapClient } from "@polywrap/client-js";
+
+export class Ethereum extends BaseEthereum {
```
dOrgJelli: As an app developer, it feels weird that I have to do this every time. I would imagine that the code below would be the "sane defaults", and if I wanted to change it I could do something like:
```typescript
import { Ethereum as DefaultEthereum } from "./wrap"

class Ethereum extends DefaultEthereum {
  protected _getDefaultClient(): CoreClient {
    // TODO: create your own custom client
  }
}
```

PR #1856, opened 2023-08-17
feat: add python app template

Niraj-Kamdar:


packages/templates/app/python/polywrap.graphql
```diff
@@ -0,0 +1 @@
+#import * into Ethereum from "ens/wraps.eth:ethereum@1.1.0"
```
dOrgJelli: We should be using wrapscan.io URIs everywhere now. For this, use `wrapscan.io/polywrap/ethers-utils@1.0`

packages/cli/src/lib/codegen/CodeGenerator.ts
```diff
@@ -87,6 +88,23 @@ export class CodeGenerator {
       ? await overrides.getSchemaBindConfig(this._config.project)
       : {};
 
+    if (language === "app-py") {
```
dOrgJelli: We should not be doing this in such a hacky way. We have "codegen overrides" for this reason.

See an example here: https://github.com/polywrap/cli/blob/e3f3597e6249e5b03363b2789583f580be690516/packages/cli/src/lib/defaults/language-overrides/wasm/golang/index.ts#L20-L31

dOrgJelli (changes requested):
Also let's make sure CI is green.

packages/cli/src/lib/codegen/CodeGenerator.ts
Niraj-Kamdar: I am doing it since codegenOverrides doesn't support what I want to do. We need to refactor codegen overrides but I think it should be a separate PR

Niraj-Kamdar: Wait! It seems like it's supported now!!!

packages/templates/app/python/polywrap.graphql
Niraj-Kamdar: python-client is still using ens URI so need to update there first

packages/cli/src/lib/defaults/language-overrides/app/python/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import fs from "fs";
+import path from "path";
+
+export function getCodegenOverrides(): CodegenOverrides {
+  return {
+    getSchemaBindConfig: async (project: PolywrapProject) => {
+      const manifestPath = project.getManifestPath();
+      const manifestDir = path.dirname(manifestPath);
+      const pyprojectPath = path.join(manifestDir, "pyproject.toml");
```
dOrgJelli, line 12: Isn't it also possible for python projects to not use a `pyproject.toml` config file, for example if they use `requirements.txt`?

In the case we don't find a `pyproject.toml` file, can we just default to `./wrap` and not throw?

packages/templates/app/python/polywrap.graphql
dOrgJelli: "encode params" can come from the ethers-utils wrap, and does not require the ethereum-wallet plugin. I've pushed a change to fix this.

packages/cli/src/lib/defaults/language-overrides/app/python/index.ts
Niraj-Kamdar: https://peps.python.org/pep-0621/
[PEP 621 – Storing project metadata in pyproject.toml | peps.python....](https://peps.python.org/pep-0621/)
Pyproject.toml is the standard way to declare project metadata for Python projects.

dOrgJelli (approved):

PR #1864, opened 2023-08-22
feat: add rust app codegen template

Niraj-Kamdar:


dOrgJelli (dismissed):

dOrgJelli (dismissed):

krisbitney (dismissed):
LGTM!

packages/templates/app/rust/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "private": true,
+  "version": "0.11.2",
+  "scripts": {
+    "build": "npx polywrap codegen",
+    "test": "ts-node ./src/index.ts",
+    "ipfs:up": "npx polywrap infra up --modules=eth-ens-ipfs",
+    "ipfs:down": "npx polywrap infra down --modules=eth-ens-ipfs"
+  },
+  "dependencies": {
+    "@polywrap/client-js": "~0.12.0"
```
cbrzn: this is not needed right

packages/templates/app/rust/tsconfig.json
cbrzn, line 1: we can also remove this one right

packages/templates/app/rust/package.json
```diff
@@ -0,0 +1,21 @@
...
+  },
+  "devDependencies": {
+    "@types/node": "18.15.0",
+    "polywrap": "0.11.2",
```
cbrzn: we can only keep `polywrap`?

packages/templates/app/rust/.nvmrc
```diff
@@ -0,0 +1 @@
+v17.9.1
```
dOrgJelli: Should be v18.15.0 like all other .nvmrc files in the CLI repo

packages/templates/app/rust/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "template_plugin_rs"
```
dOrgJelli: should be "template_app_rs"

packages/templates/app/rust/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "templates-app-rust",
+  "description": "Polywrap App Rust Template",
```
dOrgJelli: No need for name, description, version

packages/templates/app/rust/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "private": true,
+  "version": "0.11.2",
+  "scripts": {
+    "build": "npx polywrap codegen",
+    "test": "ts-node ./src/index.ts",
+    "ipfs:up": "npx polywrap infra up --modules=eth-ens-ipfs",
```
dOrgJelli: this isn't needed

packages/templates/app/rust/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "templates-app-rust",
+  "description": "Polywrap App Rust Template",
+  "private": true,
+  "version": "0.11.2",
+  "scripts": {
+    "build": "npx polywrap codegen",
+    "test": "ts-node ./src/index.ts",
```
dOrgJelli: this doesn't work

packages/templates/app/rust/tsconfig.json
```diff
@@ -0,0 +1,30 @@
+{
```
dOrgJelli: this file isn't needed

dOrgJelli (changes requested):

packages/templates/app/rust/package.json
Niraj-Kamdar: Forget to remove! before I was using local ipfs in template but then migrate to just cat instead of add and cat

PR #1867, opened 2023-08-28
Android (Kotlin) app template

krisbitney:
This PR adds an Android app template using the Kotlin app bindings.

Depends on:
https://github.com/polywrap/cli/pull/1868
https://github.com/polywrap/wrap-abi-bindgen/pull/53

Ethereum wrapper invocation is not yet working due to incompatibility in Rust wrapscan uri resolver:

```
uniffi.polywrap_native.FfiException$LoadWrapperException: uri=wrap://ens/wraps.eth:ethereum@1.0.0, err=Failed to resolve wrapper: `Error loading wrapper, uri: wrap://wrapscan.io/polywrap/wrapscan-uri-resolver@1.0: `Failed to resolve wrapper: `Failed to invoke wrapper, uri: `wrap://wrapscan.io/polywrap/http-uri-resolver@1.0`, method: `tryResolveUri`: missing required CPU features: "EnumSet(SSE2 | SSE3 | SSSE3 | SSE41 | SSE42 | POPCNT | AVX | BMI1 | BMI2 | AVX2 | AVX512DQ | AVX512VL | AVX512F | LZCNT)"
```


packages/cli/src/__tests__/e2e/p1/create.spec.ts
```diff
@@ -17,7 +17,7 @@ Commands:
   wasm [options] <language> <name>    Create a Polywrap wasm wrapper. langs:
                                       assemblyscript, rust, golang, interface
   app [options] <language> <name>     Create a Polywrap application. langs:
-                                      typescript
+                                      typescript, android
```
Niraj-Kamdar: Can there be vanilla kotlin or something? Just wondering since the template here is android specific

krisbitney: We could create another template that is not Android, sure. We have always treated Kotlin as a means of targeting android users, so I think having an Android template makes more sense.

dOrgJelli (dismissed):

dOrgJelli (approved):

PR #1870, opened 2023-08-29
Swift iphone app template

krisbitney:
depends on https://github.com/polywrap/wrap-abi-bindgen/pull/55 and https://github.com/polywrap/cli/pull/1871


packages/templates/app/ios/polywrap.graphql
```diff
@@ -0,0 +1,2 @@
+#import * into Logging from "wrapscan.io/polywrap/logger@1.0"
+#import * into Ethereum from "wrapscan.io/polywrap/ethers@1.1.0"
```
cbrzn: nit:
```suggestion
#import * into Ethers from "wrapscan.io/polywrap/ethers@1.1"
```

packages/cli/src/commands/create.ts
```diff
@@ -28,7 +28,7 @@ const urlStr = intlMsg.commands_create_options_t_url();
 
 export const supportedLangs = {
   wasm: ["assemblyscript", "rust", "golang", "interface"] as const,
-  app: ["typescript"] as const,
+  app: ["typescript", "ios"] as const,
```
cbrzn: should here be `swift` and not `ios`? btw idk why `swift` is not added in the `plugin` array below this line :thinking: we have support for plugins

packages/templates/app/ios/Template/PolywrapDemo.swift
```diff
@@ -0,0 +1,27 @@
...
+    print("Invoking: Logging.info(...)")
+
+    let logger = Logging()
+    let logArgs = LoggingArgsLog(level: .INFO, message: "Hello there")
+    try? logger.log(args: logArgs)
+
+    // Ethereum.encodeParams
+    print("Invoking: Ethereum.encodeParams(...)")
+
+    let eth = Ethereum()
```
cbrzn: why the proposed change this would change to `Ethers()`

cbrzn (commented):
this is awesome thank you for pushing all these changes :_)

packages/cli/src/commands/create.ts
krisbitney: I made it "ios" and "android" for swift and Kotlin for clarity. 

Swift is used for iOS, macOS, watchOS, and tvOS and I'm not sure what else.

Kotlin is used for Android, desktop, web back ends, multiplatform mobile apps, and more.

Do you think it makes more sense to do "swift" and "kotlin"? Maybe iOS and Android use cases are implied?

Niraj-Kamdar: I lean toward the android/ios since the template specifically targets those. If it was a generic template then the language name would make more sense

packages/templates/app/ios/Template/PolywrapDemo.swift
krisbitney: Fixed!

packages/cli/src/commands/create.ts
cbrzn: what you're saying make sense, but swift client supports `iOS` and `macOS`.

it still makes more sense to me to name this `swift` and `kotlin` rather than `ios` or `android`; this is codegen and from my perspective it should be agnostic of the OS (even tho I completely understand your point)

nerfZael: If they're android/ios specific templates then IMO this should be "ios".
But, correct me if I'm wrong, it doesn't seem that is the case: https://github.com/polywrap/wrap-abi-bindgen/pull/55/files#diff-2539b6e33504cb7efe01bdb09121a3aaba4ba6023d13c460f34dcb161ca244de

Also, whatever we do, we should be consistent. In this thread there's a push for "ios" while in that link it's "app-swift"

krisbitney: The codegen bindings are not specific to ios, but the app template is a simple ios app

namesty: iOS and Android are platforms, not languages. It's a bit weird to have "iOS, Golang, Android, Typescript, Interface"; as it seems there are 3, or at least 2, concepts mixed in there.

That said, the template you're talking about targets "iOS" in particular. If it was generic enough to be built to macOS or tvOS, then it should be swift; but it isn't. I think here the answer should be "ios"; but indeed a bit weird to have "ios" as a "supportedLang". Maybe it's just a variable naming change thing, but nvm.

dOrgJelli: This situation is VERY similar to when we had a "react" app template. So, based on how we handled that, I suggest we:
1. Merge these android & ios templates now so they can be used
2. Create new app templates that are "vanilla" kotlin & swift
3. Move ios & android templates to their own repos

dOrgJelli: Also one detail to make sure we're all clear on: "android" and "ios" are ONLY used within the `polywrap create` command. All other places (codegen, manifest, etc) "kotlin" and "swift" are used.

dOrgJelli: Also one detail to make sure we're all clear on: "android" and "ios" are ONLY used within the `polywrap create` command. All other places (codegen, manifest, etc) "kotlin" and "swift" are used.

PR #1874, opened 2023-08-31
Use Rust tests in wrap-rust template

krisbitney:
This PR replaces the JS e2e tests in the wrap-rust template with Rust tests. The Rust app bindings are generated and used for the e2e tests to provide a nice UX.

It also adds the Rust app bindings to the CLI, which hadn't been done yet, so codegen works as expected.


packages/templates/wasm/rust/package.json
```diff
@@ -8,8 +8,8 @@
     "build": "npx polywrap build",
     "deploy": "npx polywrap deploy",
     "test": "yarn test:e2e && yarn test:workflow",
-    "test:e2e": "yarn test:e2e:codegen && jest --passWithNoTests --runInBand --verbose",
-    "test:e2e:codegen": "npx polywrap codegen -m ./src/__tests__/types/polywrap.app.yaml -g ./src/__tests__/types/wrap",
+    "test:e2e": "yarn test:e2e:codegen && cargo test --release",
+    "test:e2e:codegen": "npx polywrap codegen -m ./tests/types/polywrap.app.yaml -g ./tests/types/wrap",
     "test:workflow": "npx polywrap test"
   },
   "devDependencies": {
```
cbrzn, line 15: i think we can also remove `jest` and its relatives from here?

packages/templates/wasm/rust/tests/e2e.rs
cbrzn, line 1: since in rust you would normally create the `tests` folder for integration tests (for unit you would do the tests in the same file); what do you think about moving this e2e file to `it/module.rs`
rather than having:
```
|_tests/
  |_e2e.rs
  |_types/
```
we would have:
```
|_tests/
  |_it/
    |_module.rs
  |_types/
```
this is more idiomatic imo

packages/templates/wasm/rust/Cargo.toml
```diff
@@ -1,17 +1,22 @@
...
+include = ["src"]
+
 [dependencies]
-polywrap-wasm-rs = { version = "~0.11.0-pre.4" }
+polywrap-wasm-rs = { version = "~0.11.2" }
 polywrap_msgpack_serde = "~0.0.2-beta.5"
 serde = { version = "1.0", features = ["derive"] }
 
+[dev-dependencies]
+polywrap = { version = "~0.1.8" }
```
cbrzn:
```suggestion
polywrap = { version = "0.1.9-beta.2" }
```

packages/templates/wasm/rust/Cargo.toml
```diff
@@ -1,17 +1,22 @@
...
 repository = "https://github.com/polywrap/monorepo"
 license = "MIT"
 edition = "2021"
 
+include = ["src"]
+
 [dependencies]
-polywrap-wasm-rs = { version = "~0.11.0-pre.4" }
+polywrap-wasm-rs = { version = "~0.11.2" }
 polywrap_msgpack_serde = "~0.0.2-beta.5"
```
cbrzn:
```suggestion
polywrap_msgpack_serde = "0.0.2-beta.7"
```

packages/templates/wasm/rust/Cargo.toml
krisbitney: I'm not sure we should use a beta release in the template

packages/templates/wasm/rust/package.json
krisbitney: Done!

packages/templates/wasm/rust/Cargo.toml
krisbitney: updated

packages/templates/wasm/rust/tests/e2e.rs
krisbitney: Done!

dOrgJelli (approved):

PR #1888, opened 2023-09-05
TS Wraps support

namesty:
This PR adds:

Creating TS wrap projects
Codegen for TS wrap bindings
Template for TS wrap projects
Building TS wraps with Image and VM strategies
Additionally, it separates binding languages from build languages: go, rust and assemblyscript are both; interfaces and typescript are bindings only and javascript is, currently, build only.

The reason for this is that we shouldn't need 2 different directories for JS and TS build strategies; because TS wraps are TS bindings + rollup.js config that produce a single JS file, that is then built as a JS wrap.

The reasons I built it this way are basically letting users tweak bundling configurations and debug their bundles freely, or even use other bundlers like webpack; and in general, not doing anything special or extra on our side for TS, but rather trying to get it as cheaply as possible on top of JS wraps


packages/test-cases/cases/cli/build-cmd/wasm/typescript/001-sanity/polywrap.yaml
```diff
@@ -0,0 +1,9 @@
+format: 0.5.0
+project:
+  name: ObjectTypes
+  type: wasm/typescript
+source:
+  schema: ./polywrap.graphql
+  module: ./src/index.ts
```
dOrgJelli: NIT: I think we need to (1) make this `module: ./wrap.js` and introduce a new section where we define our codegen directory:
```yaml
bindgen:
  directory: ./src/wrap
  schema: ./polywrap.graphql
  impl: wrapscan.io/polywrap/wrap-typescript-abi-bindgen@1
```

packages/test-cases/cases/cli/build-cmd/wasm/typescript/001-sanity/polywrap.build.yaml
```diff
@@ -0,0 +1,3 @@
+format: 0.3.0
+config:
+  scriptFile: ./wrap.js
```
dOrgJelli: This ideally shouldn't be needed, it feels superfluous to always have to have a `polywrap.build.yaml` file IMO.

packages/schema/bind/src/bindings/index.ts
```diff
@@ -21,6 +21,10 @@ export function getGenerateBindingFn(
       );
     case "wrap-go":
       return Golang.Wasm.generateBinding;
+    case "wrap-ts":
+      return WrapBindgen.getGenerateBindingFn(
+        "wrap://http/http.wrappers.dev/u/test/wrap-typescript-wrap-abi-bindgen"
```
dOrgJelli: This can be changed to `wrapscan.io/polywrap/wrap-typescript-abi-bindgen@1`:
https://github.com/polywrap/wrap-abi-bindgen/pull/103

packages/cli/src/lib/project/PolywrapProject.ts
```diff
@@ -128,6 +129,18 @@ export class PolywrapProject extends Project<PolywrapManifest> {
...
+  public async getBuildLanguage(): Promise<PolywrapBuildLanguage> {
+    const manifestLanguage = await this.getManifestLanguage();
+
+    if (manifestLanguage === "interface") {
+      throw new Error(`Cannot build an interface project.`);
+    }
+
+    return manifestLanguage === "wasm/typescript"
+      ? "wasm/javascript"
+      : manifestLanguage;
```
dOrgJelli, line 142: I would put this in a utility function inside of `project/manifests/polywrap/languages.ts`. This mapping can be used elsewhere I would imagine.

packages/cli/src/lib/defaults/build-strategies/wasm/javascript/vm/VERSION
```diff
@@ -0,0 +1 @@
+0.1.6
```
dOrgJelli: NIT: strange that we started this off at `0.1.6`

packages/cli/src/lib/defaults/build-strategies/wasm/javascript/vm/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ubuntu:latest
+RUN apt-get update && apt-get install -y curl git
+
+RUN curl -L https://raw.githubusercontent.com/polywrap/pwr/main/pwrup/install | bash && /root/.pwr/bin/pwrup
```
dOrgJelli, line 6: NIT: let's add comments so we know why dependencies are needed.

packages/cli/src/lib/defaults/build-strategies/wasm/javascript/local/local.sh
```diff
@@ -0,0 +1,4 @@
+curl -L https://raw.githubusercontent.com/polywrap/pwr/main/pwrup/install | bash
+source $HOME/.bashrc
+pwrup
```
dOrgJelli: NIT: let's add a comment saying what this does exactly.

packages/cli/src/lib/build-strategies/strategies/DockerVMStrategy.ts
```diff
@@ -136,6 +143,19 @@ export class DockerVMBuildStrategy extends BuildStrategy<void> {
         });
       }
 
+      // TODO: come up with more general solution for this
+      // For JS we need to copy the script file; which may be at root lvl
+      if (language === "wasm/javascript") {
+        // validated by buildManifest manifest validation
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        const scriptFile = buildManifest.config!.scriptFile as string;
```
dOrgJelli: This should be able to be done through a `COPY` command in docker. The manifest properties & config variables are available there.

packages/cli/src/lib/build-strategies/strategies/DockerVMStrategy.ts
```diff
@@ -17,7 +17,9 @@ import fse from "fs-extra";
 import path from "path";
 import Mustache from "mustache";
 
-type BuildableLanguage = Exclude<PolywrapManifestLanguage, "interface">;
+type BuildableLanguage =
```
dOrgJelli: This is a redefinition of the type defined in packages/cli/src/lib/project/manifests/polywrap/languages.ts

dOrgJelli (changes requested):
Looking really good! Just a few minor NITS

packages/test-cases/cases/cli/build-cmd/wasm/typescript/001-sanity/polywrap.yaml
namesty: I agree with this approach for this PR.

That said, I think we should look into de-escalating the introduction of new config concepts; maybe we could look into how to make these concepts command line arguments or something similar.

packages/cli/src/lib/project/PolywrapProject.ts
namesty: I think this is probably a PolywrapProject specific; just like `getManifestLanguage`, which is right above this definition. I'd even say this util method is `getManifestLanguage` + some conditions in regards to build. It isn't used anywhere outside of this context.

No strong opinions either, will move it if you feel it's more correct

packages/cli/src/lib/defaults/build-strategies/wasm/javascript/vm/VERSION
namesty: Yeah, the image I published has no version tag. Just initted this with 0.0.1

packages/cli/src/lib/build-strategies/strategies/DockerVMStrategy.ts
namesty: I changed this part so that it's more consistent with what we already do with sources copying. We're keeping the original module file path instead of just its containing directory and using that as "sources" for JS wraps.

That said, a sidenote is we can't copy these with a `COPY` command because the VM's Dockerfile isn't what copies the sources; the sources are copied into a docker shared volume directory that lives in the user's OS. When the VM script is then run inside of the VM container, it assumes the source files it needs are already there and accessible through the shared volume.

packages/test-cases/cases/cli/build-cmd/wasm/typescript/001-sanity/polywrap.build.yaml
namesty: Agreed. Removed in favor of using `polywrap.yaml`'s module file path

packages/templates/wasm/typescript/polywrap.build.yaml
```diff
@@ -0,0 +1,3 @@
+format: 0.3.0
+config:
+  scriptFile: ./bundled/wrap.js
```
dOrgJelli, line 3: can this file be removed?

packages/cli/src/lib/project/PolywrapProject.ts
dOrgJelli: Sounds good to me, single use so no need to abstract farther right now

packages/cli/src/lib/defaults/build-strategies/wasm/javascript/vm/VERSION
dOrgJelli: @namesty it looks like you might have accidentally published `0.1.6`, see here: https://hub.docker.com/r/polywrap/vm-base-js/tags

Not sure if we can "yank" the tag or not. Also let's start dev versioning at `0.1.0` like we do in our other software.

dOrgJelli (approved):

PR #1893, opened 2023-09-07
Add `embed` option to `polywrap codegen`

### LINKED ISSUE #1703: Support embedding wraps in JS via `polywrap embed` for `app/` projects
Manifest:
```yaml
format: 0.3.0
project:
  name: uniswap-v3-js
  type: app/typescript
source:
  schema: ./schema.graphql
```

Schema:
```yaml
#import * into UniswapV3 from "ens/uniswap.wraps.eth:v3@1.0.0"
```

CLI:
```
polywrap codegen
```
```
polywrap embed
```
Example: https://github.com/polywrap/toolchain/blob/origin-dev/packages/js/client-config-builder/src/bundles/embeds/async-ipfs-resolver/wrap.ts

SDK Developer (psuedo code, subject to change):
`index.ts`
```typescript
import { Uniswap } from "./wrap";
import { embeds } from "./wrap/embeds";

import {
  ClientConfigBuilder,
  PolywrapClient
} from "@polywrap/client-js";

// create default client configuration w/ embeds
  .addDefaults()
  .addPackages(embeds)
  // configure ethereum provider?
  .build();
  
const client = new PolywrapClient(config);

// set default client
Uniswap.setDefaultClient(client);

export * from "./wrap";
```

App Developer:
`npm install --save @polywrap/uniswap-v3-js`

```typescript
import { Uniswap } from "@polywrap/uniswap-v3-js";

const uniswap = new Uniswap(
  // optionally add client, if nothing then default
);

uniswap.someFunction({
  ...args
});

uniswap.someFunction({
  ...args
}/*, optional client */);
```

krisbitney:
This PR adds a `--embed` option to `polywrap codegen`. When the option is specified, code for embedded wraps will be generated.

See Rust implementation: https://github.com/polywrap/wrap-abi-bindgen/pull/107

Closes https://github.com/polywrap/cli/issues/1703


packages/cli/src/commands/codegen.ts
```diff
@@ -54,6 +57,7 @@ export const codegen: Command = {
         })}`
       )
       .option(`-b, --bindgen <URI>`, `${intlMsg.commands_codegen_options_b()}`)
+      .option(`-e, --embed`, `${intlMsg.commands_codegen_options_e()}`)
```
dOrgJelli, line 60: Currently the CLI isn't able to detect if the abi-bindgen wrap supports this feature. Not sure how to best handle this, maybe we should instead bump the feature version of the abi-bindgen interface, and the CLI could see if the implementation supports this feature?

This doesn't matter too much, but just a thought.

dOrgJelli (commented):
Super minor NIT, other than this looks great!

dOrgJelli:
Not sure why CI is not green though, need to get that fixed.

namesty (commented):
Should there be a scalibility concern if there are many embedded wraps/large amounts of data passed into the gateway where the bindgen wraps are hosted?

Not saying we should worry about it now. Just askin' if it makes sense to ocnsider it for the future

packages/cli/src/lib/project/PluginProject.ts
```diff
@@ -140,6 +141,7 @@ export class PluginProject extends Project<PluginManifest> {
         abi,
       },
       outputDirAbs: moduleDirectory,
+      config: bindConfig,
```
cbrzn: if the `Project` abstract class wasnt modified, why are we adding this now? couldn't we just pass `undefined` as the `config` value?

packages/cli/src/commands/codegen.ts
cbrzn: i would probably add a mapping in the CLI (having it hardcoded) but idk if this is the smartest approach

packages/cli/src/lib/project/PluginProject.ts
krisbitney: Sure, I can undo this change. I added this because I thought it should be possible to pass context into plugin bindgen as well.

packages/cli/src/commands/codegen.ts
krisbitney: My assumption was that the `embed-bindgen` wraps are quick and easy to write, so we would support all of the languages soon

cbrzn:
> Not sure why CI is not green though, need to get that fixed.

it seems that the rebuild of the wraps is broken due to a wrong path (weird, right?) I am debugging that in https://github.com/polywrap/cli/pull/1890

krisbitney:
> Should there be a scalibility concern if there are many embedded wraps/large amounts of data passed into the gateway where the bindgen wraps are hosted?
> 
> Not saying we should worry about it now. Just askin' if it makes sense to ocnsider it for the future

This is a great point. In the future, if this feature becomes more important, we can build it out differently. For example, we could generate bindings for one embedded wrap at a time.

packages/cli/src/lib/project/PluginProject.ts
krisbitney: This is done

packages/cli/src/commands/codegen.ts
dOrgJelli: Kris updated the implementation within the `wrap-abi-bindgen` repo to use the embeds from the `config` property on the wrap. This makes it an "optional" feature and there is no harm done if a wrap does not support it, the data will simply not be used by the bindgen wrap.

dOrgJelli (approved):

PR #1908, opened 2023-09-14
Typescript wrap touchups

namesty:
- Updated wrap template: removed build manifest, updated shims
- Updated VM build strategy version to use latest pwr


packages/cli/src/lib/defaults/build-strategies/wasm/javascript/vm/VERSION
```diff
@@ -1 +1 @@
-0.1.0
\ No newline at end of file
+0.1.1
```
dOrgJelli, line 1: NOTE: to release this, you need to add `| /workflows/cd-containers` to your PR's title: https://github.com/polywrap/cli/blob/origin-dev/.github/workflows/cd-containers.yaml#L12

Also, be sure to update the VM image version within the strategy: https://github.com/polywrap/cli/blob/fe8b36f676e11aaac958a75475b05961681700a7/packages/cli/src/lib/build-strategies/strategies/DockerVMStrategy.ts#L53

packages/templates/wasm/typescript/shims/msgpack.js
```diff
@@ -0,0 +1,2356 @@
+module.exports = `function utf8Count(str) {
```
dOrgJelli, line 1: There seem to be a lot of global functions being added here that are not apart of JS's globals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects

I worry that these will conflict with other functions used within user code.

I think a better way to solve this is to provide a "wrap lib" for typescript wrap devs to use, similar to how we do for Rust, AssemblyScript, and Go.

dOrgJelli (changes requested):

PR #1911, opened 2023-09-22
project manifest: default source paths

### LINKED ISSUE #1909: Polywrap Manifest: Make use of defaults
# Manifest: Make use of defaults

When it comes to developing a wrap, the user needs to define a `polywrap.yaml` manifest; in this manifest is necessary that the `module` and `schema` are defined. This actually may be simplified. The user should not need to define those unless it has some custom name or path (which is actually very rare).

We need to implement defaults per language and make the build process smart enough to find the needed files if not given

## Today's Devexp
```yaml
format: 0.5.0
project:
  name: asyncify
  type: wasm/rust
source:
  schema: ./polywrap.graphql
  module: ./Cargo.toml
```

## Tomorrow's Devexp
```yaml
format: 0.5.0
project:
  name: asyncify
  type: wasm/rust
```

krisbitney:
This PR adds default source paths to project manifests. If the user's schema and module entry point are in the default locations, the paths no longer need to be specified in the project manifest.

Closes https://github.com/polywrap/cli/issues/1909


dOrgJelli (approved):
:fire: :fire: :fire:

PR #1924, opened 2023-10-24
Small revisions and fixes For Wasm templates

krisbitney:
This PR updates Wasm templates in the following ways:
- makes the templates more consistent with each other
- removes workflows (`polywrap.test.yaml`, cue file, and yarn scripts)
- removes `polywrap.deploy.yaml` files
- fixes bugs in the rust template
- fixes bugs in the TS template
- updated to latest app codegen in tests

The TS template doesn't build due to https://github.com/polywrap/cli/issues/1923

The Go template doesn't build due to https://github.com/polywrap/cli/issues/1925

For reasoning on why we would remove workflows: https://discord.com/channels/796821176743362611/1166324052164628560

I removed the deploy manifests because the default behavior of `npx polywrap deploy` is IPFS deployment. Since we are encouraging use of IPFS and Wrapscan URIs, most users will not need to customize deployment. Our documentation will provide ample guidance for users who do wish to use a deploy manifest.


packages/templates/wasm/golang/polywrap.build.yaml
```diff
@@ -0,0 +1,4 @@
+format: 0.2.0
```
dOrgJelli: We don't need to be adding these build manifests, they're not doing anything meaningful and can be omitted.

dOrgJelli (changes requested):

packages/templates/wasm/golang/polywrap.build.yaml
krisbitney: Sounds good. I removed them.

dOrgJelli:
I'm good with merging this, but the CI is still red for the templates, looks like some of them aren't properly building:
https://github.com/polywrap/cli/actions/runs/6706112620/job/18221943589?pr=1924

